PR #27, opened 2020-08-18
Solution

AnastasiiaBortnichuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+  hide() {
+    if (!this.hidden) {
+      this.hidden = true;
```
rmnkk, line 19: It's better to change "hidden" property to the opposite value, not to true. If you invoke "hide" method more than one time it always will be true

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+  hide() {
+    if (!this.hidden) {
+      this.hidden = !this.hidden;
```
vsm2k17, line 19:
```suggestion
      this.hidden = !this.hidden;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    }
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore && herb.hidden === false) {
```
vsm2k17, line 26:
```suggestion
    if (herb instanceof Herbivore && !herb.hidden) {
```

PR #28, opened 2020-08-18
solved

nanccyy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden === true ? this.hidden = false : this.hidden = true;
```
Elena-brv, line 20: you can use such short variant this.hidden = !this.hidden

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+    return this;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden !== true) {
```
Elena-brv, line 28:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
nanccyy: you've got a point

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
+    this.hidden = false;
```
vkryvytskyy, line 16: super should accept name property

PR #30, opened 2020-08-18
Solution

asbh25:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
```
furude-rika: you can use 'not' operator instead of comparison herbivore.hidden with false
```suggestion
    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(food) {
+    if (food instanceof Herbivore && food.hidden === false) {
```
vkryvytskyy, line 28:
```suggestion
    if (food instanceof Herbivore && !food.hidden) {
```

PR #33, opened 2020-08-18
solution

Totowa95:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+  bite(herbivoreAnimal) {
+    if ((herbivoreAnimal.hasOwnProperty('hidden'))
```
furude-rika, line 31: you should just check if the herbivoreAnimal is instanse of Herbivore
```suggestion
    if (herbivoreAnimal instanseof Herbivore)
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+      && (herbivoreAnimal.hidden === false)) {
```
furude-rika, line 32:
```suggestion
      && (!herbivoreAnimal.hidden)) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+    Animal.alive.push(this);
```
vkryvytskyy, line 17: move this to the Animal's constructor, and you don't need the constructor here anymore

src/herbivoresAndCarnivores.js
vkryvytskyy: fix this

src/herbivoresAndCarnivores.js
vkryvytskyy: fix this

vkryvytskyy (approved):

zarva6596 (approved):

PR #35, opened 2020-08-18
solution

Oleg-Ozerov:


src/herbivoresAndCarnivores.js
```diff
@@ -2,16 +2,40 @@
...
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
Elena-brv, line 27: what if the animal will be another carnivore? you should check it

src/herbivoresAndCarnivores.js
```diff
@@ -2,16 +2,40 @@
...
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
vkryvytskyy, line 27:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

vkryvytskyy (approved):

PR #42, opened 2020-08-18
solution

nikita-parkhomenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.__proto__ === Herbivore.prototype && !animal.hidden) {
```
dan3si, line 26: animal instanceof Herbivore проще же)

dan3si (approved):
good

src/herbivoresAndCarnivores.js
vkryvytskyy: fix this

vkryvytskyy (approved):

PR #44, opened 2020-08-19
solution

ChrisZinch:


yuriimovchko (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
vkryvytskyy, line 7: hidden property should be in the Herbivore class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
```
vkryvytskyy, line 23:
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
```

vkryvytskyy (changes requested):

vkryvytskyy (approved):

PR #141, opened 2020-12-29
Solution

Layrovell:


dtsehelnyk (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,38 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
varseniuk, line 13:
```suggestion
  constructor(name) {
    super(name);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,38 @@
...
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
varseniuk, line 18:
```suggestion
  hide() {
    this.hidden = !this.hidden;
```

This method should work in such a way: if 'hidden' is 'true', then calling this method should swap it to 'false' and vice versa. You can use operator '!' for this purpose

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,38 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
varseniuk, line 24:
```suggestion
    if (animal.hidden === false && animal instanceof Herbivore) {
```
Carnivores can only bite herbivores

varseniuk (approved):
Good, but take into consideration my comments

PR #146, opened 2020-12-31
solution

5Mountains:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    !this.hidden ? this.hidden = true : this.hidden = false;
+  }
```
MaximZhuravlov, line 30: Можно сократить эту строку: this.hidden = !this.hidden

MaximZhuravlov (approved):

src/herbivoresAndCarnivores.js
5Mountains: Классное решение, Спасибо! Воспользуюсь (;

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.pushing(this);
+  }
+
+  static pushing(creature) {
+    this.alive.push(creature);
+  }
```
varseniuk, line 12:
```suggestion
    Animal.alive.push(this);
  }
```

You are using this operation only once so there's no need to create a method for it. You can remove your 'pushing' function

varseniuk (approved):
Good job, but take into consideration my comment

PR #149, opened 2021-01-02
solution

Dima-Semenov:


zhenya-mezhueva (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(health, name, hidden = false) {
+    super(health, name);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
varseniuk, line 21:
```suggestion
  hide() {
    this.hidden = !this.hidden;
  }
```

Method 'hide' should switch the 'hidden' state. If it's already set to 'true' it should swap it to 'false' and vice versa

varseniuk (approved):

PR #152, opened 2021-01-02
solution

Ruslanwp:


zhenya-mezhueva (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    this.hidden = true;
+  };
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super();
+    this.name = name;
+  }
```
varseniuk, line 29: no need for a constructor here. Class 'Carnivore' only adds the 'bite' method to "Animal' class and no new fields
```suggestion
```

varseniuk (changes requested):
Good job! Just fix some of my comments and all done

varseniuk (approved):

PR #158, opened 2021-01-03
solution

NazarDovhanchuk:


vsm2k17 (changes requested):
!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(health, name, hidden = false) {
+    super(health, name);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
varseniuk, line 20:
```suggestion
  hide() {
    this.hidden = !this.hidden;
  }
```
Method hide should switch the 'hidden' state of an animal. If it was already 'true' it should set it to 'false' and vice versa

varseniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive = Animal.alive.filter((el) => el.name !== animal.name);
+
+      return Animal.alive;
```
varseniuk, line 32: 👍

varseniuk (approved):
Good job

PR #164, opened 2021-01-04
solution

Alexandr-Petrenko:


5Mountains (approved):
Easy readable and understandable code! 👍 
But, in the method - bite in the class - Carnivore by the conditions of the task should also include a check that it does not bite its own type, otherwise from a wild flock will remain a loner (:

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden === false ? this.hidden = true : this.hidden = false;
```
varseniuk, line 20:
```suggestion
    this.hidden === !this.hidden;
```

Just FYI

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
varseniuk, line 26: Carnivores can only bite Herbivores. So you need to add a check if your animal is Herbivore

varseniuk (changes requested):
All is good, but let's add some minor fixes

varseniuk (approved):
Good

PR #166, opened 2021-01-05
 Solution

ShyianO:


zhenya-mezhueva (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target.hidden === false) {
```
varseniuk, line 26: Task description says:

> Carnivore has a bite method, which takes a herbivore object and decreases the object's health by 50. The method does not work if it is another сarnivore, or the herbivore is currently hiding.

The method does not work if it is another сarnivore. So you should also check here the type of your 'target'

varseniuk (changes requested):

varseniuk (approved):
👍

PR #179, opened 2021-02-20
solution

ArtemTopolskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
chri911, line 20: this method should change the hidden property of the beast to the opposite value, so it won't work, if current property is set as `true`
```suggestion
    this.hidden = !this.hidden;
```

chri911 (approved):
Good job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      const indexOfDeadAnimal = Animal.alive.findIndex(animal => {
+        return animal === herbivore;
+      });
+
+      Animal.alive.splice(indexOfDeadAnimal, 1);
```
solaryasha, line 35: Заюзай filter

solaryasha (changes requested):

solaryasha (approved):

PR #207, opened 2021-03-04
first commit

yurasokal:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
+      animal.health = animal.health - 50;
```
chri911, line 26: you can also write 
```suggestion
      animal.health -= 50;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    }
+
+    Animal.alive = Animal.alive.filter(
+      isAlive => isAlive.health > 0
```
chri911, line 30: `isAlive` looks like the name of boolean variables, you should better use here something without prefix 'is'

chri911 (approved):
Good job)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    );
```
Vladimir-Beliy, line 31: почему is ?

Vladimir-Beliy (changes requested):

Vladimir-Beliy (approved):

PR #241, opened 2021-05-01
Solution

vlad-trundenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
vkryvytskyy, line 21:
```suggestion
  hide() {
    this.hidden = !this.hidden;
  }
```
it should work in both directions

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden !== true && animal.hidden !== undefined) {
```
vkryvytskyy, line 26:
```suggestion
    if (!animal.hidden && animal instanceof Herbivore) {
```
there is a special keyword to check objects class

vkryvytskyy (changes requested):

src/herbivoresAndCarnivores.js
vlad-trundenko: ok, thanks)

src/herbivoresAndCarnivores.js
```diff
@@ -12,18 +12,18 @@ Animal.alive = [];
 
 class Herbivore extends Animal {
   constructor(name) {
-    super(name);
+    super();
```
vkryvytskyy, line 15: you should pass name to the parent's constructor

vkryvytskyy (approved):

PR #245, opened 2021-05-07
create animal hierarchy

RomanDubovskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    const biteDamage = 50;
```
vkryvytskyy, line 26: mb add this property in constructor to all of the Carnivores?

vkryvytskyy (approved):

src/herbivoresAndCarnivores.js
RomanDubovskyi: cool, thanks!

PR #248, opened 2021-05-12
Solution

herkoss:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    if (this.hidden) {
+      this.hidden = false;
+    } else {
+      this.hidden = true;
+    }
```
Ihor-Karpyn, line 24: this.hidden = !this.hidden

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
Ihor-Karpyn, line 31:
```suggestion
```

redundant

Ihor-Karpyn (approved):
OK. Left a couple of comments

src/herbivoresAndCarnivores.js
mykola-yarchuk: yep let's follow Ihor advice

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
mykola-yarchuk, line 34:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
mykola-yarchuk, line 38: let's use filter here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
```
mykola-yarchuk, line 37: move this condition to filter method

src/herbivoresAndCarnivores.js
mykola-yarchuk: disagree, we need to pass at least name

mykola-yarchuk (changes requested):
let's simplified it

src/herbivoresAndCarnivores.js
Ihor-Karpyn: When we haven't extra property, constructors is redundant. When extended class haven't constructor it's created automatically

https://prntscr.com/12vac13
https://learn.javascript.ru/class-inheritance

mykola-yarchuk (approved):
great

PR #251, opened 2021-05-12
solution

provolovskyi-illia:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
igor-stasiv, line 21: It works only once. What if you will need to change it again? Maybe you should create a more universally solution for this

igor-stasiv (approved):
Good job!

src/herbivoresAndCarnivores.js
provolovskyi-illia: Согласен)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target.hidden === false) {
```
varseniuk, line 26:
```suggestion
    if (target.hidden === false) {
```
As task description says:

> Carnivore has a bite method, which takes a herbivore object and decreases the object's health by 50. The method does not work if it is another сarnivore, or the herbivore is currently hiding.

So you should also check if the target is an herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+      target.health -= 50;
+
+      if (target.health === 0) {
+        Animal.alive.splice(Animal.alive.indexOf(target.name), 1);
```
```suggestion
        Animal.alive.splice(Animal.alive.indexOf(target), 1);
```
Does your array contain target.names or just target objects?

varseniuk (changes requested):
Nice, but let's fix the check to make the function work as it suppose in description

src/herbivoresAndCarnivores.js
provolovskyi-illia: Да, но у плотоядных животных нет свойства hidden, потому проверяя одно, я исключаю другое, зачем еще проверка?

varseniuk: Good point 🤔

varseniuk: But for better reading and understanding it would be better to specify that check too

varseniuk (approved):
Great 🔥

PR #255, opened 2021-05-13
herbivores and carnivores implemented

illisogori:


vitalya2626 (approved):
Everything is cool!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+    this.hidden = !this.hiden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
varseniuk, line 26: It would be better to put it inside the check `if (prey.health <= 0)`. In this case you will use filter not every time, but only when prey has no health

varseniuk (approved):
Good! 👍 
Left one comment

PR #256, opened 2021-05-13
solution

vitalya2626:


illisogori (approved):
Good job.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animalAlive => animalAlive.health > 0);
```
```suggestion
    Animal.alive = Animal.alive.filter(animalAlive => animalAlive.health > 0);
```
It would be better to put it inside the check if (animal.health <= 0). In this case, you will use a filter method not every time, but only when the victim has no health

varseniuk (approved):
Good! 👍  Pay attention to my comment

PR #259, opened 2021-05-13
solution

Namart:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
+  hide(hidden) {
+    if (!hidden) {
+      this.hidden = true;
+    } else {
+      this.hidden = false;
+    }
```
SkyPKWolf, line 23:
```suggestion
   this.hidden = !this.hidden;
```

SkyPKWolf: so shorter

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
+    }
+
+    if (prey.health === 0) {
+      const index = Animal.alive.indexOf(prey);
+
+      Animal.alive.splice(index, 1);
+    }
```
SkyPKWolf, line 41: easier to filter the array

SkyPKWolf (approved):
Good

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
varseniuk, line 16:
```suggestion
    Animal.alive.push(this);
```
DRY. You write this line two times: here (in Herbivore constructor) and in Carnivore constructor. Why don't you just write `Animal.alive.push(this);` inside the Animal constructor, from which these classes inherit?

src/herbivoresAndCarnivores.js
varseniuk: Good advice

src/herbivoresAndCarnivores.js
varseniuk: You can either use the filter method as Pavlo recommends, or use shorter version:
`Animal.alive.splice(Animal.alive.indexOf(prey), 1);`

varseniuk (changes requested):
Good job! But let's make it even better

varseniuk (approved):
Great! 👍

PR #260, opened 2021-05-13
Solution v0.1

SkyPKWolf:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.health = health;
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
```
scarlettjohanssons, line 14: так как мы наследуем Animal то health не обязательно

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
scarlettjohanssons, line 6: У нас свойство health 100 по дефолту, его можно убрать из аргументов и вписать просто 100 в свойство

scarlettjohanssons (approved):

src/herbivoresAndCarnivores.js
varseniuk: 100 is a default value, but there should be a possibility to make animals with different amounts of healts. So Pavlo is correct here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
varseniuk, line 16: You write this line two times: here (in Herbivore constructor) and in Carnivore constructor. Why don't you just write `Animal.alive.push(this);` inside the Animal constructor, from which these classes inherit?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(herbivore) {
```
```suggestion
  bite(prey) {
```
`target` or `prey` are more meaningful in this context

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    if (herbivore instanceof Herbivore && (herbivore.hidden !== true)) {
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
varseniuk, line 35: It would be better to put it inside the check `if (herbivore.health <= 0)`. In this case, you will use a filter method not every time, but only when the prey has no health

varseniuk (changes requested):
Nice! 👍 
But let's make some small fixes

varseniuk (approved):
Great 🔥

PR #262, opened 2021-05-13
solution

RuslanaPodshyvalova:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health) {
```
victor-arutiunov, line 6: Health is always 100 when object created, so don't need to transfer it.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+    return typeof first;
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
+    super(name, health);
+    this.hidden = hidden || false;
+    Animal.alive.push(this);
```
victor-arutiunov, line 20: We can push "_this_" in Animal class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
```
victor-arutiunov, line 29: And for now we don't need constructor in "_Carnivore_" class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
```
varseniuk, line 8:
```suggestion
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+  compare(first, second) {
+    return typeof first;
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
+    super(name, health);
+    this.hidden = hidden || false;
```
varseniuk, line 19:
```suggestion
  constructor(name, health, hidden = false) {
    super(name, health);
    this.hidden = hidden;
```

src/herbivoresAndCarnivores.js
varseniuk: You are repeating `Animal.alive.push(this);` two times. Why don't you write it once in the `Animal` class, and two other classes will inherit it?

src/herbivoresAndCarnivores.js
varseniuk: Yep. Victor is right. Carnivores have all properties that common `Animal` class but they should only have `bite` method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(herbivr) {
```
varseniuk, line 34:
```suggestion
  bite(prey) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
+  }
+
+  compare(first, second) {
```
varseniuk, line 11:
```suggestion
```
Why do you have created this method? You don't use it anywhere

varseniuk (changes requested):
Good start! 👍 
But let's fix some small isuues

varseniuk (approved):
Good 👍

PR #264, opened 2021-05-13
Solution

iryna-vys:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if ((herbivore instanceof Herbivore) && herbivore.hidden === false) {
```
varseniuk, line 25:
```suggestion
    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+      herbivore.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```
varseniuk, line 28: It would be better to perform this `filter` not everytime, but only if herbivore's health after bite will be 0.
You can add `if (herbivore.health <= 0) { *filter here*}`

varseniuk (approved):
Good! 👍 
Left some comments

PR #268, opened 2021-05-13
created battle

YuriiLiso:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = true;
```
Bogdan-Kotsupey, line 27: nice moov!

Bogdan-Kotsupey (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
```
varseniuk, line 7: as task description says:

> Create a Herbivore class. Herbivore has a method of `hide`

It means that only herbivores can hide (mot every Animal). In your case every instance will have `hidden` property, which should be onli in herbivores

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
```
varseniuk, line 16:
```suggestion
    Animal.alive.push(this);
```
Why do you adding only herbivores or carnivores to `alive` array? Don't you think that it will be better to put this line `Animal.alive.push(this);` to the `Animal` class and two other classes will inherit it?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = true;
+    Animal.alive.push(this);
+  }
+
+  bite(someone) {
+    if (someone.hidden === false) {
```
varseniuk, line 32:
```suggestion
    if (!someone.hidden) {
```
Again, let's read the description:

> Carnivore has a `bite` method, which takes a `herbivore` object and decreases the object's health by 50

As you may see, `bite` method should only bite herbivores. Now it can bite anyone (herbivores and carnivores both). Let's add some new check and fix this

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+      someone.health -= 50;
+    }
+
+    if (someone.health === 0) {
+      Animal.alive = Animal.alive.filter(el => el.health > 0);
```
varseniuk, line 37: 👍

src/herbivoresAndCarnivores.js
varseniuk, line 27: Carnivores cannot hide, so they don't need `hidden` property

varseniuk (changes requested):
Nice start, but let's make it as description asks

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
VinnieJ-2k20, line 21: `hide()` method should work in a different way

> Herbivore has a method of `hide`, which changes the `hidden` property of the beast to the opposite value and helps to hide from carnivores.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.name = name;
+  }
+
+  bite(someone) {
+    if (!someone.hidden && !(someone instanceof Carnivore)) {
```
VinnieJ-2k20, line 31: You should check if a bite target is a Herbivoire. Change the logic, please.

> Carnivore has a `bite` method, which takes a herbivore object and decreases the object's health by 50.

VinnieJ-2k20 (changes requested):
Almost done.
Some small edits to the logic of the task left.

src/herbivoresAndCarnivores.js
YuriiLiso: I understood the condition but didn't understand why Carnivore can't bite any Animals...

VinnieJ-2k20: Well, that's life. Maybe some other Animal types are too strong to be bitten :)

For now you should check if the target is the instance of herbivoire.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.name = name;
+  }
+
+  bite(herbivore) {
+    if (!herbivore.hidden && !(herbivore instanceof Carnivore)) {
+      herbivore.health -= 50;
+    }
```
VinnieJ-2k20, line 33: You only changed the variable name.
But here you need to change the logic and allow the bite only if the target belongs to herbivoires.

VinnieJ-2k20 (changes requested):
One small step left.

VinnieJ-2k20 (approved):
Good job 👍

PR #274, opened 2021-05-14
solution

vadym-liubych:


progmari (approved):
👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
```
varseniuk, line 7: Only herbivores should have `hidden` property
```suggestion
```

varseniuk (approved):
Good! But you should remove `hidden` attribute from the whole `Animal` class. Only herbivores can hide

PR #288, opened 2021-05-17
test

C-y-y-y:


starzhitskiy (approved):

starzhitskiy:
Good))

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  hide() {
+    this.hidden = !this.hidden;
+
+    return this.hidden;
```
VinnieJ-2k20, line 17: It is not necessary to return any value from this method.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
 }
 
+Herbivore.prototype.hidden = false;
```
VinnieJ-2k20, line 21: As we use classes, let's not use prototypes in this task.
Apply `hidden` property in the corresponding constructor.

VinnieJ-2k20 (changes requested):
Nice solution.
Let's make small adjustments.

VinnieJ-2k20 (approved):
Good job 👍

PR #289, opened 2021-05-17
Solution

olha-tsarova:


vitaliy-buchinchik (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide(hidden) {
+    this.hidden = !hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
+    if (obj instanceof Carnivore || obj.hidden) {
```
VinnieJ-2k20, line 26: This check works for now, but it will fail if there will be more types of animals.
Let's check if the target is instance of Herbivoire. This will make solution more stable.

VinnieJ-2k20 (approved):
Good job 👍 

One small adjustment to the logic.

PR #290, opened 2021-05-17
solution

den-bulaev:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,36 @@
...
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(dinner) {
+    if (!dinner.hidden && dinner instanceof Herbivore) {
+      dinner.health -= 50;
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
aliceblunt8, line 32: filter array if dinner's health is less or equal to zero

aliceblunt8 (approved):

PR #291, opened 2021-05-17
Solution

vitaliy-buchinchik:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
+    if (obj instanceof Carnivore || obj.hidden) {
+      return;
+    }
+
+    obj.health -= 50;
```
aliceblunt8, line 30: let's subtract health on condition

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+    Animal.alive = Animal.alive.filter(beast => beast.health !== 0);
```
aliceblunt8, line 32: wrap this in if condition, so that you don't filter an array each time you bite someone
also, health can be less then 0

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #293, opened 2021-05-18
solution

dmitry-puhliakov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
 'use strict';
-
 class Animal {
-  // write your code here
+  static addAnimal(obj) {
+    this.alive.push(obj);
+  }
+
+  static killAnimal(obj) {
+    if (!obj.health) {
```
VinnieJ-2k20, line 8: Not the best check. It will work incorrectly if somehow obj.health will be below 0.
Filter logic should also be changed.
```suggestion
    if (obj.health <= 0) {
```

VinnieJ-2k20 (changes requested):
One small thing to fix.

PR #364, opened 2021-08-14
done

SashaLifashkin:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.kind = 'carnivore';
+  }
+
+  bite(animal) {
+    if (animal.kind === 'herbivore' && !animal.hidden) {
```
ihor-jpeg, line 34: here you can use instanceof to check the animal< instead of creating a separated kind variable

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+      animal.health -= 50;
+    }
+
+    if (!animal.health) {
+      Animal.alive.sort((a, b) => b.health - a.health).pop();
```
ihor-jpeg, line 39: very smart way to remove a dead animal, but i think you should change the !animal.health condition. because it will only be true if heals is 0, if an animal has 30 health and then gets bitten, health would be -20 and your condition will not work

ihor-jpeg (approved):
good code! i left a few suggestions below;

src/herbivoresAndCarnivores.js
SashaLifashkin: Да, действительно. Тогда не придется создавать дополнительное свойство у объектов. Спасибо!

src/herbivoresAndCarnivores.js
SashaLifashkin: И тут ты тоже прав! Слишком заточено под одно условие.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive.sort((a, b) => b.health - a.health).pop();
```
VinnieJ-2k20, line 28: What if we use filter instead of sort to reduce complexity?

VinnieJ-2k20 (changes requested):
Solution is correct, but let's use more appropriate method.

VinnieJ-2k20 (approved):
Good job 👍

PR #372, opened 2021-08-17
solution

dima-tkachenko:
Лінтер не дозволяє створити статичну змінну всередині класу


.eslintrc.js
```diff
@@ -3,7 +3,6 @@ module.exports = {
   env: {
     jest: true
   },
-  parser: 'babel-eslint',
```
illiarerun, line 6: You just need to add this property manually here and static properties inside classes will work

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static isLive() {
+    this.alive = this.alive.filter(anim => anim.health > 0);
+  }
+
+  constructor(name, health) {
```
illiarerun, line 8: You don't take health as the parameter to constructor, the health of new Animal will always be 100
```suggestion
  constructor(name) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(creature) {
+    if (creature.hidden === false) {
```
illiarerun, line 30: The method should not work if creature is сarnivore, too. I understand that you use strict comparison and Carnivores don't even have hidden property, so your method works fine. But I guess it'd be better to add additional check

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+      creature.health -= 50;
+      Animal.isLive();
```
illiarerun, line 32: Going through whole alive array on each bite is a bit inefficient, maybe try to remake logic or add helper functions?

illiarerun (approved):
Good job 🔥 Needs just a liiitle fixes

.eslintrc.js
dima-tkachenko: дякую!  допомогло)

src/herbivoresAndCarnivores.js
dima-tkachenko: не побачив! дякую!

src/herbivoresAndCarnivores.js
dima-tkachenko: думаю, що цього достатньо)

src/herbivoresAndCarnivores.js
dima-tkachenko: це оптимізує код! дякую! гарна порада)

src/herbivoresAndCarnivores.js
danheim:
```suggestion
    if (!creature.hidden) {
```

danheim (approved):

PR #373, opened 2021-08-19
Solution

Crystallord:


sasha-morozov (approved):
Well done!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
```
varseniuk, line 24: If we read the task description, we will see that

> Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores

So we should change `hidden` to the opposite value (e.g. if it's t`rue` so the method `hide` should change it to the `false`). Now you always change it to `true` only

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

varseniuk (changes requested):
Good, but let's make some small fixes

varseniuk (approved):
🔥

PR #377, opened 2021-08-19
solution

maximbalyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
sasha-morozov, line 25: It also would be working)
```
```

sasha-morozov (approved):

src/herbivoresAndCarnivores.js
varseniuk: Yep. Sasha is right

varseniuk (approved):
Great! 👍 
Pay attention to Sasha's comment

PR #378, opened 2021-08-19
solution

DmytroMykoliv:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
```
Vova-Stoliar, line 9: Good solution )

Vova-Stoliar (approved):
Perfect 👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
+
+  hide() {
+    this.hidden = this.hidden;
```
VinnieJ-2k20, line 18:
```suggestion
    this.hidden = !this.hidden;
```

VinnieJ-2k20 (changes requested):
Almost done, fix one small, but important logic issue.

src/herbivoresAndCarnivores.js
DmytroMykoliv: In my work is look like this -> this.hidden = !this.hidden;
but when I push it, it becomes without '!'

DmytroMykoliv: Now it's corrected )

VinnieJ-2k20 (approved):
Good job 👍

PR #379, opened 2021-08-19
 js_herbivores_and_carnivores

andrii-hryhoruk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
```
sasha-morozov, line 32: Вместо ифа и использования метода splice, можно воспользоваться методом filter, в котором будет проверка animal.health > 0

sasha-morozov (approved):
Well Done! Left one suggestion above

src/herbivoresAndCarnivores.js
varseniuk: Sasha is right. For optimizing your cod let's just filter here all animals from array, the health of which is > 0

varseniuk (changes requested):
All is good, but let's optimize a bit

varseniuk (approved):
👍

PR #391, opened 2021-08-20
solution

Illia-Boiko:


MaksymBeliever (approved):
Looks good! 👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive.pop();
+    }
```
varseniuk, line 28:
```suggestion
    if (animal.health === 0) {
      Animal.alive.pop();
    }
```
Instead of using two methods (sort and pop), you can just use `filter` and filter all animals whos health > 0

varseniuk (changes requested):
Let's optimize your code a bit

varseniuk (approved):

PR #408, opened 2021-08-24
solution

00-22-11:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, hidden = false) {
+    super(name);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
DmytroMykoliv, line 21: Better write this.hidden = !this.hidden; , because you will always have true

DmytroMykoliv (approved):
Looks good )

Denchepornyuk (approved):

PR #412, opened 2021-08-27
done

yevhenpecherskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
+
+  hide() {
+    this.hidden = true;
```
varseniuk, line 17:
```suggestion
    this.hidden = !this.hidden;
```
Method `hide` should toggle from 'true' to 'false' and vice versa. Now you always setting it to 'true'

varseniuk (approved):
Pay attention to my comment

PR #419, opened 2021-09-24
solution

Bohdan-Kochetov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
Daerniss, line 24: you have to check if Animal is intance of Herbivore class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+      animal.health = animal.health - 50;
+    }
+    Animal.alive = Animal.alive.filter(item => item.health > 0);
```
Daerniss, line 27: Animal array will be filtered every time bite method is called, I dont think this is very efficient, you can check for current animal health, and then call it

Daerniss (changes requested):

Daerniss (approved):
🔥

Ihor-Karpyn (approved):

PR #436, opened 2021-10-01
solution

sTorba24:


caelum29 (approved):
👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    if (target.health === 0) {
```
aliceblunt8, line 32: health could possibly be less than 0

```suggestion
    if (target.health <= 0) {
```

aliceblunt8 (approved):

PR #446, opened 2021-10-03
solution

Ruslan-Mykhalchak:


Saharij (approved):
Great!!!🔥

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
aliceblunt8, line 25: fix according to the task

`Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores.`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
aliceblunt8, line 30:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+      animal.health -= 50;
+
+      Animal.animalAlive();
```
aliceblunt8, line 33: filter array on condition `if the health of the animal reaches 0`

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #447, opened 2021-10-03
solution

Artem5457:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
OleksiiBlezniuk, line 19: Herbivore has a method of `hide`, which changes the hidden property of the beast **to the opposite value**

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
GJ 🚀

PR #453, opened 2021-10-04
solution

AVronskyi:


Yashkindzhus (approved):
LGTM✨

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
```
aliceblunt8, line 16:
```suggestion
  constructor(name) {
    super(name);
    this.hidden = false;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
aliceblunt8, line 28: add another condition for filtering (not inside this if)

` If the health of the animal reaches 0, the beast dies and it should be removed from the static array Animals.alive.`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
```
aliceblunt8, line 24: add constructor for this class too

aliceblunt8 (changes requested):

PR #461, opened 2021-10-05
solution

Bogdansgit:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
-  // write your code here
+  constructor(name, health, hidden) {
+    super(...arguments);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
aliceblunt8, line 22: `Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores.`
```suggestion
  hide() {
    this.hidden = !this.hidden;
  }
```

PR #502, opened 2021-11-20
solution was added

JamesSunderlanddd:
I tried to make 'static alive = []' property inside Animals class but Eslint seems not supporting static property yet so I wasn't able to pass through Eslint tests because of it. That is why I had to change it to the variant that is in my solution.


SerhiiKirik (approved):
Well done 🔥 

so funny 😄 
image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
```
SerhiiKirik, line 11: I have the same problem with static 😢

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!(herbivore instanceof Carnivore) && herbivore.hidden === false) {
```
illiarerun, line 26:
```suggestion
    if (!(herbivore instanceof Carnivore) && !herbivore.hidden) {
```

src/herbivoresAndCarnivores.js
illiarerun: You can declare static field, but you can call it only by className

illiarerun (changes requested):
Good job 🔥  Let's work on some fixes now, and remember to filter files that you add to your commits

src/herbivoresAndCarnivores.js
JamesSunderlanddd: @illiarerun the problem is that if  I will put  'static alive = []' property inside Animals , then eslint will give me this mistake - "error  Parsing error: Unexpected token =" And I won't be able to push my solution.

illiarerun (approved):
LGTM 🔥

PR #504, opened 2021-11-22
Solution

Val-space:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (!(prey instanceof Carnivore) && !prey.hidden) {
+      prey.health -= 50;
+    }
+
+    if (prey.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(prey), 1);
```
dmtrhrytsak, line 31: I guess filter would look better here, that's just a preference though)

dmtrhrytsak (approved):
Won't miss this time;) Your coding skills are so good that I had to be standing in the queue to review this masterpiece😄 You're doing an outstanding job✨

src/herbivoresAndCarnivores.js
illiarerun: Splice would be a bit faster, buuut, if the prey is already dead (not in `alive` array) then indexOf will return `-1` and the last element will be deleted. So yes, filter would be better here

illiarerun (changes requested):
Good job 🔥  But let's fix one thing:

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
+
+  die() {
```
illiarerun, line 23: 😱

illiarerun (approved):
LGTM 🔥

PR #505, opened 2021-11-22
Solution

OlhaSpring:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden !== undefined && !animal.hidden) {
```
VinnieJ-2k20, line 24: Need to check is animal is not a `carnivoire`
> The method does not work if it is another сarnivore, or the herbivore is currently hiding.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+      animal.health = animal.health - 50;
+      Animal.alive = Animal.alive.filter(item => item.health !== 0);
```
VinnieJ-2k20, line 26: Better to check if `health >= 0`. If it falls below 0, animal still dies.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden !== undefined && !animal.hidden) {
+      animal.health = animal.health - 50;
```
VinnieJ-2k20, line 25: Use `-=`.

VinnieJ-2k20 (changes requested):
Some changes needed.

VinnieJ-2k20 (approved):
Good job 👍

PR #506, opened 2021-11-22
Solution

kurtkobein:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(harbivore) {
+    if (harbivore instanceof Herbivore && !harbivore.hidden) {
+      harbivore.health -= 50;
+    }
+
+    if (harbivore.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(harbivore), 1);
```
SerhiiKirik, line 29: If the prey is already dead (not in `alive` array) then indexOf will return `-1` and the last element will be deleted. So, filter would be better here, try to do it with a filter

SerhiiKirik (approved):
Well done 🔥
 
But let's fix one thing:

src/herbivoresAndCarnivores.js
kurtkobein: Yeah, great remark. Thanks :boom:
I didn't use filter, but I added one more condition.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  hidden = false;
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(harbivore) {
```
illiarerun, line 23:
```suggestion
  bite(herbivore) {
```

illiarerun (approved):
LGTM 🔥

PR #514, opened 2021-11-25
Zebras and tigers

DudarDaryna:


ImIkator (approved):
Well done !

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
illiarerun, line 27: The method should not work if target is another сarnivore

illiarerun (changes requested):
Good job 🔥  Need to fix just one little thing:

src/herbivoresAndCarnivores.js
DudarDaryna: Oh, agree

illiarerun (approved):
Great job 🔥

PR #516, opened 2021-11-25
solution

4qwerty:


IrynaHerasymchuk (approved):
👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
```
AlinaMatuschak, line 15: there must be just alive animals, so you don't need to create static aliveAnimals that returns filtered alive arr

AlinaMatuschak (changes requested):

aionova (approved):

PR #517, opened 2021-11-25
Solution

vlad-tuzenko:


lanebx (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  }
+
+  bite(obj) {
+    if (obj instanceof Herbivore) {
+      if (obj.hidden === false) {
+        obj.health -= 50;
+
+        if (obj.health <= 0) {
+          Animal.alive = Animal.alive.filter(animal => animal !== obj);
+        }
```
ihor-jpeg, line 38: too much unnecessary nesting.
also, don't check a boolean value for equality with another boolean value. simply use `!`

ihor-jpeg (changes requested):
Good work. But let's simplify the code a bit

ihor-jpeg (approved):
Good stuff!

PR #520, opened 2021-11-25
Solution

IrynaHerasymchuk:


Nadiyahr (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
```
aionova, line 15: you don't need to pass health as a parameter to constructor. It's always 100 when you create an animal

aionova (changes requested):

aionova (approved):
approve, but fix the comment

PR #521, opened 2021-11-25
herbivores and carnivores

lera-permiakova:


sviatoslav-myhal (approved):
Good job ! 🔥

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(el => el.health !== 0);
+  }
```
ihor-jpeg, line 30: I think it is best to name the parameter `creature`. it would make sense to name it `herbivore` only if it accepts herbivores.

also, image the animal has 25 health. the bite reduces it by 50, so it becomes -25. now, the animal is dead, but your filter function will not filter it out

ihor-jpeg (changes requested):
Good work! Let's do some fixing

src/herbivoresAndCarnivores.js
lera-permiakova: how animal can has 25 health if we always start from 100 and bite always 50?

and yes, it only accepts herbivore. carnivore doesn't bite carnivores

ihor-jpeg: I agree, in this particular task, it will never happen. but it is very unlikely that your tasks will be this simple in the future.  let's say we need to add a `heal` function that heals for 25 health. now, not only do you need to write that function, but also you need to remember to come and fix this
my point is that it's a good idea to make your code less prone to bugs.

lera-permiakova: Okay, okay 😉

ihor-jpeg (approved):
Well done!

PR #524, opened 2021-11-26
solution

lanebx:


vlad-tuzenko (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,19 +1,40 @@
...
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
ihor-jpeg, line 19: it is a good practice to use constructor, especially when you extend from another class

src/herbivoresAndCarnivores.js
```diff
@@ -1,19 +1,40 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (!herb.hidden && herb instanceof Herbivore) {
```
ihor-jpeg, line 24: your parameter name is `herb`, short for herbivore(?). but then you are checking whether it is a herbivore. seems strange

ihor-jpeg (changes requested):
Good work. Let's fix these few comments

ihor-jpeg (approved):
Well done!

PR #525, opened 2021-11-26
Solution of herbivores_and_carnivores

Dima-Ziakhor:


VladDyrenkov (approved):
GJ

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+  bite(animal) {
+    if (!animal.hidden && animal.__proto__ !== Carnivore.prototype) {
+      animal.health -= 50;
+    }
+
+    if (!animal.health) {
+      const index = Animal.alive.indexOf(animal);
+
+      Animal.alive.splice(index, 1);
+    }
```
ihor-jpeg, line 41: there are two potential things:

- your if will only work for health = 0; what if health becomes less than 0? 
- if the findIndex method does not find the item (for some reason), it will return -1 and splice with -1 will remove the last element. this is not likely to happen, but it is something to think about

ihor-jpeg (changes requested):

ihor-jpeg (approved):
good stuff!

PR #527, opened 2021-11-26
implemented task

tmrsl:


Veronika-Korona (approved):
Good job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(harbivore) {
+    if (harbivore instanceof Herbivore && !harbivore.hidden) {
```
ihor-jpeg, line 27: it doesn't make sense that you are passing a herbivore and then need to check whether what you have passed is a herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+      harbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter((animal) => animal.health <= 0);
```
ihor-jpeg, line 31: we don't really need to run this every time. if we only run the filter if the health goes to 0 or below, it would optimize the code quite a bit in certain scenarios

ihor-jpeg (changes requested):
Good work! Let's fix a few moments

ihor-jpeg (approved):
Well done!

PR #531, opened 2021-11-27
solution

sviatoslav-myhal:


lera-permiakova (approved):
🔥

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
aionova, line 4: you don't need to pass health as a parameter. It's always 100 when you create a new animal

aionova (approved):

PR #532, opened 2021-11-27
init

junglq:


aldmpr (approved):
LGTM

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
ihor-jpeg, line 22: is there an option to unhide? there should be

ihor-jpeg (changes requested):
All good. Just one tiny fix needed

aionova (approved):

PR #535, opened 2021-11-28
completed animals likes nomnomnom

i-stanko:


package.json
```diff
@@ -17,7 +17,7 @@
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "^7.13.0",
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.8.3",
```
WorkAccCom, line 20: Please do not add useless files to your PR.
You can use **git add src/** command to add needed files
**git status** command — for checking file status

WorkAccCom (approved):
Cool! 🔥

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+
+    Animal.alive.push(this);
```
ihor-jpeg, line 17: If you have more types of animals in the future you would need to write this in every animal's constructor. You can move it to the base class instead. Just for future reference.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
ihor-jpeg, line 22: once an animal is hidden, it would be nice to unhide it without creating a separate method for it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+
+    Animal.alive.push(this);
```
ihor-jpeg, line 29: Move it to the base class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  }
+
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+    }
```
ihor-jpeg, line 35: herbivore as a name here doesn't make sense. you pass a herbivore and then you are checking whether the herbivore is a verbivore

ihor-jpeg (changes requested):
Good work, but let's apply a few changes

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
aionova, line 4: health is always 100 for a new animal. No need to pass it as a parameter

aionova (approved):
GJ!

PR #536, opened 2021-11-28
solution herAndCarni

skachkov23:


volodymyr-soltys97 (approved):
GJ

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.filter(val => val.name !== animal.name);
+    }
```
ihor-jpeg, line 32: What if two or more animals have the same name? this filtering is not the best option here.

ihor-jpeg (changes requested):
Good work! But let's fix the comments

ihor-jpeg (approved):
Good stuff!

PR #539, opened 2021-11-28
Solution

aldmpr:


i-stanko (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+    }
```
ihor-jpeg, line 27: you are passing an argument called herbivore and then you check whether a herbivore is a herbivore. don't make much sense

ihor-jpeg (approved):
Good work. Pay attention to naming variables

PR #540, opened 2021-11-28
'Animals'

TatianaMorozova04:


aldmpr (approved):
NICE!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
aionova, line 4: health is always 100 for a new animal. No need to pass it as a parameter

aionova (approved):

PR #566, opened 2021-12-29
Solution

victorgri:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,38 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (animal.__proto__ !== this.__proto__ && animal.hidden === false) {
```
aliceblunt8, line 29:
```suggestion
    if (animal instanceof Herbivore  && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,38 @@
...
+      animal.health -= 50;
+    }
+
+    const newArr = Animal.alive.filter(item => item.health > 0);
+
+    Animal.alive = newArr;
```
aliceblunt8, line 35: filter only when animal health is less than 0 (wrap in if block)

```suggestion
    Animal.alive = Animal.alive.filter(item => item.health > 0);
```

aliceblunt8 (changes requested):

PR #575, opened 2021-12-30
solution

sergeyglazgo:


ArturErmolenko (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    const bitten = Animal.alive.find(beast => beast === victim);
```
Elena-brv, line 28: why do you need to do it?

Elena-brv (changes requested):

src/herbivoresAndCarnivores.js
sergeyglazgo: of course, victim is already our needed object and we don't need to find it, fixed

Elena-brv (changes requested):

Elena-brv (approved):

PR #576, opened 2021-12-30
Solution🐇🐺

HrynykNataliia:


ArturErmolenko (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  hidden = false;
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
Elena-brv, line 22: it could be not a herbivore, so use a more matchable variable's name

Elena-brv (approved):

PR #579, opened 2021-12-31
solution

mykola-koval:


vladyslav-harkusha (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    this.health = health;
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
+    this.hidden = false;
+    Animal.alive.push(this);
```
Elena-brv, line 15: this method should be in the main class, no need to duplicate it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
Elena-brv, line 13:
```suggestion
  constructor(name) {
    super(name);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    this.hidden ? this.hidden = false : this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
+    Animal.alive.push(this);
+  }
```
Elena-brv, line 26: no need constructor here at all

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  bite(victim) {
+    if (!victim.hidden && !(victim instanceof Carnivore)) {
```
Elena-brv, line 28: I think it will be more clear

```suggestion
    if (!victim.hidden && victim instanceof Herbivore) {
```

Elena-brv (changes requested):

src/herbivoresAndCarnivores.js
mykola-koval: I agree but i wrote exactly like this because of a possibility of implementing more  Animal subclasses in the future (e.g. omnivores)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden ? this.hidden = false : this.hidden = true;
```
Elena-brv, line 18: didn't see it. Use it

image

Elena-brv (changes requested):

Elena-brv (approved):

PR #582, opened 2022-01-03
Solution

YevheniiKorzun:


Volodymyr1806 (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if ((!beast.hidden) && (beast.hidden !== undefined)) {
```
Elena-brv, line 27: image

and why do you need to check if it is not undefined?

Elena-brv (changes requested):

src/herbivoresAndCarnivores.js
YevheniiKorzun: I did it to check does "beast" is herbivore. Becouse only herbivores have "hidden" property. But I agree that the "beast instanceof Herbivore" is more correctly and readable solution. Thank you!

Elena-brv (approved):

PR #593, opened 2022-01-04
Solution

IrynaZolko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (Object.prototype.hasOwnProperty.call(animal, 'hidden')
```
Basil-Panasiuk, line 28: maybe need check instanceOf)

Basil-Panasiuk (approved):

src/herbivoresAndCarnivores.js
Denchepornyuk, line 28: use instanceof instead of hasOwnProperty and animal.hidden should be false

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #604, opened 2022-01-05
Task is done

ArtemKa2208:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (!animal.hidden && animal.__proto__ !== Carnivore.prototype) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      const index = Animal.alive.findIndex(value => value === animal);
```
Mikedacheka, line 38: I think you can simplify it

Mikedacheka (approved):
awesome!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.name = name;
+    this.health = 100;
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.health = 100;
```
Elena-brv, line 15: do not repeat it, you have it on parent constuctor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.hidden = false;
+    Animal.alive.push(this);
```
Elena-brv, line 17: what if it will be 50 child classes? you should write this line on each. Write it on parent constructor one time and that's all

Elena-brv (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (!animal.hidden && animal.__proto__ !== Carnivore.prototype) {
```
Elena-brv, line 33: use instanceOf

Elena-brv (approved):

PR #606, opened 2022-01-08
solution

Viktor9r:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
Mikedacheka, line 22: Due to checklist ,point 3:
[CODE KNOWLEDGE] When you need to toggle values, use logical operator `!`

Mikedacheka (approved):

src/herbivoresAndCarnivores.js
Olexandr6: Agree

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (!victim.hidden && victim instanceof Herbivore) {
+      victim.health -= 50;
+
+      if (victim.health === 0) {
+        Animal.alive = Animal.alive.filter(animal => animal !== victim);
+      }
+    }
```
Olexandr6, line 34: Avoid nesting
```suggestion
    if (!victim.hidden && victim instanceof Herbivore) {
      victim.health -= 50; 
    }
    
     if (victim.health === 0) {
        Animal.alive = Animal.alive.filter(animal => animal !== victim);
    }
```

Olexandr6 (changes requested):
Good job 👍 , but some things should be optimized.

Olexandr6 (approved):

PR #607, opened 2022-01-09
Solution

AlonaSnigur:


Mikedacheka (approved):
awesome!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+  }
+
+  static newAnimal(animal) {
+    Animal.alive.push(animal);
+  }
```
Olexandr6, line 11: Actually, this static method is unnecessary. Just push into `Animal.alive` array, where it's needed

Olexandr6 (approved):

PR #620, opened 2022-01-31
solution

vsuslov29:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  bite(animal) {
+    if (
+      animal instanceof Herbivore
+      && !animal.hidden
+    ) {
+      animal.health -= 50;
+
+      Animal.alive = Animal.alive.filter(({
+        health,
+      }) => health > 0);
```
mtctxd, line 35:
```suggestion
      Animal.alive = Animal.alive
          .filter(({health}) => health > 0);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (
+      animal instanceof Herbivore
+      && !animal.hidden
```
mtctxd, line 29:
```suggestion
      animal instanceof Herbivore
          && !animal.hidden
```

mtctxd (approved):
Ніби все ок.

src/herbivoresAndCarnivores.js
Anna-Tkachenko: If you have only one iteration method without a chain you can write it in the same string

Anna-Tkachenko (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
Anna-Tkachenko, line 16: Don't forget get name in constructor and pass it in super()

PR #638, opened 2022-02-06
Solution

jursylian:


Tsekhmister (approved):
👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
Anna-Tkachenko, line 16: You should get name in constructor and pass it to super()

Anna-Tkachenko (changes requested):

PR #639, opened 2022-02-06
Solution

maxim2310:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
Anna-Tkachenko, line 21: Method hide should change hidden property to the opposite value, not always true

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
+Herbivore.hidden = false;
```
Anna-Tkachenko, line 24: It's redundant

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #646, opened 2022-02-07
solution

GitRoko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    (this.hidden)
+      ? this.hidden = false
+      : this.hidden = true;
+  }
```
Anna-Tkachenko, line 24: You can make it simpler this.hidden = !this.hidden

Anna-Tkachenko (approved):

PR #648, opened 2022-02-07
Solution

RomaBoreiko:


Oleksii-Fedorenko (approved):
красава чоловіче)))!!!!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.map(creature => creature.health > 0);
```
Anna-Tkachenko, line 32: Method map returns an array of changed elements, so it returns array with true / false values, you should use method filter here

Anna-Tkachenko (changes requested):

PR #650, opened 2022-02-07
Solution

cxrmx:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
Anna-Tkachenko, line 16: You should do it in Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
```
Anna-Tkachenko, line 26: Same here, you should add animals in Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  bite(herbivore) {
+    if ((herbivore.hidden === false)) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        for (let i = 0; i < Animal.alive.length; i++) {
+          if (Animal.alive[i] === herbivore) {
+            Animal.alive.splice(i, 1);
+          }
+        }
```
Anna-Tkachenko, line 37: It's better to use iteration method instead of cycle

Anna-Tkachenko (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    if ((herbivore.hidden === false
+      && herbivore.constructor.name === 'Herbivore')) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive = Animal.alive.map(el => {
+          if (el === herbivore) {
+            Animal.alive.splice(el, 1);
+          }
+        });
```
OleksiiBlezniuk, line 37: use filter

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+  bite(herbivore) {
+    if ((herbivore.hidden === false
+      && herbivore.constructor.name === 'Herbivore')) {
```
OleksiiBlezniuk, line 29: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof

OleksiiBlezniuk (approved):

PR #651, opened 2022-02-07
Done

tt-denis:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,37 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
+  }
```
Anna-Tkachenko, line 18: Method hide should change hidden property to the opposite value, not always set true

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job 👍

PR #657, opened 2022-02-18
Killers and victims

aspurgilus:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,47 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static birth(beast) {
+    this.alive.push(beast);
+  }
```
SanyaBratashchuk, line 6: unnecessary static property, you can push from constructors directly to Animal.alive array

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,47 @@
...
+      victim.health -= 50;
+
+      if (victim.health === 0) {
+        const index = Animal.alive.indexOf(victim);
+
+        if (index !== -1) {
+          Animal.alive.splice(index, 1);
+        }
+      }
+    }
```
SanyaBratashchuk, line 39: It is better to avoid nesting of If

SanyaBratashchuk (approved):
Good job

aspurgilus:
> Good job

Thank's for the review. I improved my code due to your remarks.

Ihor-Karpyn (approved):

PR #670, opened 2022-02-23
Solution

GoDfreE25:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  };
+
+  hide() {
+    this.hidden = true;
```
jstmpelowycz, line 20: Here you should reverse `hidden`  every time you call this method, don't you?

jstmpelowycz, line 20: I would suggest the following changes ⬇️
```suggestion
    this.hidden = !this.hidden;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && prey.hidden === false) {
```
jstmpelowycz, line 26:
```suggestion
    if (prey instanceof Herbivore && !prey.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+      prey.health -= 50;
+      prey.health === 0 && Animal.alive.splice(Animal.alive.indexOf(prey), 1);
```
jstmpelowycz, line 28: That's not readable, you would suggest the following changes ⬇️ 
```suggestion
            if (!prey.health) {
        const preyIndex= Animal.alive.indexOf(prey);

        Animal.alive.splice(preyIndex, 1);
```

Ihor-Karpyn (approved):

PR #672, opened 2022-03-03
Solution for  js_herbivores_and_carnivores

murat-yesimbietov:


Bohdan-Maksymiuk (approved):
Looks good 👍 
You can add `parser: "babel-eslint"`  into .eslintrs.js to pass the test in this task
or add `Animal.alive = [];` after Animal class

PR #687, opened 2022-03-27
write code

Batick3737:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
+    if (!(obj instanceof Carnivore) && obj.hidden === false) {
```
vkryvytskyy, line 25:
```suggestion
    if (obj instanceof Herbivore && !obj.hidden) {
```

vkryvytskyy (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(health = 100, name) {
+    super(health, name);
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
```
ArturErmolenko, line 19: Screenshot 2022-03-29 at 16 55 53

when it will be true, it should be changed to false

ArturErmolenko (dismissed):

PR #688, opened 2022-03-27
Solution

YevgeniySavchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
...
   // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
ArturErmolenko, line 25: better use this.hide = !this.hidden

PR #699, opened 2022-04-11
Solution

yaremenko-maksym:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
+    if (obj.hidden === false) {
```
VETALEK, line 27: it works, but you need to check if `obj` is herbivore too (e. g. what if carnivores will be able to hide)

VETALEK (changes requested):

.eslintrc.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   extends: '@mate-academy/eslint-config',
-  parser: "babel-eslint",
```
AlinaMatuschak, line 3: ?

AlinaMatuschak (approved):

PR #705, opened 2022-04-15
Solution

Anastasiia-Svintsova:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
-
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+    this.hidden = false;
```
ivan-kadykalo, line 6: Carnivores shouldn't have a hidden parameter. Only Herbivores should have it. You should replace it to Herbivore class.

ivan-kadykalo (changes requested):
All is good except one thing

MaksimusCRS (approved):

PR #718, opened 2022-04-19
soved

misha200119:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!(herbivore instanceof Carnivore) && !herbivore.hidden) {
```
YegorKochetkov, line 35:
```suggestion
    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```

YegorKochetkov (approved):
I like your solution with get and set)

src/herbivoresAndCarnivores.js
misha200119: probably some guy create animal as ''abstract" class Animal, but i agee with this, if i wanna deal with parent class like herbivore, i  too need move  hide prop to it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = !this.hidden;
```
MaksimusCRS, line 29: Если вызвать два раза hide, то животное окажется не спрятанным, так быть не должно)

src/herbivoresAndCarnivores.js
MaksimusCRS: исправь это тоже)

MaksimusCRS (changes requested):

src/herbivoresAndCarnivores.js
misha200119: Create a `Herbivore` class.
Herbivore has a method of `hide`, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores.
?

MaksimusCRS (approved):

PR #724, opened 2022-04-19
herbivors_and_carnivors

DragotaIV:


src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,46 @@
...
+    if ((obj instanceof Herbivore) && (obj.hidden === false)) {
+      obj.health = obj.health - 50;
+      // И если объект принадлежит классу травоядные и не спрятался
+      // Тогда у него отнимаются жизни
+    }
+
+    Animal.alive = Animal.alive.filter((el) => el.health > 0);
+  };
+};
+Animal.alive = [];
```
Bohdan-Maksymiuk, line 43: better do add after Animal class, but in this case it doesn't matter

Bohdan-Maksymiuk (approved):
Great job 👏

PR #728, opened 2022-04-20
done

kurochka-vladyslav:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
+    }
```
YegorKochetkov, line 32: use filter. mentor will advise)

YegorKochetkov (approved):
👍

src/herbivoresAndCarnivores.js
MaksimusCRS: agree, use filter)

MaksimusCRS (approved):

PR #737, opened 2022-04-20
solution

barannikovArtem:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
YegorKochetkov, line 21: I was told that if hide() - you should to hide only, not toggle hidden\non-hidden
otherwise the function name would be like toggleHidden()
```suggestion
  hide() {
    this.hidden = true;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(herbivore), 1);
```
YegorKochetkov, line 31: I was advised to use filter)

YegorKochetkov (changes requested):
I made like you and catch comments from mentor)

Andr1yK (approved):

PR #738, opened 2022-04-20
Solution

AmbeMO:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
+    Animal.born(this);
+  }
 }
 
+Animal.alive = [];
+
+Animal.born = (animal) => {
+  Animal.alive.push(animal);
+};
```
YegorKochetkov, line 16: You don't need this, you may push immediately 
```suggestion
   Animal.alive.push(this);
  }
}

Animal.alive = [];
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (!target.hidden && target instanceof Herbivore) {
+      target.health -= 50;
+    }
+
+    if (target.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(target), 1);
```
YegorKochetkov, line 36: I was advised to use filter)

YegorKochetkov (changes requested):
Good) but fix few flaws

PR #755, opened 2022-05-17
solution

efdalyalcin:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      const index = Animal.alive.indexOf(animal);
+
+      Animal.alive.splice(index, 1);
+    }
+  }
```
LukaszGryziewicz, line 35: from checklist
2. [CODE PERFOMANCE] - Try to avoid nesting loops (indexOf + splice), use filter instead

LukaszGryziewicz (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
ihor-jpeg, line 30: what if health is less than zero? the filter function is okay. the condition is not

ihor-jpeg (changes requested):

ihor-jpeg (approved):

PR #756, opened 2022-05-17
solution

LukaszGryziewicz:


andreisucman (approved):
Good job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    const isHerbivore = animal instanceof Herbivore;
+    const isNotHidden = animal.hidden === false;
+
+    if (isHerbivore && isNotHidden) {
+      const biteDamage = 50;
+      const isDead = animal.health - biteDamage === 0;
```
ihor-jpeg, line 31: what if health after bite is -20?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+      const removeFromAlive = () => {
+        Animal.alive = Animal.alive.filter(beast => beast.health <= 0);
+      };
+
+      isDead
+        ? removeFromAlive()
+        : animal.health -= biteDamage;
```
ihor-jpeg, line 38: strange. ternary is usually used to assign a value or pick which one to return

ihor-jpeg (changes requested):

ihor-jpeg (approved):

PR #757, opened 2022-05-17
task completed

andreisucman:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  bite(prey) {
+    if (prey instanceof Herbivore && Animal.alive.includes(prey)) {
+      if (!prey.hidden) {
+        prey.health -= 50;
+      }
+    }
+
+    if (prey.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(prey), 1);
+    }
```
LukaszGryziewicz, line 34: from checklist
2. [CODE PERFOMANCE] - Try to avoid nesting loops (indexOf + splice), use filter instead

LukaszGryziewicz (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(hidden) {
+    super(hidden);
+    this.hidden = false;
```
ihor-jpeg, line 16: very confusing. you pass hidden into super. then u set hidden = false here. let's fix the names

ihor-jpeg (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    if (prey instanceof Herbivore && Animal.alive.includes(prey)) {
+      if (!prey.hidden) {
+        prey.health -= 50;
+      }
+    }
+
+    if (prey.health === 0) {
+      Animal.alive = (Animal.alive.filter(animal =>
+        animal.health > 0));
+    }
```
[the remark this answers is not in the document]
andreisucman: True, updated.

PR #758, opened 2022-05-17
solution

alexandra-protyanova:


vlad-demchuk (approved):
Good job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
+    if ((name instanceof Herbivore) && (name.hidden === false)) {
+      name.health -= 50;
+    }
+
+    if (name.health === 0) {
+      Animal.alive = Animal.alive.filter(x => x.health !== 0);
```
vlad-demchuk, line 31: use descriptive names of variables instead of 'x'

Anna-Tkachenko: agree, pay attention to variable names

Anna-Tkachenko (approved):
⭐️

PR #760, opened 2022-05-17
Solution

amitpatiljc:


andreisucman (approved):
💯

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  };
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (!(victim instanceof Carnivore) && !victim.hidden) {
```
59MrRobot, line 26: I think you can just use instanceof Herbivore to make your code more simple

59MrRobot (changes requested):

src/herbivoresAndCarnivores.js
amitpatiljc: Right ! Done

ihor-jpeg (approved):

PR #762, opened 2022-05-17
animal

Andriyk0:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
Anna-Tkachenko, line 20:
```suggestion
    this.hidden = !this.hidden;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbObj) {
+    if (herbObj.hidden === false) {
```
Anna-Tkachenko, line 26: You should check is this object is herbivore (use instance of method)

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #763, opened 2022-05-17
Develop

yelnikov-andrii:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden) {
+    super(name, health, hidden);
```
Andriyk0, line 14: you dont need (health) !
you extends this key from Animal!

and hidden write only in constructor!
in super you write key only from extends class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  hide(hidden) {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
Andriyk0, line 25: you dont need this

Andriyk0 (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(hidden) {
+    super(hidden);
+    this.hidden = false;
+  }
+  hide(hidden) {
+    this.hidden = true;
+  }
```
Anna-Tkachenko, line 19: Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores

Anna-Tkachenko: You should change hidden property to the opposite value, not always to the true

Anna-Tkachenko (changes requested):

Anna-Tkachenko (changes requested):
You didn't fix my comments

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(hidden) {
+    super(hidden);
```
Anna-Tkachenko, line 14: You shouldn't pass hidden value to the Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  hide(hidden) {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore) {
+      if (herb.hidden === false) {
```
Anna-Tkachenko, line 25: redundant if
combine two if statements in one main statement using logic operator &&

Anna-Tkachenko (changes requested):

src/herbivoresAndCarnivores.js
Anna-Tkachenko: just super()

Anna-Tkachenko (approved):

PR #764, opened 2022-05-17
solution

AnastasiyaSopotnytska:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
khozhainov-aleksandr, line 7: You can also try this
`this.health = 100`

this is no need 
`(name, health **= 100**)`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, hidden = false) {
+    super(name);
+    this.hidden = hidden;
```
khozhainov-aleksandr, line 16: `this.hidden = false`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
+  hide() {
+    this.hidden = !this.hidden;
```
khozhainov-aleksandr, line 19: `this.hidden = true`

khozhainov-aleksandr (changes requested):
👨🏻‍💻 🔥 👍🏻

src/herbivoresAndCarnivores.js
Anna-Tkachenko: agree

src/herbivoresAndCarnivores.js
Anna-Tkachenko: @khozhainov-aleksandr Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores

Anna-Tkachenko (approved):
⭐️

PR #765, opened 2022-05-17
Task complete

MabuzaM:


amitpatiljc (approved):
Good job 👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
ihor-jpeg, line 31: basically, we loop over the entire array every time. even if no creature died

ihor-jpeg (changes requested):

src/herbivoresAndCarnivores.js
MabuzaM: Owkay noted. Have made changes.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(creature) {
+    if (creature instanceof Herbivore && !creature.hidden) {
+      creature.health -= 50;
+
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
ihor-jpeg, line 30: you iterate over this array even if no creature died. this is not very efficient

ihor-jpeg (changes requested):

src/herbivoresAndCarnivores.js
MabuzaM: Oh! yah I get it. How about now?

PR #766, opened 2022-05-17
Herbivores and Carnivores

khozhainov-aleksandr:


AnastasiyaSopotnytska (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
Anna-Tkachenko, line 21: Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores

Anna-Tkachenko: You should change this property to the opposite value, not always to the true

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #767, opened 2022-05-17
Develop

ZUBOV-ILLIA:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && beast.hidden === false) {
+      beast.health -= 50;
+    }
+
+    if (beast.health <= 0) {
```
SavMary, line 31: Unnecessary loop. Your code can work without 'if'

SavMary (changes requested):

src/herbivoresAndCarnivores.js
ZUBOV-ILLIA: I put here `if` in case if not need delete some animal from array, and only if need, we will spend computer resources to overwrite the array )

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
Anna-Tkachenko, line 22: Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores

Anna-Tkachenko: You should change this property to the opposite value, not always to the true

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #768, opened 2022-05-17
Solution

JustDan17:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  health = 100;
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
```
ZUBOV-ILLIA, line 8: I think health property must be in constructor ? @Anna-Tkachenko

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
+  hide() {
+    if (this.hidden) {
+      this.hidden = false;
+    } else {
+      this.hidden = true;
+    }
```
Anna-Tkachenko, line 22: Check point 3 in checklist

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore
+      && prey.hidden === false
+      && Animal.alive.includes(prey)
+    ) {
+      prey.health -= 50;
+
+      if (prey.health <= 0) {
+        Animal.alive.splice(Animal.alive.indexOf(prey), 1);
```
Anna-Tkachenko, line 35: Will be better to use filter method
Here you used 2 array iterations, with filter it'll be only 1 iteration

Anna-Tkachenko (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore
+      && prey.hidden === false
+      && Animal.alive.includes(prey)
```
Anna-Tkachenko, line 26: this check is redundant

Anna-Tkachenko (approved):

PR #771, opened 2022-05-17
animals creater

uliton:


AndriiRut (changes requested):
In 30th string You can use array method instead of 'for'. you trying check all obj's property  Health. You allready have the link to objeck in bite() Method arg. use it to finde the beast in array 'alive'

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    if (this.hidden === false) {
+      this.hidden = true;
+    } else {
+      this.hidden = false;
+    }
```
Anna-Tkachenko, line 24: CheckCheck point 3 in checklist

Anna-Tkachenko (approved):

PR #772, opened 2022-05-17
Solution

RomekIvantsiv:


valerii-derkach (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  constructor(name) {
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(hidden) {
+    super(hidden);
```
Anna-Tkachenko, line 15: you shouldn't pass hidden argument to super constructor

Anna-Tkachenko (approved):

PR #774, opened 2022-05-18
Solution

AndriiRut:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
-  // write your code here
+  type = 'Carnivore'
+
+  bite(bist) {
+    if (bist.type === 'Herbivore' && bist.hidden === false) {
+      bist.health -= 50;
+    }
+
+    if (bist.health <= 0) {
+      Animal.alive.splice(cutie => cutie === bist);
```
alexandra-protyanova, line 37: Wow! It's looks impossible, but works)

alexandra-protyanova (approved):

src/herbivoresAndCarnivores.js
AndriiRut: Fixed after meating with Alexandra in zoom)

Anna-Tkachenko (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
+    this.hidden = hidden;
+  }
+
+  hide() {
+    if (this.hidden === true) {
+      this.hidden = false;
+    } else {
+      this.hidden = true;
+
```
Anna-Tkachenko, line 26: Check point 3 in checklist

Anna-Tkachenko (approved):

PR #776, opened 2022-05-18
animals game

kitaminary:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,42 @@
 
 class Animal {
   // write your code here
+  health = 100;
```
dellachaise, line 5: I think health would be better pass as a parameter to constructor with default value, for some case.

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,42 @@
...
+    Animal.alive.push(this);
+    // return this
+  }
+
+  static alive = [];
 }
 
 class Herbivore extends Animal {
   // write your code here
+  constructor(name) {
```
dellachaise, line 21: Maybe hidden better add as a parameter to constructor with default value false, to make code more understandable.

dellachaise (approved):
Great job!

Please, look at comments.

src/herbivoresAndCarnivores.js
kitaminary: if you ask me, its redundant, becouse its weel be identically

dellachaise: ok)

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,42 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(somebody) {
+    if ((somebody instanceof Herbivore)
+     && (somebody.hidden === false) && Animal.alive.includes(somebody)) {
```
Anna-Tkachenko, line 35: includes check is redundant here

Anna-Tkachenko (approved):
⭐️

PR #777, opened 2022-05-18
Added solution to 'Herbivores and Сarnivores' task

dellachaise:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+  hide() {
+    return (this.hidden = !this.hidden);
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
+    super(name, health, hidden);
+    Animal.alive.push(this);
```
kitaminary, line 27: you don't need added this everytime, it's redundant, just left it in class Animal

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+  }
+
+  bite(object) {
+    const isCarvinore = object instanceof Carnivore;
+
+    if (object.hidden !== true && !isCarvinore) {
+      object.health -= 50;
+
+      if (object.health === 0) {
+        const index = Animal.alive.indexOf(object);
```
kitaminary, line 37: i think method filter more understandable , in this case

kitaminary (approved):
Good job, but fix comments, about static method, you can pass this method from parent class

src/herbivoresAndCarnivores.js
dellachaise: I added it because without it in Animal.alive array was only Herbivores, no Сarnivores. But I solved this problem. Thank you!

PR #781, opened 2022-05-18
Solution

anastasiia-nikita:


volodya-leskiv (approved):
Good solution 👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100, hidden = true) {
+    this.name = name;
+    this.health = health;
+    this.hidden = hidden;
```
Anna-Tkachenko, line 9: this property is redundant in Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  constructor(name, health, hidden) {
+    super(name, health, hidden);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
Anna-Tkachenko, line 24: method hide should change hidden property to the opposite value, not always to the true

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #784, opened 2022-05-18
Herbivores and carnivored

dasha-chekmariova:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
```
valerii-derkach, line 18: This method should change hidden property to the opposite value, not always to the true

valerii-derkach (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #785, opened 2022-05-18
task is solved

enrikadej:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      this.hidden = !this.hidden;
+    }
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+
+  bite(herbi) {
+    if (herbi.__proto__ === Herbivore.prototype
```
dasha-chekmariova, line 29: you can use ```instanceof``` method. This operator tests to see if the prototype property of a constructor appears anywhere in the prototype chain of an object.

dasha-chekmariova (approved):
Well done!🏆

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  hidden = false;
+
+  hide() {
+    if (this.hidden) {
+      this.hidden = false;
+    } else {
+      this.hidden = !this.hidden;
+    }
+  }
```
Anna-Tkachenko, line 23: Check point 3 in checklist

src/herbivoresAndCarnivores.js
Anna-Tkachenko: agree, use instanceof

Anna-Tkachenko (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbi) {
+    if (herbi instanceof Herbivore
+        && herbi.health > 0
+        && herbi.hidden === false) {
```
Anna-Tkachenko, line 26:
```suggestion
    if (
      herbi instanceof Herbivore
      && herbi.health > 0
      && herbi.hidden === false
    ) {
```

Anna-Tkachenko (approved):
⭐️

PR #786, opened 2022-05-18
solution

Uktam19980416:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,38 @@
...
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
amitpatiljc, line 33: what if health is less than zero? use of  filter function great but the condition is not

amitpatiljc (changes requested):
Good Job! Just one suggestion

amitpatiljc (approved):

meforoma (approved):

PR #788, opened 2022-05-18
Animal extends

Oleksandr-Ivanchenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
Anna-Tkachenko, line 16: Don't repeat yourself (16 and 26 string): move this code to the Animal constructor

Anna-Tkachenko (approved):

PR #789, opened 2022-05-19
herbivores

elkhayate:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animals) {
+    if (!animals.hidden && animals instanceof Herbivore) {
+      animals.health -= 50;
+    }
+
+    if (animals.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animals), 1);
```
amitpatiljc, line 31: from checklist
2. [CODE PERFOMANCE] - Try to avoid nesting loops (splice  + indexof), use filter instead

amitpatiljc (changes requested):
Good Job! Just one suggestion

src/herbivoresAndCarnivores.js
elkhayate: Ur right thank you, i will try to fix that asap.

amitpatiljc (approved):
Good Job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animals) {
+    if (!animals.hidden && animals instanceof Herbivore) {
+      animals.health -= 50;
```
meforoma, line 27: plural `animals` misleads when reading your code
you're not working with plural (array, list etc.) - i know `animal` is reserved already, but you can come up with another relevant singular to maintain code readability - say, `victim` / `creature` etc.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    }
+
+    if (animals.health <= 0) {
+      // Animal.alive.splice(Animal.alive.indexOf(animals), 1);
```
meforoma, line 31: cleanup the commented code

meforoma (changes requested):

src/herbivoresAndCarnivores.js
elkhayate: i will thank you

meforoma (approved):

PR #791, opened 2022-05-19
solution

PRO-GRAM-MER:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,37 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  };
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(victim) {
+    if (!(victim instanceof Carnivore) && !victim.hidden) {
```
amitpatiljc, line 28: I think you can just use instanceof Herbivore to make your code more simple

amitpatiljc (changes requested):
Good Job! Just one suggestion.

amitpatiljc (approved):
Good job 👍

meforoma (approved):

PR #796, opened 2022-05-21
The_first_commit

igoree88:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    Animal.addToAlive(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
+  hide() {
+    this.hidden = true;
+  }
```
Anna-Tkachenko, line 28: Method hide should change hidden property to the opposite value, not always to the true

Anna-Tkachenko (approved):
⭐️

PR #800, opened 2022-05-31
Solution

vlad-dan1589:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,38 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(object) {
+    if (object.hidden === false) {
```
aliceblunt8, line 29: forgot one condition

>The method does not work if it is another сarnivore, or the herbivore is currently hiding.

```suggestion
    if (!object.hidden) {
```

aliceblunt8 (changes requested):

src/herbivoresAndCarnivores.js
vlad-dan1589: Understood, thx

aliceblunt8 (approved):

PR #819, opened 2022-06-18
Solution

ik-web:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      const animalIndex = Animal.alive.indexOf(animal);
+
+      if (animalIndex !== -1) {
+        Animal.alive.splice(animalIndex, 1);
+      }
+    }
```
stanislav-topikha, line 37: Checklist, I had the same problem, try to reverse logic, and use filter
image

stanislav-topikha (changes requested):
Looks good, but check the comment 🔽

stanislav-topikha (approved):

jstmpelowycz (approved):

PR #821, opened 2022-06-20
add solution

Galyopa:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  static removeAnimal(animal) {
+    delete this.alive[this.alive.indexOf(animal)];
+  }
+
+  constructor(name, health = 100) {
```
jstmpelowycz, line 10:
```suggestion
  constructor(name) {
```
`health` always will be 100, you don't even need to pass it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(hebrivore) {
+    if (!hebrivore.hidden && hebrivore instanceof Herbivore) {
+      hebrivore.health -= 50;
+
+      if (hebrivore.health === 0) {
```
jstmpelowycz, line 33:
```suggestion
      if (!hebrivore.health) {
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #824, opened 2022-06-21
solved

mflkvsk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
-
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
jstmpelowycz, line 3: you don't need `health` passed over constructor, this one is set by default under the hood
```suggestion
  constructor(name) {
    this.health = 100;
    this.name = name;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
jstmpelowycz, line 14: the same way here

jstmpelowycz (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (!animal.health) {
+      Animal.alive = Animal.alive.filter(beast => beast.health);
```
illiarerun, line 32: I'd suggest you create method with `dying` logic

illiarerun (approved):
GJ, left one comment

PR #830, opened 2022-06-22
solution

Angelion541:


ostapfriend (approved):
Cool)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(brute) {
+    if (brute instanceof Herbivore && !brute.hidden) {
+      brute.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(beast => beast.health);
```
jstmpelowycz, line 31: it works, however, every time `bite` function is called, this filtering gets performed anyway, do you think it's okay?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    Animal.alive.push(this);
+  };
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
jstmpelowycz, line 15: You forgot to pass values to your constructor

jstmpelowycz (changes requested):

PR #831, opened 2022-06-22
Solution

NikitaKarbazin:


Kaunin1 (approved):
Well done !

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
jstmpelowycz, line 20:
```suggestion
  constructor() {
    super();
    this.hidden = false;
  }

  hide() {
    this.hidden = !this.hidden;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health);
```
jstmpelowycz, line 29: it works, however, every time `bite` function is called, this filtering gets performed anyway, do you think it's okay?

NikitaKarbazin: Agree with u! I wanted to do more simple code! But not suppose the main logic! Fixed!

illiarerun (approved):

PR #837, opened 2022-06-23
Solution

Tsymbal-Viacheslav:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
+  constructor() {
+    super();
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+
+    return this;
```
jstmpelowycz, line 21: you don't need to return anything

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden !== true && animal instanceof Herbivore) {
```
jstmpelowycz, line 27:
```suggestion
    if (!animal.hidden && animal instanceof Herbivore) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(beast => beast.health > 0);
```
jstmpelowycz, line 31: wrap with if-condition, that's not okay you will filter every time `bite` is called

jstmpelowycz (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,36 @@
...
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
illiarerun, line 13: you should pass name param here and also to super constructor

mykola-koval (approved):

PR #838, opened 2022-06-23
solution

viktorlyt:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden === true) {
+      return;
+    }
+
+    animal.health -= 50;
```
Olexandr6, line 28: Let's combine this

Olexandr6: I mean, let's check if animal is Herbivore and animal.hidden === false -> in this case you should decrease animal.health

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+    if (animal.health === 0) {
```
Olexandr6, line 30: what if health is lower than 0?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+      Animal.alive.splice(animal);
```
Olexandr6, line 31: You need to delete animal from `Animal.alive` array, but looks like you forgot how splice method works.
Check [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)

Olexandr6: Let's use `filter` here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Animal {
-  // write your code here
+  // static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+    this.hidden = false;
+  }
```
Olexandr6, line 11: Pay attention to logical order
```suggestion
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;
    this.hidden = false;
    
    Animal.alive.push(this);
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Animal {
-  // write your code here
+  // static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+    this.hidden = false;
```
Olexandr6, line 10: `hidden` property should be only in Herbivores, not in all animals

Olexandr6 (changes requested):
Let's work on fixes

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  // static alive = [];
```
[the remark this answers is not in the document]
viktorlyt: done

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  constructor(name, health = 100) {
```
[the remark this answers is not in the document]
viktorlyt: done

src/herbivoresAndCarnivores.js
viktorlyt: done

src/herbivoresAndCarnivores.js
viktorlyt: done

src/herbivoresAndCarnivores.js
viktorlyt: done

Olexandr6 (approved):

PR #842, opened 2022-06-23
Solution

Bohdan-Pyschchyk:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,42 @@
...
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    if (this.hidden === true) {
+      this.hidden = false;
+    } else {
+      this.hidden = true;
+    }
```
yanaost, line 26: check checklist 3, you could write this condition in one line, not 4

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,42 @@
...
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (!animal.health) {
+      Animal.alive = Animal.alive.filter(beast => beast.health);
```
yanaost, line 38: you should check if this animal is in the array as it stated in checklist N1

yanaost (changes requested):

yanaost (approved):
nice)

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,38 @@
 
 class Animal {
   // write your code here
+  constructor(name, health = 100) {
```
mykola-koval, line 5: `health` should not be passed to the constructor as it is always 100 when animal is created

mykola-koval (approved):

PR #847, opened 2022-06-23
Solution

oleg-zolkiewskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,39 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
+};
 
 class Carnivore extends Animal {
-  // write your code here
-}
+  bite(animal) {
+    if (!animal.hidden) {
```
vkryvytskyy, line 27: there should be check that animal that has been bitten is Herbivore

vkryvytskyy (changes requested):

ArturErmolenko (approved):

PR #852, opened 2022-06-24
Solution

olya-yevsieienko:


IvanGrabovsky (approved):
good job

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
```
jstmpelowycz, line 4: you don't need `health`, check the examples of how it should work

jstmpelowycz: also, you don't even use it, just remove

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Carnivore || prey.hidden) {
+      return;
+    }
+
+    prey.health -= 50;
+
+    if (prey.health === 0) {
```
jstmpelowycz, line 34:
```suggestion
    if (!prey.health) {
```

jstmpelowycz (changes requested):

mykola-koval (approved):

PR #859, opened 2022-06-24
Develop

stepband:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+      Animal.alive = [];
+    }
+  }
+}
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
vkryvytskyy, line 17: move this to the Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
```
vkryvytskyy, line 29: remove this

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  bite(target) {
+    if (target.hidden === false) {
```
vkryvytskyy, line 32: there should also be `instanceof` check

vkryvytskyy (changes requested):

src/herbivoresAndCarnivores.js
stepband: Why I need the check here?

stepband: I understand now. never mind)

stepband:
I have made changes

PR #862, opened 2022-06-25
solution

SaperZP:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (!prey.hidden && prey instanceof Herbivore) {
+      prey.health -= 50;
+    }
+
+    if (prey.health < 1) {
```
Olexandr6, line 35:
```suggestion
    if (prey.health <= 0) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      Animal.killAnimal(prey);
```
Olexandr6, line 36: You don't need here any other methods, let's just filter `Animal.alive` array.

[Checklist](https://github.com/mate-academy/js_herbivores_and_carnivores/blob/master/checklist.md) the 3rd point

Olexandr6 (commented):
Almost done

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
```
jstmpelowycz, line 15: you don't need `health` here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    super(name, health);
```
jstmpelowycz, line 16: and you have nothing to do with it in your parent class

jstmpelowycz (approved):

PR #869, opened 2022-06-26
Solution

yanaost:


Bohdan-Pyschchyk (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(health, name, hidden) {
+    super(health, name);
```
mykola-koval, line 14:
```suggestion
  constructor() {
    super();
```

you don't actually need those parameters, check it out

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(enemy) {
+    if (!enemy.hidden && enemy.constructor.name === 'Herbivore') {
+      enemy.health -= 50;
+    }
+
+    if (enemy.health === 0 && Animal.alive.indexOf(enemy) !== -1) {
+      Animal.alive = Animal.alive.filter(beast => beast !== enemy);
+    }
```
mykola-koval, line 31: you don't need to use `indexOf` here

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(enemy) {
+    if (!enemy.hidden && enemy.constructor.name === 'Herbivore') {
```
mykola-koval, line 25: maybe use `instanceof` ?

mykola-koval (changes requested):

src/herbivoresAndCarnivores.js
yanaost: yep, tests are ok w/o them
but how it's possible even w/o hidden - new properties for this constructor?

src/herbivoresAndCarnivores.js
yanaost: @mykola-koval it was stated in the checklist that we should check if rabbit is herbivore

mykola-koval (approved):

PR #873, opened 2022-06-27
finished with this zoo

omelika:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
```
mykola-koval, line 16: do you need to pass name to `Animal` class?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
mykola-koval, line 27: it is better to use different variable name, since the animal passed here is not always a herbivore

mykola-koval (approved):

PR #876, opened 2022-06-27
Solution Herbivores and Carnivores

DimaShm:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
```
aliceblunt8, line 16: it should change to the opposite value, if hidden is true, then hide changes it to false and vice versa

> Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
aliceblunt8, line 26: you also need to check whether this animal is herbivore

> The method does not work if it is another сarnivore, or the herbivore is currently hiding

```suggestion
    if (!animal.hidden && another condition) {
```

aliceblunt8 (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false && herbivore instanceof Herbivore) {
```
aliceblunt8, line 26:
```suggestion
    if (!herbivore.hidden && herbivore instanceof Herbivore) {
```

src/herbivoresAndCarnivores.js
aliceblunt8, line 16:
```suggestion
    this.hidden =  !this.hidden;
```

aliceblunt8 (approved):

PR #881, opened 2022-07-03
animals

TanyaHryn:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
ArtemTopolskyi, line 25: what if animal is already hidden

ArtemTopolskyi (approved):

PR #898, opened 2022-07-13
herbivore_carnivore

Hovera-Liubomyr:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(tvar) {
+    if (!tvar.hidden && tvar instanceof Herbivore) {
```
Bohdan-Maksymiuk, line 25: tvar is traryna or tvar?)

Bohdan-Maksymiuk: better use full name, and do not short

Bohdan-Maksymiuk (approved):

PR #900, opened 2022-07-14
Solution v01

Taras-Kozii:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  bite(victim) {
+    const animal = Animal.alive.find(elem => elem === victim);
+
+    if (animal.hidden !== true && !(animal instanceof Carnivore)) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive.splice([Animal.alive.indexOf(animal)], 1);
+    }
```
Bohdan-Maksymiuk, line 39: you could just minus health of victim and then filter Animal (health > 0) instead of such logic

Bohdan-Maksymiuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, hide) {
+    super(name);
+  }
+
+  bite(victim) {
+    if (victim.hidden !== true && !(victim instanceof Carnivore)) {
```
amitpatiljc, line 31: let's optimise code a bit

amitpatiljc, line 31:
```suggestion
     if (!victim.hidden && victim instanceof Herbivore) {```

amitpatiljc (changes requested):
Good Job! just 1 change and good to go!

amitpatiljc (approved):
Well Done!

PR #903, opened 2022-07-17
Solution

Lentyaeev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
```
VETALEK, line 26:
```suggestion
```

looks like we don't want carnivores to hide

VETALEK (changes requested):
Good start 💪

VETALEK (approved):
🔥

PR #909, opened 2022-07-19
Solution

LordAsmodey:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.constructor === Herbivore && animal.hidden === false) {
```
VETALEK, line 25: try using `instanceof` here 😉

VETALEK (changes requested):
GJ 👍 
commented

src/herbivoresAndCarnivores.js
LordAsmodey: I didn't know or forgot about instanceof:) Tnx!

PR #914, opened 2022-07-21
solution

denys-danyliuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
```
illnino380, line 28: You can check `herbivore.hidden` value by using `!not` operator) but your solution looks more obvious)

illnino380 (approved):
Good Job!)

src/herbivoresAndCarnivores.js
denys-danyliuk: good point)

PR #927, opened 2022-07-22
Solution

rodiontseva:


AndrewBessik (approved):
ok:)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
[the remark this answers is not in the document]
rodiontseva: done

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
ivan-kadykalo, line 21: If you call the hide method again your animal will unhide. Are you sure it is right?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    if (target.health === 0) {
```
ivan-kadykalo, line 31: What if there will be less than 0?

ivan-kadykalo (changes requested):

src/herbivoresAndCarnivores.js
rodiontseva: We are deleting animal with health > 0, so we never get into this case. or not?

rodiontseva: if we have some variables for (-50) - ok, I agree with you. But here we have to use directly -50. So it's impossible to go to the case health < 0.

src/herbivoresAndCarnivores.js
rodiontseva: yes, you're right. Animal starts to be unhide. It's not normal. Help me, please, to implement this case using point #3 of our checklist https://github.com/mate-academy/js_herbivores_and_carnivores/blob/master/checklist.md. Thank you!

rodiontseva: is it correct?
  hide() {
    if (this.hidden) {
      this.hidden = this.hidden;
    }
    this.hidden = !this.hidden;
  }

ivan-kadykalo, line 21: That's all))
```suggestion
    this.hidden = true;
```

src/herbivoresAndCarnivores.js
ivan-kadykalo: Good point, you are correct. But for the future, it is safer to use `<= 0`

ivan-kadykalo (approved):

PR #936, opened 2022-07-24
Solution

LovelyDrug:


HennadiiKupr:
GJ

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
ivan-kadykalo, line 21: Do you need to unhide in second call?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(anim) {
+    if (anim instanceof Herbivore && !anim.hidden) {
+      anim.health -= 50;
+    }
+
+    if (anim.health === 0) {
```
ivan-kadykalo, line 31: What will be in case of less than 0?

ivan-kadykalo (changes requested):

src/herbivoresAndCarnivores.js
LovelyDrug: "Create a Herbivore class. Herbivore has a method of hide, which changes the hidden property of the beast to the opposite value and helps to hide from carnivores"
Is it not right? I thought we need just the opposite value?

src/herbivoresAndCarnivores.js
LovelyDrug: Fixed

src/herbivoresAndCarnivores.js
ivan-kadykalo: Good point, but we need just hide herbivores))

ivan-kadykalo (approved):

src/herbivoresAndCarnivores.js
LovelyDrug: Ok, I'll fix it)

PR #957, opened 2022-08-10
Solution by DK

DimaK123:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.type = 'carnivore';
+    delete this.hidden;
+  }
+
+  bite(animal) {
+    if (animal.type === 'herbivore'
+    && animal.hidden !== true) {
```
marina-tilniak, line 35:
```suggestion
    && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.type = 'carnivore';
+    delete this.hidden;
+  }
+
+  bite(animal) {
+    if (animal.type === 'herbivore'
```
marina-tilniak, line 34: use instanceOf here

marina-tilniak (changes requested):

src/herbivoresAndCarnivores.js
DimaK123: All done

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  };
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.type = 'herbivore';
```
marina-tilniak, line 17: why do you need types?

marina-tilniak (changes requested):

src/herbivoresAndCarnivores.js
DimaK123: forgot to delete

marina-tilniak (approved):

PR #959, opened 2022-08-11
solution

frontend112:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+
+    Animal.alive.push(this);
+  }
+
+  bite(victim) {
+    if (victim.hidden === false) {
```
amitpatiljc, line 34: you should add one more condition as to be a victim it should be `instanceof Herbivore`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+      victim.health -= 50;
+
+      if (victim.health === 0) {
```
amitpatiljc, line 37: we just need to remove animal , no need to have true or false condition

amitpatiljc (changes requested):

amitpatiljc (approved):
Good Job!

PR #962, opened 2022-08-13
Task_herbivores-and-carnivores

MarkShapka:


src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,38 @@
...
+    this.hidden = false;
+  };
+
+  hide() {
+    if (this.hidden) {
+      this.hidden = false;
+    } else {
+      this.hidden = true;
+    }
+  };
```
Bohdan-Maksymiuk, line 24: > if the function has a name hide this function should just hide but not toggle

Bohdan-Maksymiuk (changes requested):

PR #964, opened 2022-08-17
Solution

aleksandra-prognimak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
+  }
```
tiserett, line 21:
```suggestion
  hide() {
    this.hidden = !hidden;
  }
```

**changes the hidden property of the beast to the opposite value**

I guees changing to opposite means that you can hide for a period of time, which sets hidden to true, then you can call this method again and stop hiding, which sets hidden to opposite, hidden = !hidden

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herd) {
+    if (herd instanceof Herbivore && herd.hidden === false) {
+      herd.health -= 50;
+    }
+
+    Animal.alive.splice(Animal.alive.indexOf(herd), 1);
+  }
```
tiserett, line 31:
```suggestion
  bite(herd) {
    if (herd instanceof Herbivore && herd.hidden === false) {
      herd.health -= 50;
    }

    Animal.alive.splice(Animal.alive.indexOf(herd), 1);
  }
```

it will delete animal on the first bite, when it's health is still 50, so you should add if statement to check if herd's health is equal to 0, only then you can delete herd from alive array

tiserett (changes requested):
few small fixes required

tiserett (approved):
gj ;)

src/herbivoresAndCarnivores.js
danylo-onishchuk: agree

danylo-onishchuk (changes requested):

danylo-onishchuk (approved):

PR #968, opened 2022-08-18
solution

AndreyJigijalkin:


src/herbivoresAndCarnivores.js
```diff
@@ -28,7 +28,7 @@ class Carnivore extends Animal {
 
   bite(victim) {
     if (victim.__proto__ === Herbivore.prototype
-    && victim.hidden === false) {
+    && !victim.hidden) {
       victim.health -= 50;
     }
     Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```
Dmytro-Matsehora, line 34: if the victim does not die, this code performs an unnecessary action, I suggest adding an `if` construct that checks if the animal has died =)

Dmytro-Matsehora (changes requested):
Hello Andrey, good job, just a small fix I can suggest

Dmytro-Matsehora (approved):
well done 🚀

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super();
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
AndriiRut, line 20: also, you can do the method hidden reusable if you make it to change the hidden value to opposite(optional)

AndriiRut (approved):
🚀

PR #970, opened 2022-08-19
I have a question about 29th row of my code

bohdan-tron:
Can you test my code pls? 
I dont understand what's the point with filter method. Its working like I write in code and like that too:
Animal.alive = Animal.alive.filter((a) => a.health **<=** 0);

And this one is made all tests pass:
Animal.alive = Animal.alive.filter((a) => a.health **<** 0);

How?

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+
+    this.hidden = false;
+  }
+
+  hide() {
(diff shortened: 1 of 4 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
Virass, line 21: `this.hidden = !this.hidden;`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target.hidden === false) {
```
Virass, line 27: I think we need to check the target instance of Herbivore or not
`!target.hidden && target instanceof Herbivore`

Virass (changes requested):

Virass (approved):

PR #971, opened 2022-08-19
Solution

VolodymyrChuyko:


Ohiiko (approved):
Good job.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
+
+      if (prey.health === 0) {
```
AndriiRut, line 29: Do you realy need if() - condition to filter the array?

AndriiRut (changes requested):

src/herbivoresAndCarnivores.js
VolodymyrChuyko: There could be situations when prey.health decreases to 50, not to 0. In this case, filtering is useless. So I decided to add If() - condition to avoid running an unnecessary loop.

AndriiRut: @VolodymyrChuyko you already know what object you changed. fined it inside the array and delete it from the Animal array without copping it by filter)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+        const index = Animal.alive.findIndex(animal => animal === prey);
+
+        Animal.alive.splice(index, 1);
+      }
```
dellachaise, line 33: Try to avoid nesting loops (indexOf + splice), use filter instead - checklist.

dellachaise (approved):
Well done 👍 

Better try to avoid nesting loops (indexOf + splice), use filter instead(checklist).

PR #979, opened 2022-08-25
Solution

Jamal-Amoian:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+      return;
+    }
+
+    findVictim.health -= 50;
+
+    if (findVictim.health === 0) {
+      Animal.alive.splice(indexOfVictim, 1);
+    }
+  }
 }
```
VasylZinchenko, line 41: Possible it's unnecessary to create `const findVictim`  equal to `victim`
You can simplify your code a bit, using `victim` instead of `findVictim` and changing a little the first block `if`

```suggestion
  bite(victim) {
    const indexOfVictim = Animal.alive.indexOf(victim);

    if (victim.hidden === false && victim instanceof Herbivore) {
      victim.health -= 50;
    }

    if (victim.health === 0) {
      Animal.alive.splice(indexOfVictim, 1);
    }
  }
}
```

VasylZinchenko (changes requested):
Good job! 👍

VasylZinchenko (approved):

AndriiRut (approved):
🔥

PR #985, opened 2022-08-25
Solution

Ohiiko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
VolodymyrChuyko, line 21: There is a checklist requirement: if the function has a name hide this function should just hide but not toggle

VolodymyrChuyko (changes requested):
Good job. Just small issues

VolodymyrChuyko (approved):
Looks good to me!

danylo-onishchuk (approved):

PR #987, opened 2022-08-26
Solution

DimaKobzar7:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
VolodymyrChuyko, line 6:
```suggestion
  constructor(name) {
    this.name = name;
    this.health = 100;
```
health value is not a passed as a constructor parameter

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
-  // write your code here
+  bite(target) {
+    const indexOfTarget = Animal.alive.indexOf(target);
+
+    if (target.hidden === false) {
+      target.health -= 50;
+    }
+
+    if (target.health === 0) {
+      Animal.alive.splice(indexOfTarget, 1);
```
VolodymyrChuyko, line 34: Try to avoid nesting loops (indexOf + splice), use filter instead

VolodymyrChuyko (approved):

src/herbivoresAndCarnivores.js
dellachaise: if this animal already died, you anyway delete the last animal from the array. indexOf return -1 if the animal already died, and you will splice the last animal - splice(-1, 1)

you have this point in the checklist )

dellachaise (approved):
Good job 👍 

Do not forget to add changes according to comment 👇

PR #988, opened 2022-08-26
Solution

kir8mir:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
tiserett, line 15:
```suggestion
    Animal.alive.push(this);
```

you can do this inside parent constructor (Animal)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.name = name;
+    this.health = health;
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
```
tiserett, line 13:
```suggestion
  constructor(name) {
    super(name);
```

you don't need health parametr here, because it has default value in parent constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(obj) {
+    if (obj instanceof Herbivore && obj.health >= 50 && obj.hidden === false) {
```
tiserett, line 30:
```suggestion
    if (obj instanceof Herbivore && obj.hidden === false) {
```

if creature's health < 50, it will be deleted from array, that's why you don't need this check inside 'if' statement

tiserett (changes requested):
few changes required

tiserett (approved):
gj ;)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
```
AlinaMatuschak, line 26: Unnecessary, you do the same on Animal class

AlinaMatuschak (approved):

PR #991, opened 2022-08-27
created herbivores and carnivores

Anatolii-Bukoros:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(health, name, hidden = false) {
+    super(health, name);
+
+    this.hidden = hidden;
+  }
```
Andrii-Stiebalov, line 18: You are trying to provide value in the func. constructor that it does not expect. You don't need to provide `health`.
image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
-  // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore && herb.hidden === false) {
+      herb.health -= 50;
+    }
+
+    if (herb.health === 0) {
+      Animal.alive.pop();
+    }
+  }
```
Andrii-Stiebalov, line 33: Dead `herb` will not always be at the **end** of the array. Replace its method `pop` with `splice` 😉

Andrii-Stiebalov (changes requested):
GJ, but needs a fix )

src/herbivoresAndCarnivores.js
Anatolii-Bukoros: i thought there was no point in biting an animal that had already died =))

Andrii-Stiebalov (approved):
Good job!!! 🚀👏

AndriiRut (approved):
🚀

PR #994, opened 2022-08-28
Solution

ostUp:


Anastasiia-Khmelovska (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.filter(a => a.health !== 0);
```
AndriiRut, line 32:
```suggestion
      Animal.alive = Animal.alive.filter(creating => creating.health !== 0);
```

better use indexOf() + splice() instead of filter)

AndriiRut (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
dellachaise, line 21: Method hide should change hidden property to the opposite value, not always to the true.

src/herbivoresAndCarnivores.js
dellachaise, line 32: Your variables should be meaningful. Instead of `a` try to use `animal` or `creature`.

dellachaise (approved):
Good job 👍 

But do not forget to add changes according to comments 👇

PR #996, opened 2022-08-28
solution

Maxim-Gumeniuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,40 +1,37 @@
...
 
 class Herbivore extends Animal {
   constructor(name) {
     super(name);
-
     this.hidden = false;
   }
-
   hide() {
     this.hidden = true;
```
Mejeezz, line 19: 3rd point of checklist.  I guess it must be changed to !this.hidden

src/herbivoresAndCarnivores.js
```diff
@@ -1,40 +1,37 @@
...
+  bite(animal) {
+    const indexOfAnimal = Animal.alive.indexOf(animal);
 
-    if (victim.hidden === true || victim instanceof Carnivore) {
+    if (animal.hidden === true || animal instanceof Carnivore) {
       return;
     }
 
-    victim.health -= 50;
+    animal.health -= 50;
```
Mejeezz, line 31: Have some doubts about this option of code writing, but assumt this code is fine too.

Mejeezz (changes requested):

Mejeezz (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    const indexOfAnimal = Animal.alive.indexOf(animal);
+
+    if (animal.hidden === true || animal instanceof Carnivore) {
+      return;
+    }
```
dellachaise, line 29: try to rewrite this condition to avoid unnecessary `return` - write code for `bite` case.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+    animal.health -= 50;
+
+    if (animal.health === 0) {
+      Animal.alive.splice(indexOfAnimal, 2);
+    }
```
dellachaise, line 35: Try to avoid nesting loops (indexOf + splice), use filter instead - cehcklist.

dellachaise (changes requested):
Good job 👍 

Just add changes according to comments 👇

dellachaise (approved):
Well done 👍

PR #997, opened 2022-08-28
Solution

Mejeezz:


Maxim-Gumeniuk (approved):
GJ!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(creature => creature.health > 0);
```
AndriiRut, line 35: why do you need to check all array instead of find an index of your Victim and splice it from the array?)

AndriiRut (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= 50;
+    }
+
+    const deadAnimal = Animal.alive.indexOf(victim);
+
+    if (victim.health === 0) {
+      Animal.alive.splice(deadAnimal, 1)
+    }
```
dellachaise, line 39: Try to avoid nesting loops (indexOf + splice), use filter instead - checklist.

dellachaise (approved):
Well done 👍 

Would be grate to rewrite it using `filter`.

src/herbivoresAndCarnivores.js
Mejeezz: 2 mentors - 2 options))  I wrote it with filter at first, but AndriiRut asked me to change it )))))

PR #999, opened 2022-08-28
Herbivores and carnivores solution

ibondrk:


OlgaOchichenko (approved):
Good Job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
dellachaise, line 20: Method hide should change hidden property to the opposite value, not always to the true.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    if (creature instanceof Herbivore
+      && creature.hidden === false) {
+      creature.health -= 50;
+    }
+
+    if (creature.health === 0) {
+      const isAnimalDead = Animal.alive.indexOf(creature);
+
+      Animal.alive.splice(isAnimalDead, 1);
+    }
```
dellachaise, line 35: Try to avoid nesting loops (indexOf + splice), use filter instead - checklist.

dellachaise (changes requested):
Great start 👍

Just add fixes according to comments 👇

src/herbivoresAndCarnivores.js
ibondrk: In a third point of checklist. "this function should just hide but not toggle". But i changed it.

dellachaise: Good comment 👍  Truly checklist disagrees with the task description. 
I will keep that in mind 👌

dellachaise (approved):
Well done 👍

PR #1001, opened 2022-08-28
Dev

Anton-Liada:


max-kravchenko (approved):
Great solution!

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
AlinaMatuschak, line 24: hidden value is already boolean 

```suggestion
    if (!animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,36 @@
...
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(({ health }) => health > 0);
+  }
 }
 
+Animal.alive = [];
+
```
AlinaMatuschak, line 33: it should be near Animal class

AlinaMatuschak (approved):

src/herbivoresAndCarnivores.js
Anton-Liada: I wrote it because I thought it was easier to read 😔 and with the exclamation mark, the tests do not pass

PR #1002, opened 2022-08-28
solution 1

kostya-kosyuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    Animal.alive.push(this);
+  }
+
+  death() {
+    Animal.alive.splice(Animal.alive.indexOf(this), 1);
```
andrii-dmytruk, line 14: Avoid nesting loops. Use filter instead of splice(indexOf)

andrii-dmytruk (changes requested):

andrii-dmytruk (approved):

src/herbivoresAndCarnivores.js
kostya-kosyuk: @andrii-dmytruk, Thank you, dude

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
```
dellachaise, line 21: Method hide should change hidden property to the opposite value, not always to the true.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    if (target.health <= 0) {
+      Animal.alive = Animal.alive.filter(({ health }) => health > 0);
```
dellachaise, line 32: Destructuring 🔥

dellachaise (approved):
Great job 👍 

Just make sure, that method hide should change hidden property to the opposite value, not always to the true.

PR #1006, opened 2022-08-29
Animal, Herbivore and Carnivore classes implemented

andrii-dmytruk:


kostya-kosyuk (approved):
gj

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
+    }
```
dellachaise, line 10: Why you can`t simply add?
```suggestion
      Animal.alive = [];
```

Why do you need this check?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
dellachaise, line 22: Method hide should change hidden property to the opposite value, not always to the true.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (!(prey instanceof Carnivore) && !prey.hidden) {
+      prey.health -= 50;
+
+      if (prey.health <= 0) {
+        Animal.alive = Animal.alive.filter(el => el !== prey);
```
dellachaise, line 32: Better use meaningful variable names, try to use `animal` or `creature` - for example, instead of `el`.

dellachaise (changes requested):
Great start 👍 

Just add fixes according to comments 👇

src/herbivoresAndCarnivores.js
andrii-dmytruk: [FUNCTION NAMING] - if the function has a name hide this function should just hide but not toggle

src/herbivoresAndCarnivores.js
andrii-dmytruk: Because if i do it like:

`   Animal.alive = [];
    Animal.alive.push(this);`

It will create empty array every time when constructor is called

andrii-dmytruk: it could be better to init `alive` using `static`, but then i had to disable linter

src/herbivoresAndCarnivores.js
dellachaise: Good comment 👍 Truly checklist disagrees with the task description.
I will keep that in mind 👌

src/herbivoresAndCarnivores.js
dellachaise: Ok. I understand. With `static` it truly would be better. You have problems with linter if you add
`static alive = [];` in line 4, before constructor - too?

dellachaise (approved):
Well done 👍

src/herbivoresAndCarnivores.js
andrii-dmytruk: Yes, it throws exception "Unexpected token ="

PR #1008, opened 2022-08-29
solution of herbivores and carnivores

Alzay007:


Desertiks (approved):
Good work 😄

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+
+    Animal.alive.push(this);
```
AlinaMatuschak, line 19: Don’t need to push here animal, you already do it in Animal class

AlinaMatuschak (approved):
👌

PR #1017, opened 2022-08-29
Herbivores and Сarnivores task solution

anyakrainiuk:


roma-shafihullin (approved):
good solution 💪

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
+      animal.health -= 50;
+    }
```
dellachaise, line 27: Please read task one more time. As I remember, you should also check isn't this exemplary Carnivor.

dellachaise (approved):
Well done 👍 

Just add one more condition according to comment 👇

src/herbivoresAndCarnivores.js
anyakrainiuk: hidden exists only in herbivores, I thought I don't have to check type of animal.

PR #1020, opened 2022-08-29
Herbivores and Сarnivores solution

vladgusakovskiy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
Jamal-Amoian, line 4: You don't need to send in every constructor parametr "health = 100", because
1. You don't send this parametr in arguments
2.If you write just `this.health = 100`, you can write this once, in Animal class, and doesn't rewrite this else

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
Jamal-Amoian, line 14: you don't send parametr hidden, you should just write `this.hidden = false`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
Jamal-Amoian, line 17: As i remembered, you can not push this, if you leave it only in class Animal, it will work too

Jamal-Amoian (changes requested):

src/herbivoresAndCarnivores.js
vladgusakovskiy: Yeah, you are right!

vladgusakovskiy: @Jamal-Amoian Yeah, you are right!

Jamal-Amoian (approved):
Great Job!🔥🔥🔥

dellachaise (approved):
Well done 👍

PR #1021, opened 2022-08-29
solutionHerbivoresAndCarnivores

YuriiPronin:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
ibondrk, line 6: You can just write:
```
this.health = 100;
```
and remove health from constructor parameters.

ibondrk (approved):
Good job, and it'll become better if u add this changes)

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim.hidden === false) {
+      victim.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter((indicatorOfLife) =>
+      indicatorOfLife.health > 0
+    );
```
AlinaMatuschak, line 32: Don’t need to run the filter every time on run bite. You can do it just if health is 0

AlinaMatuschak (approved):

PR #1023, opened 2022-08-29
animal, carnivore and herbivore classes with methods were created

roma-shafihullin:


SvyatoslavYatsenko (approved):
Good solution

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden !== true && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(health => health > 0);
```
AlinaMatuschak, line 30: Don’t need every-time filter alive animals. Just if the current instance has <= 0 health

AlinaMatuschak (approved):

PR #1030, opened 2022-08-29
solution

yevhenii-podplietnii:
I wrote Animal.alive = static;
because ESlint did not allow to write 
static alive = [];
image

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,33 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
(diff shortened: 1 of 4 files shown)


lizapolokhina (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(({ health }) => health > 0);
```
AlinaMatuschak, line 29: Don’t need every-time filter alive animals. Just if the current instance has <= 0 health

AlinaMatuschak (approved):

src/herbivoresAndCarnivores.js
yevhenii-podplietnii: 🚀

PR #1035, opened 2022-08-30
Herbivores and Carnivores created

Reptiloid044:
Hi!

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(animal) {
+    this.name = animal;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(animal, health) {
+    super(animal, health);
+    this.hidden = false;
+  }
+
+  hide() {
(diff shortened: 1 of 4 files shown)


plemiannik:
> Hi!

hi

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(predator) {
```
plemiannik, line 26: why predator?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    if (predator.hidden === false) {
+      predator.health -= 50;
+    }
+
+    if (predator.health === 0) {
```
plemiannik, line 31: good solution 4 my point of view, cos array will filtering only when animal health is 0, not after every bite 👍

plemiannik (approved):
✈️

src/herbivoresAndCarnivores.js
Reptiloid044: Yeah, you`re right,
Looks not correct in this case, so I have changed it in the correct way.
And I have noticed your solution with a health default value, I have taken your method 🤫

AlinaMatuschak (approved):
👍

PR #1042, opened 2022-09-04
Solution

MihailKostov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(food) {
```
Virass, line 24: Named target or victim

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    if (food.hidden === false) {
+      food.health -= 50;
+    }
+
+    if (food.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(food));
```
Virass, line 30: Use filter method

Virass (changes requested):

SanyaBratashchuk (approved):

PR #1057, opened 2022-09-19
Develop

grafit-off:
feat: Animals classes were implemented

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -2,14 +2,44 @@
 
 class Animal {
   // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static kill() {
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
+  }
 }
+
+Animal.alive = [];
 
 class Herbivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+
+      if (herbivore.health === 0) {
+        Animal.kill();
+      }
+    }
+  }
 }
 
 module.exports = {


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,44 @@
 
 class Animal {
   // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
TheVovchik, line 7: think we can declare it not in constructor

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,44 @@
...
+
+    Animal.alive.push(this);
+  }
+
+  static updateAliveList() {
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
TheVovchik, line 13: don't think we need a static method. we can use this lile in bite method

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,44 @@
...
+  }
 }
 
+Animal.alive = [];
```
TheVovchik, line 17: I declare it as static object in class definition, linter is crying but in vsCode it's works

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,44 @@
...
+
 class Herbivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
```
TheVovchik, line 24: we can declare this one in class, so constructor will be redundant

TheVovchik (approved):
GJ 🔥 Leave some comments

src/herbivoresAndCarnivores.js
grafit-off: Maybe yes, but I like to visually separate the methods and properties. I don't know is it a mistake or not, it's just looks more understandable for me

src/herbivoresAndCarnivores.js
grafit-off: I prefer to separate this logic

src/herbivoresAndCarnivores.js
grafit-off: My linter didn't let me do it :(

src/herbivoresAndCarnivores.js
TheVovchik: don't sure about difference, but Misha in his video use it. probaly you rigth and this have no difference. for me it sounds good practice to get const value from constructor

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,44 @@
...
 class Animal {
   // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static updateAliveList() {
```
Andr1yK, line 12: 🔥

Andr1yK (approved):

PR #1071, opened 2022-09-26
Solution

AntonVaida:


xiimiik (commented):
GJ 🔥

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (!beast.hidden && beast instanceof Carnivore === false) {
```
naziamoff, line 31: Looks more logical this way
```suggestion
    if (beast instanceof Herbivore && !beast.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+      beast.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0)
```
naziamoff, line 34: now you filter the array after every bite. In order to optimize do it only when the bitten prey died

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
+/* eslint-disable */
 
 class Animal {
+  static newBeast(contecst) {
+    this.alive.push(contecst)
```
naziamoff, line 6: what is contecst?

naziamoff (changes requested):
Fix comments below and add some spacing between methods to make it more readable

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
+/* eslint-disable */
 
 class Animal {
-  // write your code here
+  static newBeast(newAnimal) {
+    this.alive.push(newAnimal)
+  }
```
Anna-Tkachenko, line 7: this method is redundant, you can simply use push after setting needed properties in constructor

Anna-Tkachenko (approved):

PR #1072, opened 2022-09-27
All working

NazikVr:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
```
qoa11a, line 27: It's better to check if animal.health is less than or equals to zero.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+        Animal.alive.splice(animal);
```
qoa11a, line 28: It won't remove the dead animal from an array. Think how to do it correctly.

qoa11a (changes requested):

OleziO (approved):

PR #1074, opened 2022-09-28
Solution

Shepeha-Andriy:


Veremchukk (commented):
Well done

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.health = 100;
+    this.name = name;
+
+    Animal.alive.push(this);
+  }
+
+  static died(target) {
+    const index = this.alive.indexOf(target);
+
+    this.alive.splice(index, 1);
```
Anna-Tkachenko, line 14: let's do it using one array iteration instead of 2

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
Anna-Tkachenko, line 28: check checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #1075, opened 2022-09-28
Herbivores and Сarnivores

Vladosik1337:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = !this.hidden;
```
imaksweb, line 24: I'm not sure that `hide` method should toggle hidden property, I think it should set `true`

imaksweb (approved):
Well done ⭐

denys-cheporniuk (approved):

PR #1080, opened 2022-09-28
solution

fatihsamur:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static removeDeads(dead) {
+    Animal.alive = Animal.alive.filter(animal => animal.name !== dead.name);
+  };
```
kamal-tufekcic, line 6: can be removed from here as used immediately in the condition below

kamal-tufekcic (approved):

ihor-jpeg (approved):

ihor-jpeg (approved):

PR #1081, opened 2022-09-28
solution

artem-nasadchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
imaksweb, line 32: To avoid unnecessary filtering of `alive` array it's better to filter this array only if animal health = 0.

imaksweb (approved):
GJ 🔥

src/herbivoresAndCarnivores.js
artem-nasadchuk: agree, will fix it

Denchepornyuk (approved):

PR #1082, opened 2022-09-28
No animals were harmed while making this feature

Yevheniya-Deryka:
No animals were harmed while making this feature )

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,16 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,38 @@
...
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+
+      Animal.alive = Animal.alive
+        .filter(animal => animal.health > 0);
+    }
+  }
 }
 
+Animal.alive = [];
```
ArturNagaycev, line 34: it would be better if this property is closer to the class Animal ;)

ArturNagaycev (approved):
i liked it

Andr1yK (approved):

PR #1084, opened 2022-09-28
Solution

anastasiia-nahaiska:


jojnajbude (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static deleteDeadAnimal(animal) {
+    this.alive = this.alive.filter(name => name !== animal);
```
OleksiiBlezniuk, line 5: but it's an array of objects, can name be an object?
just remove animals with health less or equal 0, why to compare objects?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
OleksiiBlezniuk, line 26: How can this animal stop hiding? 🙂 Better to change `hidden` value to opposite here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
```
OleksiiBlezniuk, line 33: As for me is more logical to first check if this animal is an instance of needed class, and the check it's `hidden` value

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
OleksiiBlezniuk: I was wrong, ignore this comment 🙃

OleksiiBlezniuk (approved):
👍

PR #1085, opened 2022-09-28
Solution

Andrew-Skoryk:
Linter is trolling...

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,33 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && prey.hidden === false) {
+      prey.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(({ health }) => health > 0);
```
imaksweb, line 29: To avoid unnecessary filtering of alive array it's better to filter this array only if animal health <= 0.

imaksweb (approved):
Well done 🔥

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && prey.hidden === false) {
```
naziamoff, line 26: Better to use shortcuts for boolean variables
```suggestion
    if (prey instanceof Herbivore && !prey.hidden) {
```

naziamoff (approved):
Well done 🔥 
Just one small comment

PR #1089, opened 2022-09-29
Solution

Yeliseiev:


Maksym845 (approved):
I didn`t find any mistakes.
Good job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
naziamoff, line 7: should all animals have `hidden` method?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    Animal.addAnimal(this);
+  }
+
+  static addAnimal(animal) {
+    this.alive.push(animal);
+  }
+
+  static delAnimal() {
```
naziamoff, line 15: `delAnimal` looks not clear enough as a name for the method

naziamoff (changes requested):

src/herbivoresAndCarnivores.js
Yeliseiev: If we are talking about animals. Carnivores hide no worse than herbivores))

Anna-Tkachenko (approved):
Good job!

naziamoff (approved):

PR #1090, opened 2022-09-29
solution

vmamchur:


vmamchur:
> I remember Denys was saying that you better not to use delete for an array. Because it can cause problems for future loops or iteration methods. Maybe you could use splice or even filter as it stated in checklist.

you're right, hadn't thought about that.
i'll fix that soon. thanks!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
OleksiiBlezniuk, line 22: How can this animal stop hiding? 🙂 Better to change `hidden` value to opposite here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (!target.hidden && target instanceof Herbivore) {
+      target.health -= 50;
+    }
```
OleksiiBlezniuk, line 29: Actually it's a strange condition
Better to check instance first (if this object can have `hidden` property) and then check this property value

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target.hasOwnProperty('hidden') && !target.hidden) {
+      target.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
OleksiiBlezniuk, line 31: Let's add check here, there is no need to filter array if our `target.health` is greater then 0

OleksiiBlezniuk (changes requested):

PR #1091, opened 2022-09-29
Task was solved

DenysKleimenov:


ihorandrianov (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(herb) {
+    if (herb.hidden === false && !(herb instanceof Carnivore)) {
+      herb.health -= 50;
+    }
```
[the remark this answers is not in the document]
OleksiiBlezniuk: It's more logical, we have only two types of animal here, so we can just check if our animal is instance of needed class
Also if we first checks it's hidden property, then why do we need to check if it's needed type of animal (if only one of types can have this property)?

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+
+    if (Animal.alive.some(key => key.health <= 0)) {
+      Animal.alive = Animal.alive.filter(anim => anim.health > 0);
```
OleksiiBlezniuk, line 34: What is `anim`? Avoid that words shorting

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  }
+
+  bite(herb) {
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
+    }
+
+    if (Animal.alive.some(key => key.health <= 0)) {
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
+    }
```
OleksiiBlezniuk, line 35: Ok, but this `some` is unnecessary, just check `herb.health`

OleksiiBlezniuk (approved):

PR #1092, opened 2022-09-29
solution of Herbivores and Сarnivores task

Dikey945:


volodymyrklymkiv (approved):
awesome

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,47 @@
...
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static died(animal) {
+    const diedAnimalIndex = this.alive.indexOf(animal);
+
+    this.alive.splice(diedAnimalIndex, 1);
+  }
```
OleksiiBlezniuk, line 16: Better to avoid array mutation, use filter instead

OleksiiBlezniuk: Also it will help you to iterate array only once

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,47 @@
...
 class Herbivore extends Animal {
   // write your code here
-}
+  constructor() {
+    super();
+    this.hidden = false;
+  };
+
+  hide() {
+    this.hidden = true;
```
OleksiiBlezniuk, line 29: And how can this animal stop hiding? 🙂 Better change `hidden` to opposite in this method

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,47 @@
...
+  }
+};
 
 class Carnivore extends Animal {
   // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore.hasOwnProperty('hidden')) {
```
OleksiiBlezniuk, line 36: Instead of `hasOwnProperty` you can check if object is instance of needed class by `instanceof`

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,47 @@
...
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Array.from(new Set(Animal.alive));
```
OleksiiBlezniuk, line 40: Why do you need this?

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,47 @@
...
+
+    if (herbivore.health === 0) {
+      Animal.died(herbivore);
+    }
```
OleksiiBlezniuk, line 44: Better to check if it's health smaller or equal 0, we can change carnivore damage, or animal health and it shouldn't lead us to refactoring because of that change

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
OleksiiBlezniuk: I was wrong, ignore this comment 🙃

OleksiiBlezniuk (approved):

PR #1095, opened 2022-09-29
solution

artem-musii:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    Animal.alive.push(this);
+  }
+
+  bite(animalToBite) {
+    if (animalToBite instanceof Herbivore && !animalToBite.hidden) {
+      animalToBite.health -= 50;
+
+      if (animalToBite.health === 0) {
+        Animal.alive.splice(Animal.alive.indexOf(animalToBite), 1);
+      }
```
myroslav-diiak, line 39: I think filter method will be better here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+
+    Animal.alive.push(this);
```
myroslav-diiak, line 18: You can do Animal.alive.push() once in Animal constructor

myroslav-diiak (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animalToBite) {
+    if (animalToBite instanceof Herbivore && !animalToBite.hidden) {
+      animalToBite.health -= 50;
+
+      if (animalToBite.health === 0) {
```
OleksiiBlezniuk, line 31: Better to check if `health` is less or equal to 0
We may change animal health or carnivore damage and it shouldn't require any refactoring

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+        Animal.alive = Animal.alive.filter(({ health }) => health !== 0);
+      }
```
OleksiiBlezniuk, line 33: Also there is no need to nest this `if` blocks, but ok, it's not critical here

OleksiiBlezniuk (approved):

PR #1096, opened 2022-09-29
Solution

Apollonia5:


glugovsky (approved):
Well done 👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  };
+
+  static died(animal) {
```
naziamoff, line 10: think of better name for this method.
All the function names should describe us what they're doing. usually it they start with `get...`, `set...` `calculate...` and so on

naziamoff (approved):
Well done! But pay attention to the naming

PR #1097, opened 2022-09-29
solution

nikitaovch99:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,28 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
vmamchur, line 7: we never get health as an argument, it'd be better to move this inside the constructor

vmamchur (approved):
GJ

src/herbivoresAndCarnivores.js
OleksiiBlezniuk: It was ok too 🙂

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,28 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
OleksiiBlezniuk, line 8: Can `Carnivore` has this property? I think no, also task said that this property should be in `Herbivore` class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,28 @@
...
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
```
OleksiiBlezniuk, line 16: How can this animal stop hiding? 🙂 Better to change `hidden` value to opposite here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,28 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(alive => alive.health > 0);
```
OleksiiBlezniuk, line 24: No need to filter array if this animal have more than 0 health, let's add check here

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
OleksiiBlezniuk: I was wrong, ignore this comment 🙃

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,30 @@
...
+    this.hidden = true;
+  }
 }
-
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+      Animal.alive = Animal.alive.filter(alive => alive.health > 0);
```
OleksiiBlezniuk, line 25: Ok, now it's a bit better, but still this array will be filter even if animal is not dead yet 🙂 Just check if this animal has 0 or less health and perform filter only in that case

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

PR #1098, opened 2022-09-29
solution to herbivores and carnivores

dsydorenko534:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super();
+  }
```
iPanda177, line 27: You don't need this at all :)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  bite(herbivore) {
+    if (!herbivore.hidden && Herbivore.prototype === herbivore.__proto__) {
```
iPanda177, line 30:
```suggestion
    if (!herbivore.hidden && herbivore instanceof Herbivore) {
```

iPanda177: i guess it's more accurate

iPanda177 (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
naziamoff, line 25: We are not sure whether it is an herbivore here, we check it inside this method
```suggestion
  bite(animal) {
```

naziamoff (approved):
Good job!

PR #1099, opened 2022-09-29
solution was added

Maksym845:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Carnivore !== true && target.hidden !== true) {
+      target.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
Yeliseiev, line 30: I think it is better to add a condition for this filtering so that it is not performed every time a bite() is made

Yeliseiev (approved):
Good job!

src/herbivoresAndCarnivores.js
naziamoff: I totally agree here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Carnivore !== true && target.hidden !== true) {
```
naziamoff, line 26: We can simplify it a bit 😄 
```suggestion
    if (target instanceof Herbivore && !target.hidden) {
```

naziamoff (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1101, opened 2022-09-29
Develop

anton-iskryk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Carnivore === false && !herbivore.hidden) {
```
artem-nasadchuk, line 27: maybe better just Carnivore === false => Herbivore?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
+  }
```
artem-nasadchuk, line 32: To avoid unnecessary filtering of alive array it's better to filter this array only if animal health = 0.

artem-nasadchuk (approved):

src/herbivoresAndCarnivores.js
anton-iskryk: okay, thanks, I fixed it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
```
OleksiiBlezniuk, line 18:
```suggestion
  constructor(name) {
    super(name);
    
    this.hidden = false;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
```
OleksiiBlezniuk, line 22: So how can this animal stop hiding? 🙂

OleksiiBlezniuk: Let's just set value to opposite here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health === 0);
```
OleksiiBlezniuk, line 31: 1 - no need to filter animals every time
2 - better to check if animal health is smaller or equal 0, if we'll change herbivore damage (or animals health) to avoid logic change

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
OleksiiBlezniuk: Ignore this comment 🙃

OleksiiBlezniuk (approved):

PR #1103, opened 2022-09-29
Herbivores and Carnivores

iam-datsenko:


iPanda177 (approved):
GJ!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health) {
```
silchyn, line 6: Why you need that unused 'health' parameter?

silchyn (changes requested):

silchyn (approved):
Cool!

PR #1104, opened 2022-09-29
Herbivores and carnivores

glugovsky:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  static removeDeadAnimal() {
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
```
Apollonia5, line 20: You don't need to use  'name' here.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target.hidden === false) {
```
Apollonia5, line 32: `    if (animal instanceof Herbivore && !animal.hidden) {`

Apollonia5: we need to check if target is herbivore

glugovsky: But it is only herbivores that have a hidden property, there is no other class that could have this property being equal to false

PR #1105, opened 2022-09-30
Solution

ArtemHulkoDev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    const currentAnimal = Animal.alive.find(animal => animal === herbivore);
+
+    if (currentAnimal.hidden !== undefined && currentAnimal.hidden !== true) {
+      currentAnimal.health -= 50;
+    }
+
+    if (currentAnimal.health <= 0) {
+      Animal.alive = Animal.alive.filter(animal => animal !== currentAnimal);
+    }
+  }
```
glugovsky, line 34: Not sure why would you create `currentAnimal`. `herbivore` is already the current animal, there is no need for extra variable. Everything should work with simply using the `herbivore` parameter instead of every instance of `currentAnimal`
```suggestion
    if (herbivore.hidden !== undefined && herbivore.hidden !== true) {
      herbivore.health -= 50;
    }

    if (herbivore.health <= 0) {
      Animal.alive = Animal.alive.filter(animal => animal !== herbivore);
    }
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
```
OleksiiBlezniuk, line 9:
```suggestion
  static alive = [];
  
  constructor(name) {
    this.name = name;
    this.health = 100;
    Animal.alive.push(this);
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
```
OleksiiBlezniuk, line 16:
```suggestion
  constructor(name) {
    super(name);
    
    this.hidden = false;
  }
```

src/herbivoresAndCarnivores.js
OleksiiBlezniuk: Agree with @glugovsky

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    const currentAnimal = Animal.alive.find(animal => animal === herbivore);
```
OleksiiBlezniuk, line 25: Why do you need to find this object in `alive` array if you already have this object? 🙂

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+    if (currentAnimal.hidden !== undefined && currentAnimal.hidden !== true) {
+      currentAnimal.health -= 50;
+    }
```
OleksiiBlezniuk, line 29: This action should happen only if `hidden` is true, actually there is no need to check if it's not undefined or not true, simplify this condition. Also to check if this object is an instance of `Herbivore` better to use `instanceof`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+    if (currentAnimal.health <= 0) {
+      Animal.alive = Animal.alive.filter(animal => animal !== currentAnimal);
+    }
```
OleksiiBlezniuk, line 33: Ok, but object comparing in this case looks strange, as for me better to filter animals by `health` property

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
OleksiiBlezniuk: Oh, no, I was wrong here, sorry, just check if this object is instance of `Herbivore` and if `hidden` is `false` here

src/herbivoresAndCarnivores.js
ArtemHulkoDev: yeah, I see that there was no reason to do it, fixed, thanks)

src/herbivoresAndCarnivores.js
ArtemHulkoDev: done)

src/herbivoresAndCarnivores.js
ArtemHulkoDev: done)

silchyn (approved):
Cool!

src/herbivoresAndCarnivores.js
ArtemHulkoDev: done)

PR #1111, opened 2022-10-01
Solution

oleksiy-uchaev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
SanyaBratashchuk, line 30:
```suggestion
    if (animal.health <= 0) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+      const indexToDell = Animal.alive.indexOf(animal);
+
+      Animal.alive.splice(indexToDell);
```
SanyaBratashchuk, line 33: use filter

SanyaBratashchuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
Bohdan-Maksymiuk, line 21: [FUNCTION NAMING] - if the function has a name hide this function should just hide but not toggle

Bohdan-Maksymiuk (changes requested):
Good job - pay attention to checklist

Bohdan-Maksymiuk (approved):

PR #1114, opened 2022-10-01
Solution

Oleksandra-Shevchenko:


poznianski (approved):
good, but maybe hidden property shouldn't go to global Animal class, coz only herbivore can have it

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
...
+  bite(animal) {
+    if (animal instanceof Carnivore) {
+      return;
+    }
+
+    if (animal.hidden === true) {
+      return;
+    }
+
+    animal.health -= 50;
```
Anna-Tkachenko, line 33: let's avoid using return statement here
let's just add the correct condition, and do some actions if condition = true

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
...
+
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.filter(el => el.health > 0);
```
Anna-Tkachenko, line 36: what is el?

Anna-Tkachenko (changes requested):

PR #1115, opened 2022-10-01
Solution

oleksandr-korobko:


MAKSOti (approved):
GJ)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(item => item.health > 0);
```
OleksiiBlezniuk, line 31: No need to filter array every time, just do it, when `animal` has 0 or less health (if no one is dead, why filter array?)

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
oleksandr-korobko: Done)

OleksiiBlezniuk: Then push commit with your changes 🙂

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
OleksiiBlezniuk, line 31: Better to check if health is less or equal to 0, then just equality
It works good now, but in case if we'll want to change animals health, or carnivores damage check will already work good for that cases

PR #1116, opened 2022-10-01
Solution

korniienkokarina:


Rina07 (approved):
LGTM :star:

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,48 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
+
+  hide() {
+    this.hidden = true;
```
OleksiiBlezniuk, line 18: How can this animal stop hiding? 🙂 Better to change `hidden` value to opposite here

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,48 @@
...
+
+      if (animal.health === 0) {
+        const index = Animal.alive
+          .findIndex(someAnimal => someAnimal === animal);
+
+        if (index >= 0) {
+          Animal.alive.splice(index, 1);
+        }
+      }
+    }
```
OleksiiBlezniuk, line 40: 3 nested if blocks😵

Let's make it simpler =)

Also better to avoid objects comparing, let's just filter array and remove animals with health less or equal 0 (better to check this condition instead of just equality, 'cause we can change animal health or carnivore damage and it shouldn't require any refactor)

src/herbivoresAndCarnivores.js
OleksiiBlezniuk: I was wrong, ignore this comment 🙃

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      Animal.alive = Animal.alive.filter(someAnimal => someAnimal.health > 0);
```
OleksiiBlezniuk, line 32: So if our animal will have 50 health we'll filter our array? It's unnecessary, do it only when you know that this animal have 0 health or less

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive = Animal.alive.filter(someAnimal => someAnimal.health > 0);
+      }
+    }
```
OleksiiBlezniuk, line 35:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
      animal.health -= 50;
    }
    
    if (animal.health <= 0) {
      Animal.alive = Animal.alive.filter(someAnimal => someAnimal.health > 0);
    }
```

PR #1118, opened 2022-10-02
Solution

dmytristepanov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+
+      Animal.alive = Animal.alive
+        .filter(animal => animal.health > 0);
```
MaxManis, line 31: you shoud use `static` metod `died `inside of ur main class for this logic

MaxManis (changes requested):
GJ, just one thing to go!)
Move you logic of animal dieing into` static died` method

MaxManis (approved):
Good job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static died(animal) {
```
Anna-Tkachenko, line 11: method should have action name
what is died? what this function should do?

Anna-Tkachenko (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static ifBeastDies(animal) {
```
Anna-Tkachenko, line 11: function name should be the action
ifBeastDies - good name for some boolean variable, not function

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #1119, opened 2022-10-02
Solution

MAKSOti:


oleksandr-korobko (approved):
Good Job!)

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
+    }
+
+    if (prey.health === 0) {
+      Animal.alive = Animal.alive.filter(animal => animal !== prey);
+    }
```
OleksiiBlezniuk, line 33: I don't think that object comparing is a good idea, we can filter by health value, why to compare objects? What if we'll add some functionality that will make a deep copy of objects? How can we understand why array isn't filtered?

By the way it will be more logical to check if health is less or equal then 0, in case if we'll decide to change animal health or damage (it will make possible for `health` to have negative value) - we'll be able to do it without any refactoring

I understand that in that case everything works correct, but in case if that project  would be much bigger, it will lead to spending time to research code to find where is a cause of problem that there are animals with negative health, and for refactoring. So as for me better to learn how to make your code able for scaling and functionality changing without any huge refactoring from this small projects 🙂

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
OleksiiBlezniuk, line 29: But why have you removed check for health? =(
Now you'll filter array even if we don't need it

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
GJ!

PR #1124, opened 2022-10-02
solution

poznianski:


Oleksandra-Shevchenko (approved):
LGTM 👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
OleksiiBlezniuk, line 25: How can this animal stop hiding? 🙂 Better to change `hidden` value to opposite here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey.hidden === false) {
+      prey.health -= 50;
+    }
+
+    if (prey.health === 0) {
```
OleksiiBlezniuk, line 34: Better to check if this value is smaller or equal than 0
We may change animal health or carnivore damage and it shouldn't require any refactor

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
👍

PR #1127, opened 2022-10-02
complete the task

barsuchellos:
image

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,16 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static died(animal) {
+    this.alive = this.alive.filter(el => el !== animal);
+  };
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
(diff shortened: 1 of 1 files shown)


poznianski (approved):
good

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
```
silchyn, line 4: What is the purpose of that 'health' parameter if you do not use it?

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  }
+
+  static died(animal) {
+    this.alive = this.alive.filter(el => el !== animal);
+  };
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
```
silchyn, line 17: Unused parameter here also

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(el) {
```
silchyn, line 28: What is el?

silchyn (changes requested):

PR #1128, opened 2022-10-03
Herbivores and carnivores

Yuriy-Komolov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  bite(victum) {
+    if (victum instanceof Herbivore && victum.hidden === false) {
+      victum.health -= 50;
+    }
+
+    if (victum.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(victum), 1);
+    }
+  }
 }
```
avkovlit, line 33: CleanShot 2022-10-04 at 13 55 37@2x

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(victum) {
+    if (victum instanceof Herbivore && victum.hidden === false) {
+      victum.health -= 50;
+    }
+
+    if (victum.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(victum), 1);
+    }
```
avkovlit, line 31: CleanShot 2022-10-04 at 13 57 03@2x

try to use filter here

avkovlit (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
```
OleksiiBlezniuk, line 9:
```suggestion
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;

    Animal.alive.push(this);
  }
```

OleksiiBlezniuk: Not critical, but better to split blocks of code by blank lines

```
// variables/properties assigning
// variables/properties assigning
// variables/properties assigning
                   (blank line to separate variables block)
// methods/functions creating
                   (blank line to separate function block, 'cause usually it takes more than 2-3 lines)
// methods/functions creating
                   (blank line to separate function block, 'cause usually it takes more than 2-3 lines)
// methods/functions calling
// methods/functions calling
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
```
OleksiiBlezniuk, line 17:
```suggestion
  constructor(name) {
    super(name);

    this.hidden = false;
  }

```

src/herbivoresAndCarnivores.js
OleksiiBlezniuk: Use filter here, also would be better to check if health is less or equal than 0 (in case if we want to change health or damage - it shouldn't require any refactoring)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victum) {
+    if (victum instanceof Herbivore && victum.hidden === false) {
```
OleksiiBlezniuk, line 25:
```suggestion
    if (victum instanceof Herbivore && !victum.hidden) {
```

OleksiiBlezniuk: Use shorthand

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

PR #1129, opened 2022-10-03
Solution

OleksandrGlulhovskiy:


volodymyrklymkiv (approved):
Great

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
+    if (name.hasOwnProperty('hidden') && !name.hidden) {
```
silchyn, line 27: Carnivore can also be hidden theoretically

silchyn (changes requested):

src/herbivoresAndCarnivores.js
OleksandrGlulhovskiy: no, because he hasn't hidden property

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
```
naziamoff, line 26: i believe there's a better name for this variable. I would come up with `animal` or sth
The point is that the variable's name should perfectly describe what is stored inside. In that case it is definitely not a `name`

src/herbivoresAndCarnivores.js
naziamoff: I agree with Olexandr, I think only Herbivores should have the `hidden`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    if (name.hasOwnProperty('hidden') && !name.hidden) {
+      name.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
naziamoff, line 31: Now you do this action after each bite, but actually we need to do it if we know that there's an animal that should be removed from the list of animals.

naziamoff (approved):
Nice job, but check comments

PR #1132, opened 2022-10-03
Solution

avkovlit:


Yuriy-Komolov (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
+  }
+
+  static died() {
```
OleksiiBlezniuk, line 10: probably this method should have other name, function name should have a verb and we should understand that it's an action

OleksiiBlezniuk, line 10:
```suggestion
  static removeDead() {
```

OleksiiBlezniuk (approved):

PR #1136, opened 2022-10-05
commit1

Gintako666:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(goal) {
+    if (goal.hidden !== true && !(goal instanceof Carnivore)) {
+      goal.health -= 50;
+    }
+
+    delete Animal.alive[Animal.alive.findIndex(x => x.health === 0)]
```
alexandra-protyanova, line 31: u cannot use delete for arrays

alexandra-protyanova: use filter here

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1160, opened 2022-10-15
So;ution

Khabailiuk:


TarasHoliuk (approved):
:+1:

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```
loralevitska, line 25: it would be more logical to call the variable `animal`, because if the animal is not an instance of herbivore, then the variable contains a value that doesn't reflect from reality

loralevitska (approved):
great!

PR #1171, opened 2022-10-22
solution

tYuriy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  }
+
+  bite(obj) {
+    if (obj instanceof Herbivore && !obj.hidden) {
+      obj.health -= 50;
+
+      if (obj.health <= 0) {
+        const index = Animal.alive.indexOf(obj);
+
+        Animal.alive.splice(index, 1);
```
nazarmatsevych, line 37: Use filter method insted of indexOf and splice

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
nazarmatsevych, line 20: this method should only hide the animal, not toggle hidden value

nazarmatsevych (changes requested):
Good work 👍 
Check comments pls

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(obj) {
```
TarasHoliuk, line 30: `obj` is a bad name for variable/argument, rename it please

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    if (obj instanceof Herbivore && !obj.hidden) {
+      obj.health -= 50;
+
+      if (obj.health <= 0) {
+        Animal.alive = Animal.alive.filter(an => an.health > 0);
```
TarasHoliuk, line 35: do not use abbreviations for variable/argument names

TarasHoliuk (changes requested):

TarasHoliuk (approved):
:+1:

PR #1175, opened 2022-10-26
Solution

bokorzita:
Hello, I've been trying to figure out why making the alive array a static property didn't work, I eventually removed it from the Animal Class to pass the linter, and it passes the tests as well, but I'm not sure that this is what I was supposed to do.
Screenshot (248)
Screenshot (246)
Screenshot (249)

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -5,11 +5,7 @@ class Animal {
     this.name = name;
     this.health = 100;
 
-    const indexFound = Animal.alive.findIndex(
-      (animal) => animal.name === this.name,
-    );
-
-    if (indexFound === -1) {
+    if (Animal.alive.findIndex(animal => animal.name === this.name) === -1) {
       Animal.alive.push(this);
     }
   }
@@ -44,7 +40,7 @@ class Carnivore extends Animal {
     }
 
     if (!victim.health) {
-      Animal.alive = Animal.alive.filter(x => x.name !== victim.name);
+      Animal.alive = Animal.alive.filter(({ name }) => name !== victim.name);
     }
   }
 }


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
+
+    for (const animal of Animal.alive) {
+      if (animal.name === this.name) {
+        shouldInclude = false;
+      }
+    }
+
+    if (shouldInclude === true) {
+      Animal.alive.push(this);
+    }
```
jstmpelowycz, line 18: let's use more straightforward approach with `findIndex`
also, to handle the issue with `static` keyword, we just have to put declaration below the constructor (hope linter likes this approach)
Screenshot 2022-10-27 at 10 18 21

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
+    Animal.alive.sort((a, b) => b.health - a.health);
```
jstmpelowycz, line 19: I understand why you've used `sort` here – to make sure `pop` works fine
but it could have been done in a more simple way – just by using `filter` method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
+
+  bite(obj) {
+    if (obj instanceof Herbivore && obj.hidden === false) {
+      obj.health -= 50;
+    }
+
+    if (obj instanceof Herbivore && obj.health <= 0) {
+      Animal.alive.pop();
+    }
+  }
```
jstmpelowycz, line 50: I believe it looks better this way:
- what is `obj`? let's call it `victim` instead
- double-checking for `instanceof` – not OK, we can exit the function right at the beginning if needed
- you can avoid using `===` for types which is convertible to boolean
```
bite(victim) {
  if (!(victim instanceof Herbivore)) {
    return;
  }
  
  if (!victim.hidden) {
    victim.health -= 50;
  }

  if (!victim.health) {
    Animal.alive.filter(...);
  }
}
```

jstmpelowycz (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+    this.name = name;
+    this.health = 100;
+
+    const indexFound = Animal.alive.findIndex(
+      (animal) => animal.name === this.name,
+    );
+
+    if (indexFound === -1) {
+      Animal.alive.push(this);
+    }
```
jstmpelowycz, line 14: now I understand that this is a kind of extra logic, but OK

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+    if (!(victim instanceof Herbivore)) {
+      return;
+    }
+
+    if (!victim.hidden) {
+      victim.health -= 50;
+    }
+
+    if (!victim.health) {
+      Animal.alive = Animal.alive.filter(x => x.name !== victim.name);
```
jstmpelowycz, line 47: let's give meaningful names
```suggestion
      Animal.alive = Animal.alive.filter(animal => animal.name !== victim.name);
```
or with destructuring assignment
```suggestion
      Animal.alive = Animal.alive.filter(({ name }) => name !== victim.name);
```

jstmpelowycz (approved):
great job!

PR #1183, opened 2022-10-28
solved

Deniso4ka23:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
```
loralevitska, line 16: this expression should be used not only in the Herbivore class. All animals alive from start, Carnivore too. So move it to 9 line, and avoid code duplicate from 31-33 lines

loralevitska (changes requested):

loralevitska (approved):
great!

PR #1184, opened 2022-10-29
solution

MaxFedak:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,54 @@
...
+  static addAnimal(animal) {
+    this.alive.push(animal);
+  }
+
+  static removeAnimal(animal) {
+    const indexOfAnimal = this.alive.indexOf(animal);
+
+    if (indexOfAnimal !== -1) {
+      this.alive.splice(indexOfAnimal, 1);
+    }
```
Illia-Skladnik, line 16: [CODE PERFOMANCE] Try to avoid nesting loops (indexOf + splice), use filter instead

Illia-Skladnik (approved):
Good work

src/herbivoresAndCarnivores.js
Olexandr6, line 16: Better to use just filter here, it would be better for performance side

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,54 @@
...
+    this.health = 100;
+  }
 }
 
 class Herbivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.addAnimal(this);
```
Olexandr6, line 30: You should add animals in the base `Animal` constructor once, in this case, you don't duplicate this logic in all constructors. It's how the `super` works

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,54 @@
...
+  }
+
+  hide() {
+    this.hidden = this.hidden ? 0 : true;
+  }
```
Olexandr6, line 35: [Checklist](https://github.com/mate-academy/js_herbivores_and_carnivores/blob/master/checklist.md) the 3rd point

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,54 @@
...
 }
 
 class Carnivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+    Animal.addAnimal(this);
+  }
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore.constructor === Herbivore) {
```
Olexandr6, line 45: Use `instanceOf` here instead 
```suggestion
  bite(herbivore) {
    if (!herbivore.hidden && herbivore.constructor === Herbivore) {
```

Olexandr6: At first you should check on instanceOf and then on hidden property it's logically correct

Olexandr6 (changes requested):

Olexandr6 (approved):
Well done 🔥

PR #1186, opened 2022-10-31
Solution

dmytro-legerda:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,46 @@
 
 class Animal {
   // write your code here
+
```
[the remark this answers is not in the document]
TarasHoliuk: the same below

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,46 @@
...
+  addBeast(obj) {
+    Animal.alive.push(obj);
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+
+  bite(obj) {
```
TarasHoliuk, line 36:
```suggestion
  bite(animal) {
```
`obj` is bad name, your variable/argument name should explain what is it

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,46 @@
...
+    if (!obj.hidden && obj instanceof Herbivore) {
+      obj.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(elem => {
+      return !elem.health === 0 && !elem.hidden;
```
TarasHoliuk, line 42: no need to check here if the animal is hidden
too late for it 😅

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,46 @@
...
 
 class Carnivore extends Animal {
   // write your code here
+
+  bite(obj) {
+    if (!obj.hidden && obj instanceof Herbivore) {
+      obj.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(elem => {
```
TarasHoliuk, line 41: and again `elem` is bad name

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,46 @@
...
 class Herbivore extends Animal {
   // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    this.addBeast(this);
+  }
+  hide() {
+    this.hidden = true;
+    this.addBeast(this);
```
TarasHoliuk, line 25: it is definitely redundant here
no need to add a new animal every time when it hides

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,46 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
   // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    this.addBeast(this);
```
TarasHoliuk, line 21: should be only in the `Animal` class constructor (actually, you've added it on the 10 line). If it exists in the constructor of the inherited class it will work for Herbivore as well.

TarasHoliuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(creature => {
+      return !creature.health === 0;
```
AlisaVlasova, line 35:
```suggestion
      return !creature.health <= 0;
```

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #1187, opened 2022-11-01
solution

metinbicaksiz:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,50 @@
 
 class Animal {
   // write your code here
+  // static alive = [];
+
```
jstmpelowycz, line 6: you don't need these comments
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,50 @@
...
+  constructor() {
```
jstmpelowycz, line 7: you should pass `name` parameter via constructor

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,50 @@
...
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
   // write your code here
+  constructor() {
+    super();
+    this.hidden = false;
+
+    Animal.alive.push(this);
```
jstmpelowycz, line 20: you can push each animal in `Animal` class, you'll avoid repeating yourself

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,50 @@
...
 
 class Carnivore extends Animal {
   // write your code here
+  constructor() {
+    super();
+
+    Animal.alive.push(this);
+  };
+
+  bite(name) {
```
jstmpelowycz, line 36: `name` intuitively refers to a string, let's call it `victim`
```suggestion
  bite(victim) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,50 @@
...
+    if (name.hidden === false
+      && name instanceof Herbivore
+      && name.health > 0) {
+      name.health -= 50;
+    }
```
jstmpelowycz, line 41: let's simplify our condition: you don't need verifying `health`, the following code will take care of it 
```suggestion
    if (!victim.hidden && victim instanceof Herbivore) {
      name.health -= 50;
    }
```

jstmpelowycz (changes requested):
awesome! left a few comments

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
```
jstmpelowycz, line 4: you've add it, that's OK now, but you don't use it 😄
```suggestion
  constructor(name) {
    this.name = name;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    super(name, victim);
+    this.victim = name;
+  };
+
+  bite(victim) {
+    if (!victim.hidden && victim instanceof Herbivore) {
+      victim.health -= 50;
+    }
+
+    if (victim.health === 0) {
```
jstmpelowycz, line 35: here is just a simplified way to use such conditions 
```suggestion
    if (!victim.health) {
```

jstmpelowycz (changes requested):
hey, Metin, a few last hints!

PR #1188, opened 2022-11-01
Solution

Julia2063:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(object) {
+    if (object instanceof Herbivore && object.hidden === false) {
+      object.health -= 50;
+    }
+
+    if (object.health === 0) {
+      Animal.alive = Animal.alive.filter(one => one !== object);
```
polosanya, line 30: What is one?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+
+    this.hidden = hidden;
+  }
+  hide() {
+    this.hidden = !this.hidden;
```
polosanya, line 19:
```suggestion
    this.hidden = true;
```
I think it's better to clarify. Also the second call will make hidden false

polosanya (changes requested):
Good job! Please check my comments to improve your code

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden === false) {
+      target.health -= 50;
+    }
+
+    if (target.health === 0) {
```
qoa11a, line 29: I recommend to check if target health is less than or equals to 0, because a target with less than zero health is also dead.

qoa11a (approved):

PR #1191, opened 2022-11-05
Solution

Illia-Skladnik:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.health = 100;
```
iryna-lelyk, line 15: You don't need to add health, because this property is inherited from class Animal

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.health = 100;
```
iryna-lelyk, line 28: Same here

iryna-lelyk (approved):
Great work!)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
denys-danyliuk, line 16: you can do this inside Animal class instead of doing it in every class that extends it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  }
+
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health === 0) {
+      Animal.alive = Animal.alive.filter(ifDead => ifDead !== herbivore);
```
denys-danyliuk, line 36:
```suggestion
      Animal.alive = Animal.alive.filter(animal => animal !== herbivore);
```

denys-danyliuk (changes requested):

denys-danyliuk (approved):

PR #1194, opened 2022-11-07
Solution

taraselnik:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(...arguments);
```
masha-girya, line 15: sure!!! i tried to do that way but but I didn't succeed

masha-girya (approved):
i think it is a good job!)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(anybody) {
```
denys-danyliuk, line 25:
```suggestion
  bite(animal) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    if (anybody instanceof Herbivore && !anybody.hidden) {
+      anybody.health -= 50;
+
+      if (anybody.health <= 0) {
+        const positionAlive = Animal.alive.indexOf(anybody);
+
+        Animal.alive.splice(positionAlive, 1);
```
denys-danyliuk, line 32: 1 rule in checklist

denys-danyliuk (changes requested):

src/herbivoresAndCarnivores.js
taraselnik: ok

src/herbivoresAndCarnivores.js
taraselnik: AAA ok)) thanks

src/herbivoresAndCarnivores.js
illiarerun: come on, you have only 1 argument, don't be shy of writing it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0 && Animal.alive.includes(animal)) {
+        const positionAlive = Animal.alive.indexOf(animal);
+
+        Animal.alive.splice(positionAlive, 1);
+      }
```
illiarerun, line 33: would be easier to use `filter` here and actually create a static method for that

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
```
illiarerun, line 8: again, better to create a static method

illiarerun (changes requested):

illiarerun (approved):

PR #1198, opened 2022-11-07
Task solved

sergey-vorobei:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
```
polosanya, line 4:
```suggestion
  constructor(name, health = 100) {
```
You can set value by default like this

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
polosanya, line 24:
```suggestion
  bite(target) {
```
We don't know for sure if it's herbivore. 
Your name is a bit confusing

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      Animal.alive = Animal.alive.filter(a => a.health > 0);
```
polosanya, line 30: Please give more informative name than just `a`

polosanya (changes requested):
Almost perfect! Please pay attention to my comments

liliia-livandovskaya (approved):

PR #1199, opened 2022-11-07
animals MG

masha-girya:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
taraselnik, line 4: cool

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+
+  static addToAlive(animal) {
+    this.alive.push(animal);
+  }
+
+  static removeFromAlive(animal) {
+    if (this.alive.includes(animal)) {
+      const indexOfAnimal = this.alive.indexOf(animal);
+
```
taraselnik, line 13: interesting idea. but a bit hard;)

taraselnik (approved):

src/herbivoresAndCarnivores.js
masha-girya: yep, I thought that we have to do this by static methods :))

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+      this.alive.splice(indexOfAnimal);
+    }
```
Olexandr6, line 15: Let's use `filter` once, instead of includes + indexOf + splice

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore) {
+      animal.health = animal.hidden ? animal.health : animal.health - 50;
```
Olexandr6, line 39: You can use && operator and move you ternary operator logic to `if` statement

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+    }
+
+    if (animal instanceof Herbivore && animal.health === 0) {
```
Olexandr6, line 42: This check is redundant here. If an animal is dead, it doesn't matter it's herbivore or carnivore)
```suggestion
    if (animal.health === 0) {
```

Olexandr6 (changes requested):

Olexandr6 (approved):
Good job 👍

PR #1200, opened 2022-11-07
Animals created

vadymbaranov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static addAnimal(animal) {
+    Animal.alive.push(animal);
+  }
```
Svitlana-Yudina, line 6: Is it necessary to create a metod? I'm not sure, but I think, you can just push inside constructor...

Svitlana-Yudina (approved):
Great Job!)) 🔥

src/herbivoresAndCarnivores.js
Largoxoxo: I think you can do both. Using method allows you to automate it.  So, all animals will be added despite of their type. Just a way to avoid duplication)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.addAnimal(this);
```
Olexandr6, line 12: This static method is redundant here, just push here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
Olexandr6, line 31: Here can not only harbivores, but carnivores as well, so rename this var to `animal`

Olexandr6: image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
```
Olexandr6, line 32: You should check on instances of Herbivore at first and then on hidden

G-YaSha (changes requested):

src/herbivoresAndCarnivores.js
Largoxoxo: Does it affect performance? && searches for first falsy value anyway. So since only herbivores have hidden ability it doens't really matter. Thought it would matter if the hidden ability was added to carnivores too.

src/herbivoresAndCarnivores.js
Olexandr6: This method is redundant here

G-YaSha (approved):

PR #1202, opened 2022-11-07
herbivores & carnivores

Svitlana-Yudina:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.health = health;
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
Largoxoxo, line 16: You can move this into method in Animal to avoid duplication. I wrote an example above.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+  }
 }
```
Largoxoxo, line 11: I believe linter currently doesn't like this notation. You can move in outside Animal object, like this. There is no error here per se.
```suggestion

  static addAlive(animal) {
  Animal.alive.push(animal)
  }
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;
  }
}

Animal.alive = [];
```

Largoxoxo (approved):
GJ! Wrote a couple of suggestions, you can implement them if you find them valid enough)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
denys-danyliuk, line 4: your tests failed because of this and because we use node version which doesn't support static fields in classes

src/herbivoresAndCarnivores.js
denys-danyliuk, line 11: you can create this field here
```suggestion
Animal.alive = [];
```

src/herbivoresAndCarnivores.js
denys-danyliuk: agree, you can do this in Animal class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  constructor(name, health = 100) {
```
denys-danyliuk, line 6:
```suggestion
  constructor(name) {
```
it's not necessary to take health as constructor parameter

src/herbivoresAndCarnivores.js
denys-danyliuk: it's due to node version we using which doesn't support static fields

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(herbivore) {
+    if (!(herbivore instanceof Carnivore) && !(herbivore.hidden)) {
```
denys-danyliuk, line 30:
```suggestion
    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```

denys-danyliuk (changes requested):

denys-danyliuk (approved):

PR #1220, opened 2022-11-09
Solution

Anatolii-Voloshyn:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 
 class Animal {
-  // write your code here
+  static addAnimal(animal) {
+    Animal.alive.push(animal);
+  }
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
Andriy-Shymkiv, line 10:
```
  constructor(name, health = 100, hidden = false) {
    this.name = name;
    this.health = health;
    this.hidden = hidden;
```

Andriy-Shymkiv (changes requested):
make understandable constructor , because if I was not familiar with this task, i wouldn't get it

src/herbivoresAndCarnivores.js
Anatolii-Voloshyn: Agree with you.. Your solution is more understandable!

Andriy-Shymkiv (approved):
GJ

illiarerun (approved):

PR #1223, opened 2022-11-09
complete

spiritkyiv:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    this.hidden = true;
+  }
 }
 
+Animal.alive = [];
+
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivoreAnimal) {
+    const eatenAnimal = herbivoreAnimal;
```
Anna-Shapovalova, line 26: Are U sure U need here an extra variable?)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+    if (!eatenAnimal.hidden && !(eatenAnimal instanceof Carnivore)) {
+      eatenAnimal.health -= 50;
+
+      if (eatenAnimal.health <= 0) {
```
Anna-Shapovalova, line 31:
```suggestion
      if (eatenAnimal.health === 0) {
```
I don`t think it can be less then 0)

Anna-Shapovalova (approved):
Good job!)))

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+        Animal.alive.splice(Animal.alive.indexOf(eatenAnimal), 1);
```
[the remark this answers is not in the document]
spiritkyiv: @Anna-Shapovalova always ietarate all array when eaten one animal?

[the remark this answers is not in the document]
Anna-Shapovalova: Please, look throw the checklist) 

Screenshot_23

src/herbivoresAndCarnivores.js
illiarerun: Agree, it's redundant
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
+Animal.alive = [];
+
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivoreAnimal) {
+    const eatenAnimal = herbivoreAnimal;
+
+    if (!eatenAnimal.hidden && !(eatenAnimal instanceof Carnivore)) {
```
illiarerun, line 28: why just don't check if it's `Herbivore`, it's easier to work with positive checks

src/herbivoresAndCarnivores.js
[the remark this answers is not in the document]
illiarerun: What if animal wasn't found? Better use filter and create a static method for this logic
image

illiarerun (changes requested):

Olexandr6 (approved):

PR #1224, opened 2022-11-09
Animals solution

shtht24:


mitattt (approved):
cool 👍

sh1gatsu (approved):
🔥

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
```
illiarerun, line 7: please, add a static method for that

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim.hidden === false) {
```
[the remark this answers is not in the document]
illiarerun: Also, you should check if victim IS Herbivore, please add this check

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+      victim.health -= 50;
+    };
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
illiarerun, line 30: What if victim was bitten but it's still alive? Nothing will change, add some type of check on whether you should filter the array. And create a static method for this logic

shtht24: ty for review, check please the new version:)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (!victim.hidden && victim.hidden !== undefined) {
```
illiarerun, line 34: > Also, you should check if victim IS Herbivore, please add this check

please pay attention to comments and add proper check, replace your `undefined` check with `instanceof` for that

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.birth();
+  }
+
+  birth() {
+    Animal.alive.push(this);
```
illiarerun, line 11: it's more convenient to make methods that work with `static` fields, in fact, `static`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  }
+
+  death() {
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
+  }
```
illiarerun, line 16: same here

illiarerun (changes requested):

src/herbivoresAndCarnivores.js
shtht24: yes, it's much much better) i've done final version, please check :)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    this.name = name;
+    this.health = 100;
+    this.birth();
+  }
+
+  birth() {
+    Animal.alive.push(this);
+  }
+
+  death() {
```
denys-danyliuk, line 14: find a better name, name of the methods should start with a verb

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.birth();
+  }
+
+  birth() {
```
denys-danyliuk, line 10: same here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.birth();
```
denys-danyliuk, line 7: it's redundant

denys-danyliuk (changes requested):

src/herbivoresAndCarnivores.js
denys-danyliuk, line 7: it's redundant

shtht24: sorry for late response. so i need to put this constructor to each animal or what? cant actually understand what i should fix here(

illiarerun (approved):

PR #1227, opened 2022-11-09
Solution

Esceype:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static pushAnimal(animal) {
+    Animal.alive.push(animal);
+  }
```
Oleksandr-Stryzhavchuk, line 6: I think that you don't need this

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  static deleteAnimal(animal) {
+    Animal.alive = Animal.alive.filter(beast => beast !== animal);
+  }
+
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+
+    Animal.pushAnimal(this);
+  }
```
Oleksandr-Stryzhavchuk, line 17: just here you can push

Oleksandr-Stryzhavchuk (approved):
GJ!

src/herbivoresAndCarnivores.js
Olexandr6: Agree, looks like a redundant method. Let's get rid of it.

src/herbivoresAndCarnivores.js
Olexandr6: +1

Olexandr6 (approved):

PR #1229, opened 2022-11-09
Solution

Vlad-Fedorishchev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,36 @@
...
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
qoa11a, line 20:
```suggestion
  hide() {
    this.hidden = true;
  }
```
It would be confusing if method hide would unhide an animal.

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+      Animal.alive = Animal.alive.filter((el) => el.health !== 0);
```
qoa11a, line 27: Animals with less than 0 health are probably dead as well.

qoa11a (changes requested):

polosanya (approved):
Good job!

PR #1232, opened 2022-11-09
added solution

kryvets-vadym:


Stepico (approved):
So, it's ok, I like your solution with the static method, but think about this hidden method in the Animal class

mykola-koval (approved):

PR #1234, opened 2022-11-10
Solution

alpopkov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
Stepico, line 4: Due to assignment health is always 100, so it's not necessary to put it as a parameter to the constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
Stepico, line 15: The same here, hidden is always false at the beginning, so it's a little bit redundant when you put it into constructor, but it always has the default value

Stepico (changes requested):
Good job, just pay attention to my comments, only few fixes and it will be perfect :)

PR #1237, opened 2022-11-11
vol_1

RudBeee:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(x) {
```
marina-tilniak, line 25: x - it's bad name. Rename it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    if (x.hidden === false) {
+      Animal.alive.forEach(element => {
+        if (element.name === x.name) {
+          element.health -= 50;
+        }
+      });
+    }
+
+    const result = Animal.alive.filter(elem => elem.health > 0);
```
marina-tilniak, line 34:
```suggestion
    const result = Animal.alive.filter(animal => elem.health > 0);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(x) {
+    if (x.hidden === false) {
+      Animal.alive.forEach(element => {
+        if (element.name === x.name) {
```
marina-tilniak, line 28: you don't need it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(x) {
+    if (x.hidden === false) {
```
marina-tilniak, line 26:
```suggestion
    if (!x.hidden) {
```

marina-tilniak: also you should check the instanceof x

marina-tilniak (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (!victim.hidden && victim instanceof Herbivore) {
+      Animal.alive.forEach(animal => {
+        animal.health -= 50;
+      });
```
TarasHoliuk, line 29: In the https://github.com/mate-academy/js_herbivores_and_carnivores/pull/1237#discussion_r1020319204 Marina meant that you don't need a loop at all (now you reduce health for all animals if only one has been bitten). You already have an animal that should lose health - it is `victim`.

TarasHoliuk (changes requested):

vlad-demchuk (approved):

PR #1238, opened 2022-11-12
herbivoresAndCarnivores was implemented

Andrii-Medintsev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
```
SanyaBratashchuk, line 23: where is constructor?

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
+    if (herb.hidden === false) {
```
SanyaBratashchuk, line 24:
```suggestion
    if ( herb instanceof Herbivore && !herb.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
+      herb.health -= 50;
+    }
+
+    if (herb.health === 0) {
+      Animal.alive
+        .splice(Animal.alive.indexOf(item => herb.name === item.name), 1);
```
SanyaBratashchuk, line 30: use filter

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb.hidden === false) {
+      herb.health -= 50;
+    }
+
+    if (herb.health === 0) {
```
SanyaBratashchuk, line 28:
```suggestion
    if (herb.health <= 0) {
```

SanyaBratashchuk (changes requested):

src/herbivoresAndCarnivores.js
Andrii-Medintsev: I don't get it.. As Carnivore class extends Animal class, it inherits parents constructor, right?

polosanya (approved):
Good job!

PR #1239, opened 2022-11-14
Solution

smiyka:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivoreName) {
+    if (herbivoreName instanceof Herbivore && !herbivoreName.hidden) {
+      herbivoreName.health -= 50;
+    };
+
+    if (herbivoreName.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(herbivoreName, 1));
```
nazarmatsevych, line 32: Use filter method instead of splice and indexOf

tetiana-ivanchenko (approved):

PR #1241, opened 2022-11-16
Sol

TarasV123:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(whoAlive => animal.health > 0);
```
tetiana-ivanchenko, line 31: here better to add condition, and not filter array if animal.health > 0

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1243, opened 2022-11-17
Task was done

Sergey-Mironenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.hidden = false;
+  }
+
+  hide() {
+    if (this.hidden === true) {
+      this.hidden = false;
+    } else {
+      this.hidden = true;
+    }
+  }
```
tetiana-ivanchenko, line 25: it's shouldn't be a toggle

tetiana-ivanchenko (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if ((beast instanceof Herbivore) && (beast.hidden === false)) {
```
AlisaVlasova, line 26:
```suggestion
    if ((beast instanceof Herbivore) && (!beast.hidden)) {
```

AlisaVlasova (approved):

PR #1247, opened 2022-11-21
solution

euFarafonov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
+    if (obj instanceof Herbivore && obj.hidden === false) {
```
Viktor-Kost, line 28:
```suggestion
    if (obj instanceof Herbivore && !obj.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+      obj.health -= 50;
+
+      if (obj.health === 0) {
+        for (let i = 0; i < Animal.alive.length; i++) {
```
Viktor-Kost, line 32: Here u can simplify your code using array method `filter`

Viktor-Kost (changes requested):
Very nice!

Some optimization needed)

euFarafonov:
optimizated

Viktor-Kost (approved):
Great!)

PR #1249, opened 2022-11-23
test Ok

diolna:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,47 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(herbivore) {
+    if (herbivore.hidden !== true
+      && herbivore instanceof Herbivore
+      && herbivore.health >= 50) {
```
loralevitska, line 31: move this condition to variable for better readability

loralevitska (changes requested):
hi! nice solution

src/herbivoresAndCarnivores.js
diolna: fixed

diolna:
fixed

PR #1251, opened 2022-11-25
first commit

viktoria-krasnyuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
+    }
+
+    if (herb.health <= 0) {
+      Animal.alive.splice(herb);
```
TarasHoliuk, line 28: I don't know why the tests passed, but we can't use [splice](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) in such case.

Use another `Array` method instead

TarasHoliuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
+    }
+
+    if (herb.health <= 0) {
+      delete Animal.alive[Animal.alive.indexOf(herb)];
```
polosanya, line 28: You can just filter the animals and leave all except dead herb 
`Animal.alive = Animal.alive.filter(...);`

polosanya (changes requested):
Almost perfect! Please check the comment

loralevitska (approved):
good job!

PR #1255, opened 2022-11-28
Solution

ShkarupaDenys:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.kind = 'carnivore';
+  }
+
+  bite(animal) {
+    if (animal.kind === 'herbivore' && !animal.hidden) {
```
tetiana-ivanchenko, line 39: use "instanceof" instead of this

tetiana-ivanchenko (changes requested):

TarasHoliuk (approved):

PR #1260, opened 2022-12-01
Solution

vadiimvooo:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,51 @@
...
+  constructor(
+    name,
+    health = 100,
+    hidden = false,
+  ) {
+    super(name, health);
+
+    this.hidden = hidden;
+
+    Animal.alive.push(this);
```
romanblyashuk94, line 24: Redundant!
This class extends Animal class. 
So, you push every Herbivore instance 2 times: in Animal constructor, and here.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,51 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(
+    name,
+    health = 100,
+  ) {
+    super(name, health);
+  }
```
romanblyashuk94, line 38: Useless constructor.
It does the same as Annimal constructor.
you can delete it.

romanblyashuk94 (changes requested):

romanblyashuk94 (approved):
Gj

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super();
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
Ihor-Karpyn, line 22:
```suggestion
    this.hidden = true;
```

Ihor-Karpyn (approved):

PR #1262, opened 2022-12-01
Initial commit

artem-hirzhev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
+    }
```
nazarmatsevych, line 33: do not manipulate with array in such way - better just filter it

nazarmatsevych (changes requested):

src/herbivoresAndCarnivores.js
artem-hirzhev: Done.

TarasHoliuk (approved):

PR #1265, opened 2022-12-02
js_herbivores_and_carnivores

ysrckr:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+    }
+
+    if (victim.health === 0) {
```
UmizDemud, line 30: Numbers are working, so no need for <=

UmizDemud (approved):

src/herbivoresAndCarnivores.js
meforoma, line 30:
```suggestion
    if (!victim.health) {
```
would make sense to evaluate `victim.health <= 0` if constructor would allow setting the health points, but in this case it's hardcoded to 100.

meforoma (approved):
good one. One tiny thing to just consider. 👍

PR #1271, opened 2022-12-07
solution

lefteris-L:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = false || true;
```
ysrckr, line 20: I think you tried to make it a toggle. If so you can write `this.hidden = !this.hidden` otherwise `this.hidden = true` is enough.

ysrckr (approved):

src/herbivoresAndCarnivores.js
lefteris-L: That's good. It had to be a toggle, yes

meforoma: @lefteris-L what does this statement ALWAYS return?
`false || true` - it will alway return true. So it does not make sense.
You need to fix such things.
```suggestion
    this.hidden = true;
```

meforoma (changes requested):
good job 👍 
one minor, yet very obvious thing i just can't pass by)

meforoma (approved):

PR #1273, opened 2022-12-08
Solution

romanblyashuk94:
The last test passes only when we run it separately.

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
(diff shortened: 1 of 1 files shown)


GabrielMaksimovich (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      const deadAnimalIndex = Animal.alive
+        .findIndex(({ name }) => name === herbivore.name);
+
+      Animal.alive.splice(deadAnimalIndex, 1);
+    }
```
GabrielMaksimovich, line 35: the problem is here. I have no idea why findIndex is not working but indexOf is working, but if you push animals not is Animal class< but instead in Carnivore and Herbivore classes,

GabrielMaksimovich (approved):
Nice!

valerii-derkach (approved):

PR #1277, opened 2022-12-09
solution

GabrielMaksimovich:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,42 @@
...
+  constructor(name, health) {
+    super(name, health);
+
+    Animal.alive.push(this);
+  }
+
+  bite(obj) {
+    if (!obj.hidden && !(obj instanceof Carnivore)) {
+      obj.health -= 50;
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
romanblyashuk94, line 33: You are doing filtration always, for every bite
It will be better to do it only when it was deadly bite (when obj.health <= 0)

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,42 @@
...
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
romanblyashuk94, line 15: You can save few rows, if write it in Animal class constructor :)

romanblyashuk94 (changes requested):

romanblyashuk94 (approved):
Perfect! ;)

valerii-derkach (approved):

PR #1282, opened 2022-12-10
Solution

webdevw:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  bite(beast) {
+    if (beast instanceof Herbivore
+        && beast.hidden === false) {
+      beast.health -= 50;
+    }
+
+    if (beast.health <= 0) {
+      const dead = Animal.alive.findIndex((el) => el.health <= 0);
+
+      Animal.alive.splice(dead, 1);
```
vlad-demchuk, line 37: let's use `filter` for deletion

vlad-demchuk (changes requested):

loralevitska (approved):
GJ)

PR #1292, opened 2022-12-13
solution

KrisPlanka:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!(herbivore instanceof Carnivore) && herbivore.hidden === false) {
```
batstolya, line 28: I think it would be a little clearer, but it's up to you
```suggestion
    if (herbivore instanceof Herbivore && !herbivore.hidden)  {
```

batstolya (approved):
🔥

Ihor-Karpyn (approved):

PR #1304, opened 2022-12-14
Solution

novytskyi-d:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,38 @@
...
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
```
olena-yanovska, line 14: it seems you don't need health here, just (name)

olena-yanovska (changes requested):

Ihor-Karpyn (approved):
GJ

PR #1305, opened 2022-12-14
a solution that works well for removing students from Mate

RedkoSerhii:


oleksandrnikora (approved):
It looks good for me)

Ihor-Karpyn (approved):
GJ

PR #1308, opened 2022-12-14
solution

anastasiia-roiek:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,36 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
alyonkaradchuk, line 13: maybe it's better to add arguments to constructor and super
like this :
 constructor(name) {
    super(name);

alyonkaradchuk (approved):
in general looks good to me)
good job)

src/herbivoresAndCarnivores.js
Ihor-Karpyn: Agree, name will be undefined

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):

PR #1309, opened 2022-12-14
Animal, Herbivore and Carnivore classes were added

viktoriia-kravchenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
-  // write your code here
+  static add(creature) {
+    this.alive.push(creature);
+  }
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    Animal.add(this);
```
daryna-zinchenko, line 12: I suggest to push 'this' directly to 'Animal.alive' array instead of creating  static method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+
```
daryna-zinchenko, line 21: you don't need constructor here, its already inherited

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(creature) {
+    if (creature instanceof Herbivore && !creature.hidden) {
+      creature.health -= 50;
+    }
+
+    if (creature.health === 0) {
+      Animal.alive = Animal.alive.filter(beast => beast.health > 0);
```
daryna-zinchenko, line 37: you could use filter without condition

daryna-zinchenko (changes requested):

src/herbivoresAndCarnivores.js
viktoriia-kravchenko: I think in my case it's not an option, since I don't specify hidden property in Animal class.

daryna-zinchenko: I guess you could add this property without constructor here

src/herbivoresAndCarnivores.js
viktoriia-kravchenko: I could, but I think it will lead to unnecessary iterations over the array. Every time the bite method is called, and the first condition is satisfied, the filter method will be executed.

daryna-zinchenko (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
```
valerii-derkach, line 14:
```suggestion
  constructor(name, health) {
```


no need to pass hidden as a parameter here - just add property hidden to the class
this.hidden = false;

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #1312, opened 2022-12-14
solution

DimaKolesnyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
```
OlggaJul, line 6: Health is  always 100 by default.

OlggaJul (approved):

src/herbivoresAndCarnivores.js
DimaKolesnyk: TY

ivan-kadykalo (approved):
LGTM

PR #1315, opened 2022-12-14
solution

vladimir-arkanov:


OlggaJul (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,29 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, hidden = false) {
+    this.name = name;
+    this.health = 100;
```
ivan-kadykalo, line 6: Class Animal should have only name and health parameters

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,29 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(allmostDead) {
```
ivan-kadykalo, line 21: `almostDead` is not the best name for this variable. 
Method bite gets the name of the animal, let's write smt like `animalName` or `target`

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #1316, opened 2022-12-14
Solution

NadiiaKoch:


OlggaJul (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super();
```
ivan-kadykalo, line 16: You should set all the parameters to method super()
```suggestion
  constructor(name) {
    super(name);
```

ivan-kadykalo (changes requested):
You are almost there 💪

src/herbivoresAndCarnivores.js
NadiiaKoch: fix

ivan-kadykalo (approved):
Good job 🔥

PR #1317, opened 2022-12-14
Solution

oleh-yablunovskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
daryna-zinchenko, line 16: You don't need constructor here, it should be inherited already.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Carnivore || beast.hidden) {
+      return;
+    }
+
+    beast.health -= 50;
+
+    if (beast.health <= 0) {
```
daryna-zinchenko, line 34: This condition is not nesesary. You also can filter without it.

daryna-zinchenko (changes requested):

src/herbivoresAndCarnivores.js
oleh-yablunovskyi: Can't agree on that! This cunstructor is there to assign hide method to Herbivors. If we move it to Animal, Сarnivors will have this method too (but they shouldn't,  if I understand the task correctly).

src/herbivoresAndCarnivores.js
oleh-yablunovskyi: Well, maybe you are right. Slightly more filtering, but less code. Might be a fair tradeoff. Changed!

src/herbivoresAndCarnivores.js
daryna-zinchenko: good point

daryna-zinchenko (approved):

Ihor-Karpyn (approved):
GJ

PR #1321, opened 2022-12-14
solution

lukacsja:


Lun4trick (approved):
👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Carnivore || herbivore.hidden) {
+      return;
+    }
+    herbivore.health -= 50;
```
naziamoff, line 29: let's make it vice versa, your code will look simpler this way)

I mean if the animal is herbivore and not hidden - we can use bite method there

naziamoff (approved):
Well done, just a couple of suggestions

PR #1323, opened 2022-12-14
solution

mmahfuzi:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+
+  bite(herbivor) {
+    if (herbivor instanceof Herbivore && herbivor.hidden === false) {
```
Grayfox147, line 31:
```suggestion
    if (herbivor instanceof Herbivore && !herbivor.hidden) {
```
You could check if the hidden is false just by using this simplification of the statement, this would mean if herbivor.hidden is not true then proceed. :D basically the same but a little shortcut.

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
...
+      herbivor.health -= 50;
+    }
+
+    if (herbivor.health === 0) {
+      Animal.alive = Animal.alive.filter(a => a.name !== herbivor.name);
```
Grayfox147, line 36: Really Smart approach I never though about using filter in this situation.

Grayfox147 (commented):
Overall really amazing implementation, an excellent job.

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
...
 
 class Carnivore extends Animal {
   // write your code here
+
+  bite(herbivor) {
+    if (herbivor instanceof Herbivore && herbivor.hidden === false) {
+      herbivor.health -= 50;
+    }
+
+    if (herbivor.health === 0) {
```
naziamoff, line 35: Ofc in this case when we have only damage 50 there's no chance that the health is going to be less than zero, but if we decide to change the damage of the `bite()` method to let's say 27 - the animal health will be less than zero and the animal won't be removed in this case

naziamoff (approved):
good job overall, but check my and @Grayfox147 comments

PR #1327, opened 2022-12-14
solution

illiashcherbak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
olena-yanovska, line 22: may be this.hidden = true will be better

olena-yanovska (changes requested):

src/herbivoresAndCarnivores.js
illiashcherbak: you right! thank you

olena-yanovska (approved):

valerii-derkach (approved):

PR #1331, opened 2022-12-15
Complete

SalehAkaJim:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
ysrckr, line 20:
```suggestion
    this.hidden = true;
```
It is better this way because it is not asked to be a toggle.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey.constructor.name === 'Herbivore' && !prey.hidden) {
```
ysrckr, line 26:
```suggestion
    if (prey instanceof Herbivore && !prey.hidden) {
```
Looks cleaner

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+      prey.health -= 50;
+    }
+
+    if (prey.health <= 0) {
```
ysrckr, line 30: Normally I would not warn about this but I was rejected with this. 😄
Because health is fixed 100 in the beginning and we remove 50 when bit we can just
`prey.health === 0`

ysrckr (changes requested):
Couple of changes are needed.

ysrckr (approved):
Looks great 👏🏻

meforoma (approved):
➕

PR #1335, opened 2022-12-15
solution

victoria-kovalenko:


NadiiaKoch (approved):
Nice!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,29 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, hidden = false) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
valerii-derkach, line 8:
```suggestion
  constructor(name) {
    this.name = name;
    this.health = 100;
    Animal.alive.push(this);
```

only Herbivores should have `hidden` property
and don't pass it as a parameter, just ad to the constructor with default value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,29 @@
...
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(bittenAnimal) {
+    if (bittenAnimal.hidden === false && bittenAnimal instanceof Herbivore) {
```
valerii-derkach, line 22:
```suggestion
    if (!bittenAnimal.hidden && bittenAnimal instanceof Herbivore) {
```

PR #1337, opened 2022-12-15
animalsDK

Kukharuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+
+  bite(animal) {
+    if (animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive.indexOf(animal);
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
+    }
```
loralevitska, line 36: Use filter method for removing animals from array
And check is animal herbivore (`instanse of Herbivore`)

loralevitska (changes requested):
almost done)

tetiana-ivanchenko (approved):

PR #1338, opened 2022-12-15
Solution

Grayfox147:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health,);
```
ysrckr, line 15: If you apply the suggestion you don't need health in neither of them.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive.pop(animal);
```
ysrckr, line 31: `pop` doesn't work this way. Try some other way.

ysrckr (changes requested):

meforoma (approved):
that's a score 🎯

PR #1339, opened 2022-12-15
solution

kenvastolord:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
```
ysrckr, line 5:
```suggestion
  constructor(name) {
    this.health = 100;
```
Looks much cleaner like this and you don't have to pass health to each class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(health = 100, name) {
+    super(health, name);
```
ysrckr, line 14: If you apply the suggestion you don't need health neither of them.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(health = 100, name) {
+    super(health, name);
```
ysrckr, line 25: If you apply the suggestion you don't need health neither of them.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
ysrckr, line 28:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

ysrckr (changes requested):
Couple of chances are needed.

ysrckr (approved):
Good job 🤟🏻

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(a => a.health !== 0);
```
naziamoff, line 28: what if we need to add some new method with different damage? I mean why don't just check whether the value is positive?

naziamoff (approved):
LGTM 🔥 
Just one suggestion

PR #1341, opened 2022-12-15
Solution

Iksich:


ysrckr (approved):
A few suggestions.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
naziamoff, line 24: why not just `animal` instead of herbivore? I mean we do not know for know whether exactly the herbivore is passed, so we check it inside the if

naziamoff (changes requested):
let's fix naming inside the bite method, now it is really confusing

meforoma (approved):

PR #1342, opened 2022-12-15
Solution

Rijo-J:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
+        const removeByIndex = Animal.alive.indexOf(animal);
+        Animal.alive.splice(removeByIndex, 1);
```
ysrckr, line 31: You can also use filter method here.

ysrckr (approved):
Looks good. Just left one suggestion.

src/herbivoresAndCarnivores.js
Rijo-J: totally 💯 thank you!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
```
naziamoff, line 29: better to check whether the health is positive

naziamoff (approved):
LGTM 🔥

PR #1343, opened 2022-12-15
added Animal, Herbivore, Carnivore classes

TheOneThatKnowsNothing:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.health = health || 100;
```
Cardlice, line 5:
```suggestion
    this.health = 100;
```
health is 100 by default

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
+    super(name, health);
+    this.hidden = hidden || false;
+  }
+  hide() {
+    this.hidden = !this.hidden;
```
Cardlice, line 18:
```suggestion
    this.hidden = false;
  }
  hide() {
    this.hidden = true;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    const isVisible = !herbivore.hidden;
+
+    if (isVisible && isHerbivore) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive = Animal.alive.filter(animal =>
+          animal.health > 0);
+      }
+    }
```
Cardlice, line 34:
```suggestion
    if (isVisible && isHerbivore) {
      herbivore.health -= 50;
    }
    
    if (herbivore.health <= 0) {
        Animal.alive = Animal.alive.filter(animal =>
          animal.health > 0);
    }
```
there is no need for it nested conditions

Cardlice (changes requested):
GJ, small fixes

src/herbivoresAndCarnivores.js
TheOneThatKnowsNothing: we can avoid filtering all animals in case animal has 50hp after bite

Cardlice: filtering won't work in both cases, but in case of nesting, we just won't check the second condition if the first one is false, maybe you're right, that's better)

Cardlice (approved):

src/herbivoresAndCarnivores.js
ivan-kadykalo: Don't use nested ifs when it is possible

ivan-kadykalo (changes requested):
You are almost there 💪

PR #1344, opened 2022-12-16
Herbivores and carnivores

Mohameddahi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
ysrckr, line 4: This is doable in newer version of JS but in this projects version this does not work. Probably you couldn't pass the test. You should try to change it.

ysrckr (changes requested):

src/herbivoresAndCarnivores.js
Mohameddahi: @ysrckr ok, got it !

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animalName) {
```
naziamoff, line 24: is this actually an animalName or just animal?

naziamoff (approved):
🔥

PR #1345, opened 2022-12-16
solution

grslckr:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast.__proto__ === Herbivore.prototype && beast.hidden !== true) {
```
Rijo-J, line 25:
```suggestion
    if (beast instanceof Herbivore && !beast.hidden) {
```

Rijo-J (approved):
great!! 🚀 passing to mentors

src/herbivoresAndCarnivores.js
naziamoff: agree with @Rijo-J

naziamoff (approved):
🔥 
but take a look at @Rijo-J comment, it totally makes sense

PR #1346, opened 2022-12-16
Solution

sercansimsek:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
ysrckr, line 27: Just small suggestion -> We do not yet know if it is herbivore. So you can use more generic parameter name.

ysrckr (approved):

meforoma (approved):
➕

PR #1347, opened 2022-12-16
Solution

NgineZZ:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  };
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
ysrckr, line 25: `bite(herbivore)` we do not yet know that it is a herbivore so you can give a more generic parameter name.

ysrckr (approved):
Just one tiny suggestion. Otherwise it looks good.

meforoma (approved):
it's a score 🎯

PR #1348, opened 2022-12-16
done

IoghiStudio:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    const isHerbivore = victim instanceof Herbivore;
+
+    if (isHerbivore && !victim.hidden) {
+      victim.health -= 50;
+
+      if (victim.health === 0) {
```
ysrckr, line 32: Roman suggests to do it like this and Nazar suggests to take in the consideration that if we add another method that may cause health to go below zero. So you decide 😄

ysrckr (approved):
Just one thing. Other than that it looks good.

meforoma (approved):
➕

PR #1350, opened 2022-12-17
Solution

IVxVI:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore
+      && this instanceof Carnivore
+    && herb.hidden !== true) {
```
UnknownAI, line 29:
```suggestion
    && !herb.hidden) {
```

UnknownAI: It would be more readable and cleaner to check if `herb.hidden` is false instead of not true

UnknownAI (changes requested):

src/herbivoresAndCarnivores.js
IVxVI: Ok ty

UnknownAI (approved):
👍

PR #1352, opened 2022-12-17
Finished

JzSol:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
-'use strict';
+"use strict";
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
ysrckr, line 6:
```suggestion
  constructor(name) {
    this.name = name;
    this.health = 100;
```
This way is better so you do not need to assign healt in each class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
```
ysrckr, line 15: If you follow the suggestion above;

```suggestion
  constructor(name) {
    super(name);
```
class extending would just need the name.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
```
dolnaro, line 15: I am not really sure if we really need these arguments in constructor and super.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive = Animal.alive.filter((animal) => animal !== herbivore);
+      }
+    }
+  }
```
dolnaro, line 33: Other than that, everything looks great!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
ysrckr, line 25: `bite(herbivore)` we do not know yet that animal is herbivore or carnivore. So a generic parameter name would be nicer.

ysrckr (changes requested):
Just a couple of things needs your attention

src/herbivoresAndCarnivores.js
dolnaro: @ysrckr I am a bit confused here, do we really need these assignments here? Because as I understand, we already reaching these assignments by extending the Herbivore class to the Animal class. Or am I missing something?

ysrckr: You need to take in the `name` in Animal Class. You don't need to take in the health, you can just assign it to 100. In herbivore, we need to take in the name and pass that to the super. I hope it cleared it out for you.

meforoma (approved):
➕

PR #1353, opened 2022-12-17
sol1

dolnaro:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Carnivore || beast.hidden) {
+      return;
+    }
+
+    beast.health -= 50;
```
ysrckr, line 32:
```suggestion
    if (beast instanceof Herbivore && !beast.hidden) {
          beast.health -= 50;
    }
```

ysrckr: Wouldn't it be better this way?

ysrckr (changes requested):
Just needs a small tweak

src/herbivoresAndCarnivores.js
dolnaro: That line was looking off to me, but now it is more understandable :)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
ysrckr, line 16:
```suggestion
  constructor(name) {
    super(name);
```
You need to take `name` in order to have access to it in `const checkName = animal => animal.name !== beast.name;`

Sorry I didn't see it in my first review.

ysrckr (changes requested):

src/herbivoresAndCarnivores.js
dolnaro: I didn't get this part actually. I mean, code works without name because we are extending the constructor to Animal. I didn't get what I am missing here. Can you explain?

ysrckr: Actually, after reading some documentation if you do not take in new information or you don't need to modify any of the existing properties that were in parent class. You don't need a constructer at all. So you can just delete `constructor` and `super` completely.

ysrckr (approved):
LGTM 🤟🏻

meforoma (approved):

PR #1357, opened 2022-12-19
solution

Roman-Malysh:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.addAnimal(this);
+  }
+
+  static alive = [];
```
tetiana-ivanchenko, line 11: you need to define slive after Animal class creation

tetiana-ivanchenko (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
nazarmatsevych, line 4: add `Animal.alive = []` after Animal class creation

nazarmatsevych (changes requested):

tetiana-ivanchenko (approved):

PR #1362, opened 2022-12-19
js_herbivores_and_carnivores

rafmamedov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  bite(animal) {
+    if (animal instanceof Herbivore
+      && animal.hidden !== true) {
+      animal.health -= 50;
+    }
+
+    for (let i = 0; i < Animal.alive.length; i++) {
+      if (Animal.alive[i].health === 0) {
+        Animal.alive.splice(i, 1);
+      }
```
vlad-demchuk, line 34: it's better to use filter method

vlad-demchuk (changes requested):

src/herbivoresAndCarnivores.js
rafmamedov: exactly, thanks!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore
+      && animal.hidden !== true) {
```
```suggestion
      && !animal.hidden) {
```

SanyaBratashchuk (approved):

src/herbivoresAndCarnivores.js
rafmamedov: got it, thanks!

PR #1368, opened 2022-12-24
js_herbivores_and_carnivores done

EhorSaprykin:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
+
+      if (herb.health === 0) {
+        const removedAnimal = Animal.alive.indexOf(herb);
+
+        Animal.alive.splice(removedAnimal, 1);
```
vlad-demchuk, line 36: it will be better to use `filter` method

vlad-demchuk (changes requested):

src/herbivoresAndCarnivores.js
EhorSaprykin: @vlad-demchuk  Done

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
```
loralevitska, line 29: don't cut variable names, it seems as if you expect plants to be instance of animals
```suggestion
  bite(herbivore) {
```

loralevitska (approved):

src/herbivoresAndCarnivores.js
EhorSaprykin: @loralevitska ok, ty

PR #1369, opened 2022-12-24
Solution

vovakov123:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
polosanya, line 17: You don't need to duplicate it as both Herbivores and Carnivores extend from Animal class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(live) {
```
```suggestion
  bite(target) {
```
Let's use more correct name

polosanya (changes requested):
Almost done! Please check the comments

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
```
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  }
+
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health !== 0);
```
polosanya, line 35: I think it's better to add some check before filtering in order not to filter the animals after every bite

polosanya (approved):
Good job! Just please check my comments

PR #1370, opened 2022-12-26
js_herbivores_and_carnivores

mykolahuryn:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(x => x.health > 0);
```
polosanya, line 28: Would be great to add a health check before filtering in order not to filter the animals after every bite

polosanya (approved):
Good job!

PR #1373, opened 2022-12-30
Solution

skomskiy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,30 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
SanyaBratashchuk, line 7:
```suggestion
```
hidden should be only in instance of Herbivore

SanyaBratashchuk (changes requested):

agniya-i (approved):

PR #1377, opened 2023-01-02
solution

Malva37:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivoreObject) {
+    if (herbivoreObject.hidden !== undefined && !herbivoreObject.hidden) {
+      herbivoreObject.health -= 50;
+    }
+
+    if (herbivoreObject.health === 0) {
+      const indRemovedAnimal = Animal.alive.findIndex(el => el.health === 0);
```
nzrii-kysel, line 29: could just use `indexOf` here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivoreObject) {
+    if (herbivoreObject.hidden !== undefined && !herbivoreObject.hidden) {
```
nzrii-kysel, line 24: could use `instanceof` for checking if target is herbivore or carnivore

nzrii-kysel (approved):
well done

ArturErmolenko (approved):

PR #1380, opened 2023-01-03
Solution

diakovMax:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
polosanya, line 7: It is redundant here. Not all animals have hidden property.

polosanya (changes requested):
Almost done! Please check the comments

qoa11a (changes requested):

PR #1383, opened 2023-01-05
first commit

haduigon:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,48 @@
 
 class Animal {
   // write your code here
```
vlad-demchuk, line 4: remove comments

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,48 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
+
+Animal.deleteDead = function(dead) {
+  for (let i = 0; i < Animal.alive.length; i++) {
+    if (Animal.alive[i].name === dead) {
+      Animal.alive.splice(i, 1);
```
vlad-demchuk, line 17: better to use `filter` method here

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,48 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(victim) {
+    if (victim.hasOwnProperty('hidden') && victim.hidden === false) {
```
vlad-demchuk, line 37: check if `herbivore instanceof Carnivore` as well
and do you need a `victim.hasOwnProperty('hidden')` check?

vlad-demchuk (changes requested):

PR #1384, opened 2023-01-05
herb and carn task completed

MykolaKaradzha:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden === true) {
```
OlEzhA0, line 26:
```suggestion
    if (animal instanceof Carnivore || animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    }
+
+    animal.health -= 50;
+
+    if (animal.health <= 0 && Animal.alive.includes(animal)) {
+      const index = Animal.alive.indexOf(
+        creature => creature === animal);
+
+      Animal.alive.splice(index, 1);
+    }
```
OlEzhA0, line 37: add this method to `Animal` class and use just filter method to remove animal from list

OlEzhA0 (changes requested):
hey! check my comments

src/herbivoresAndCarnivores.js
MykolaKaradzha: Yeap, with filter is much easier, but why I should add bite method to Animal?

mykhalenych (approved):

PR #1385, opened 2023-01-05
Solution

Dmytro-Litinskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
vlad-demchuk, line 31: it will be better to use `filter` method here

vlad-demchuk (changes requested):

tetiana-ivanchenko (approved):

PR #1386, opened 2023-01-05
NVyrsta

NVyrsta:


Avichez (approved):
Great solution 👍 
I forgot about such a simple method filter 🥲

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
```
alexandra-protyanova, line 6: `health` is not passed as an argument by the task

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
alexandra-protyanova, line 21:
```suggestion
  constructor(name) {
    super(name);
    this.hidden = false;
  }

  hide() {
    this.hidden = !this.hidden;
  }
```

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1389, opened 2023-01-07
Solution

MrFox32:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
polosanya, line 16: Can't you do it in Animal class?
Both herbivores and carnivores are animals.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive = Animal.alive.filter(el => el !== animal);
```
polosanya, line 36: Are you sure that health can be < 0?
Also, would be great to use names better than `el`

polosanya (changes requested):
Almost perfect! Let's improve your code a bit)

Anastasiia-Svintsova (approved):

PR #1393, opened 2023-01-10
Develop: Found the solution

sonyaxent:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
```
SanyaBratashchuk, line 12: call parent's costructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
```
```suggestion
    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```

SanyaBratashchuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.carnivore = true;
```
SanyaBratashchuk, line 29:
```suggestion
    this.carnivore = true;
```
Why do you need it?

SanyaBratashchuk (approved):

PR #1394, opened 2023-01-10
Herbivores and carnivores

anastasia-radchencko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore
+      && !herbivore.hidden) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      delete Animal.alive[Animal.alive.indexOf(herbivore)];
```
agniya-i, line 32: use `filter` method instead of `delete` to update the array

agniya-i (changes requested):

tetiana-ivanchenko (approved):

PR #1395, opened 2023-01-10
solution

igor-kashchenko:


OlhaYevstifieieva (approved):
Nice gob 🙋🏻‍♀️

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden) {
+      return;
+    }
+
+    animal.health -= 50;
+
+    if (!animal.health) {
+      Animal.alive = Animal.alive.filter(n => n.health > 0);
```
alexandra-protyanova, line 33:
```suggestion
      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden) {
```
alexandra-protyanova, line 26: by logic better to use `if` with nondefault condition, in your case you decrease health only if animal is not hidden and it is `Herbivore`

alexandra-protyanova (approved):

PR #1399, opened 2023-01-11
solution

vasyliev-anton:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Animal {
-  // write your code here
+  createArray() {
+    if (!Animal.alive) {
+      Animal.alive = [];
+    }
+  }
+  // The ways with static alive = []
+  // or Animal.alive = [] didnt pass the Elinter;
```
loralevitska, line 10: Which line you add? It should be fine. If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if ((victim instanceof Herbivore) && victim.hidden === false) {
+      victim.health -= 50;
+    }
+
+    if (victim.health <= 0) {
+      delete Animal.alive[Animal.alive.indexOf(victim)];
```
loralevitska, line 35: use `filter` here

loralevitska (changes requested):

PR #1405, opened 2023-01-12
Solution

chermak-oleh:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
```
GarryRocksman, line 4: you don't need to add health in constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
+    if (name.hidden || name instanceof Carnivore) {
```
GarryRocksman, line 26: try to use more relevant naming of variables  - name it's not the best use beast or animal instead. Adl this IF is is a bit strange am I right?  Maybe it's better to check if the animal is Herbivore and hidden or note to minus health

GarryRocksman (approved):
GJ, but need some fixes

src/herbivoresAndCarnivores.js
Anna-Tkachenko: agree, fix @GarryRocksman comment
health won't be changed for all animals, it'll be 100 for all of them

src/herbivoresAndCarnivores.js
Anna-Tkachenko: fix @GarryRocksman comment

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+      return;
+    }
+
+    name.health -= 50;
+
+    if (name.health === 0) {
+      const index = Animal.alive.indexOf(name);
+
+      Animal.alive.splice(index, 1);
```
Anna-Tkachenko, line 35: for now you delete an animal using 2 iterations
can we do it in 1 iteration?

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1410, opened 2023-01-13
Herbivores and Сarnivores - competed

GarryRocksman:
Have a nice day)

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,33 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = []
+  health = 100;
+
+  constructor(name) {
+    this.name = name;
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 
 class Animal {
-  // write your code here
+  static alive = []
+  health = 100;
+
+  constructor(name) {
+    this.name = name;
+    Animal.alive.push(this);
+  }
```
Anastasiia-Khudych, line 10: If you want to pass linter tests, you should write all added properties with this and set the alive array in another way like this 
image

Anastasiia-Khudych (approved):
Good job!

Anna-Tkachenko (approved):
⭐️

PR #1419, opened 2023-01-16
solution

Oleksandr-Maly:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+  bite(herbivore) {
+    if (
+      herbivore.hidden === false
```
anna-inozemets, line 31: We can use just (!herbivore.fidden)

anna-inozemets (approved):
Great solution!)

ArturErmolenko (approved):

PR #1421, opened 2023-01-16
Solution

stadnytskyi-ivan:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(health = 100, name) {
+    this.health = health;
```
Anastasiia-Khudych, line 5: Health is equal to 100 by default, so you can write it in the constructor and don't pass it as an argument.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  }
+
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(herbivore), 1);
+    }
```
Anastasiia-Khudych, line 35: Second point in the checklist (Try to avoid nesting loops (indexOf + splice), use filter instead)

Anastasiia-Khudych (changes requested):
Good job, but check the comments

ArturErmolenko (approved):
GJ 👍

PR #1424, opened 2023-01-16
Solution

tania-boiar:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
Anastasiia-Khudych, line 6: Health isn't passed in the constructor so you can remove it from the parameters.

Anastasiia-Khudych (approved):
Good job!

src/herbivoresAndCarnivores.js
Anna-Tkachenko: agree, fix @Anastasiia-Khudych comment, health value won't change and will be the same for all animals

Anna-Tkachenko (approved):

PR #1426, opened 2023-01-16
Solution

antskpps:


nzrii-kysel (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    };
+
+    Animal.alive = Animal.alive.filter(deadAnimal => deadAnimal.health > 0);
```
Anna-Tkachenko, line 31: it's not clear why you define each animal as a dead one 🥲

Anna-Tkachenko (approved):

PR #1432, opened 2023-01-16
Task was solved

ihor-soloviov:


Avichez (approved):
Great and clear solution 🙂 👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  bite(animal) {
+    if (animal instanceof Herbivore
+      && animal.hidden === false) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
+        const indexOfPrey = Animal.alive.indexOf(animal);
+
+        (indexOfPrey !== -1)
+        && Animal.alive.splice(indexOfPrey, indexOfPrey + 1);
```
ArturErmolenko, line 33: checklist point 2

ArturErmolenko (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore
+      && animal.hidden === false) {
```
Anna-Tkachenko, line 26:
```suggestion
    if (
      animal instanceof Herbivore
      && animal.hidden === false
    ) {
```

Anna-Tkachenko (approved):

PR #1444, opened 2023-01-17
solution_1

Borys-Andrew:


0xNazarii (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive = Animal.alive.filter(el => el.health !== 0);
```
Anna-Tkachenko, line 32: what is el?

Anna-Tkachenko (approved):

src/herbivoresAndCarnivores.js
Borys-Andrew: it means animal from array of Animal...
 =|

PR #1446, opened 2023-01-19
Finish task

U-Ra-All:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,39 @@
...
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
tetiana-ivanchenko, line 35: use filter instead of splice

tetiana-ivanchenko (changes requested):

src/herbivoresAndCarnivores.js
U-Ra-All: Okay, done

mykhalenych (approved):

PR #1460, opened 2023-02-01
Solution

novak-k:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
+  bite(animal) {
+    if (animal instanceof Carnivore
+      || (animal instanceof Herbivore && animal.hidden === true)) {
+      return;
+    }
+    animal.health = animal.health - 50;
+
+    if (animal.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
vlad-demchuk, line 39: It will be better to use filter method here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    this.hidden = true;
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (animal instanceof Carnivore
+      || (animal instanceof Herbivore && animal.hidden === true)) {
+      return;
+    }
+    animal.health = animal.health - 50;
```
vlad-demchuk, line 36:
```suggestion
    if ((animal.hidden === false) && (animal instanceof Herbivore)) {
      animal.health -= 50;
    }
```

vlad-demchuk (changes requested):

src/herbivoresAndCarnivores.js
novak-k: sure

PR #1475, opened 2023-02-11
Solution

undepend:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 }
 
+Animal.alive = []; // need to use static propery instead
+
+Animal.deleteDead = function() {
+  for (let i = 0; i < Animal.alive.length; i++) {
+    if (Animal.alive[i].health === 0) {
+      Animal.alive.splice(i, 1);
+    }
+  }
```
nazarmatsevych, line 19: not a good idea to manipulate with array in this way, instead you can you filter method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = []; // need to use static propery instead
```
nazarmatsevych, line 12: Static properties are quite new, so they are not yet supported in tasks from the platform. You can remove this comment

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
```
nazarmatsevych, line 34: Use better naming

nazarmatsevych (changes requested):
Almost perfect,
pls check my comments

src/herbivoresAndCarnivores.js
undepend: image

undepend: not allowed to change name here

IvanFesenko (approved):
Well done

PR #1483, opened 2023-02-15
● Solution

d-hubych:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+
+  constructor(name) {
+    this.name = name;
+    Animal.born(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
```
IvanFesenko, line 29: It should be inside constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && beast.hidden === false) {
```
IvanFesenko, line 38:
```suggestion
    if (beast instanceof Herbivore && !beast.hidden) {
```

IvanFesenko (changes requested):
Please check my comments

nazarmatsevych (approved):

PR #1484, opened 2023-02-16
Solution__v001

MykolaButylkov:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,43 @@
...
   // write your code here
+  bite(nameHerb) {
+    if (nameHerb.hidden === false) {
+      nameHerb.health -= 50;
+
+      if (nameHerb.health === 0) {
+        delete Animal.alive.pop(nameHerb);
+      }
+    }
+  }
```
Viktor-Kost, line 41: 1) it's not a nested if condition. Both conditions should be at one level
2) update first condition 
```
    if (nameHerb.hidden === false && nameHerb instanceof Herbivore) {
      nameHerb.health -= 50;
    }
```
3) update second condition. Because what will happen if your animal has 30hp and you make a bite and take 50hp?
```
      if (nameHerb.health <= 0) {
        Animal.alive.pop(nameHerb);
      }
```

Viktor-Kost (changes requested):

src/herbivoresAndCarnivores.js
MykolaButylkov: yeah, you're right!

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,39 @@
...
 
 class Carnivore extends Animal {
   // write your code here
+  bite(nameHerb) {
+    if (nameHerb.hidden === false && nameHerb instanceof Herbivore) {
+      nameHerb.health -= 50;
+    }
+
+    if (nameHerb.health <= 0) {
+      Animal.alive.pop(nameHerb);
```
IvanFesenko, line 35: need to use filter. because there is no guarantee that it will be the last

IvanFesenko (approved):
All god, use `filter` insted `pop`

PR #1486, opened 2023-02-17
Solution

taras-nakonechnyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
nazarmatsevych, line 21: It's not toggle method,  so it should change hidden value to true

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
nazarmatsevych, line 16: You can do it once it Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    const isHerbivore = animal instanceof Herbivore;
+    const isHidden = !animal.hidden;
```
nazarmatsevych, line 32:
```suggestion
    const isHidden = animal.hidden;
```

Or change it to isVisible

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
+    if (isHerbivore && isHidden) {
```
```suggestion
    if (isHerbivore && !isHidden) {
```

Or change it to isVisible

nazarmatsevych (changes requested):

OleziO (approved):

PR #1488, opened 2023-02-19
solution

kuznetsova-anastasiia:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = true;
+    Animal.alive.push(this);
+  }
```
stanislavcodes, line 29: I think this part is redundant.
`Carnivore` shouldn't have `hidden` property

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
stanislavcodes, line 16: Why not move this to the `Animal` constructor or to separate static method?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = true;
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (!animal.hidden) {
```
stanislavcodes, line 32: I see that your `Carnivore` objects can bite any other animals. That's really bad 😭 

Please check task details again

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive = Animal.alive.filter((beast) => beast.health > 0);
```
stanislavcodes, line 37: You're changing static properties outside of the `Animal` class. What about another static method for that?

stanislavcodes (changes requested):

stanislavcodes (approved):
Looks good to me! 👍🏻

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.removeKilledAnimal();
+    }
```
OleksiiBlezniuk, line 37: This solution will be more scalable, if you'll change logic and will check if health is less or equal to 0 (we may pass `damage` as a parameter and pass number that can be greater than 50, or we can change animal health)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static removeKilledAnimal() {
+    Animal.alive = Animal.alive.filter((beast) => beast.health > 0);
```
OleksiiBlezniuk, line 12:
```suggestion
    Animal.alive = Animal.alive.filter(({ health }) => health > 0);
```

OleksiiBlezniuk (approved):

PR #1500, opened 2023-02-25
Solution

timoxa2t:


vladyslav-mosor (changes requested):

vladyslav-mosor (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (!(prey instanceof Herbivore)
+      || prey.hidden
+    ) {
```
illiarerun, line 28: can we fit it in one line?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+      return;
+    }
+
+    prey.health -= 50;
+
+    Animal.alive = Animal.alive.filter(
+      ({ health }) => health > 0
+    );
```
illiarerun, line 36: do we need to do it if the health of the animal is not lower than 0?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
```
illiarerun, line 7: let's create a `static` method for it, you usually want to work with static values using static methods

src/herbivoresAndCarnivores.js
illiarerun, line 36: same here

illiarerun (changes requested):

src/herbivoresAndCarnivores.js
timoxa2t: the task said that animal should be removed when it's health reaches zero. So yeah

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (!(prey instanceof Herbivore) || prey.hidden) {
+      return;
+    }
+
+    prey.health -= 50;
+
+    Animal.checkAlive();
```
OleksiiBlezniuk, line 42: Let's do it only when we know that there is an animal that have 0 or less health

OleksiiBlezniuk (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1505, opened 2023-02-26
:white_check_mark: implement solution for herbivoreAndCarnivore kata

Bogdan-Mykhailov:
implement solution for herbivoreAndCarnivore kata

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive.splice(Animal.alive.indexOf(herbivore), 1);
+      }
+    }
+
+    return this.health;
+  }
 }
 
 module.exports = {


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
timoxa2t, line 26: looks strange. Better to avoid equality check with true|false
```suggestion
    if (!herbivore.hidden) {
```

timoxa2t: Also before checking if hidden, check if your herbivore is actualy a herbivore ( instanceof )

timoxa2t (changes requested):
Good work. Just make some changes in bite func

src/herbivoresAndCarnivores.js
Bogdan-Mykhailov: 👍🏼

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden
+      && herbivore instanceof Herbivore) {
```
illiarerun, line 27: can we fit it in one line?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
```
illiarerun, line 7: would be greate to create static method for this, e.g. `birth`. usually you want to change static values only via static methods

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden
+      && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive = Animal.alive.filter(({ health }) => health > 0);
```
illiarerun, line 31: same here, `death` can be created 💀

PR #1507, opened 2023-02-26
herbivores and carnivores solution

StetsenkoViktoriia:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (!victim.hidden && victim instanceof Herbivore) {
+      victim.health -= 50;
+    }
+
+    if (victim.health === 0) {
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
+    }
```
qoa11a, line 34:
```suggestion
    if (victim.health <= 0) {
      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
    }
```
I recommend making it less or equal to zero. Because if you'd subtract, for example, 23 health for every bite, an animal wouldn't have 0 health. But if animal has health below 0 it's considered dead as well

qoa11a (approved):

PR #1510, opened 2023-02-27
solution

MSA1991:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(beast => beast.health > 0);
```
Volodymyr-Melnyk-Fron10der, line 30: you are filtering animals on every bite. I suggest to make condition if health of current animal is 0 `(!animal.health)` and only the filter animals to remove current dead animal from array

Volodymyr-Melnyk-Fron10der (changes requested):
Keep it up, just a few fixes

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (!animal.health) {
+      Animal.alive = Animal.alive.filter(beast => beast.health > 0);
```
```suggestion
      Animal.alive = Animal.alive.filter(({ health })=> health > 0);
```

Volodymyr-Melnyk-Fron10der (approved):
WellDoneGIF

PR #1512, opened 2023-02-28
solution

Roman-Kulchytskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+
+    Animal.alive.push(this);
```
DarkMistyRoom, line 17: no need to repeat this line in each class, just do it in the Animal

DarkMistyRoom (changes requested):

mykhalenych (approved):

PR #1514, opened 2023-02-28
solution

Zelinskyi-Serhii:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
+    if (name instanceof Herbivore) {
+      const find = Animal.alive.find(animal => animal === name);
```
rashyd-hasratov, line 25: I think you don't need this line of code. The instance of the animal is already passed to the function, so you can use it to check if it is the instance of Herbivore, to check its `hidden` property and so on

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+      const findIndex = Animal.alive.findIndex(animal => animal === name);
+
+      if (find.hidden === false) {
+        find.health -= 50;
+      }
+
+      if (name.health <= 0) {
+        Animal.alive.splice(findIndex, 1);
```
rashyd-hasratov, line 33: Checklist says that we can improve our code if we use filter method to delete an animal from the array, we don't need to search for index, and then delete it using splice.

rashyd-hasratov (changes requested):
Nice! Just little improvements are needed

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,42 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
+    if (name instanceof Herbivore) {
+      if (name.hidden === false) {
+        name.health -= 50;
+      }
+
+      if (name.health <= 0) {
+        const res = Animal.alive.filter(beast => beast.health > 0);
```
illiarerun, line 30: what's `res`? choose a better variable name

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,42 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
+    if (name instanceof Herbivore) {
+      if (name.hidden === false) {
```
illiarerun, line 25: can we combine both conditions?

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,42 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
```
illiarerun, line 23: what's `name`? choose a better parameter name

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,42 @@
...
+    if (name instanceof Herbivore) {
+      if (name.hidden === false) {
+        name.health -= 50;
+      }
+
+      if (name.health <= 0) {
+        const res = Animal.alive.filter(beast => beast.health > 0);
+
+        Animal.alive = res;
```
illiarerun, line 32: we can do it in one line
```suggestion
        Animal.alive = Animal.alive.filter(beast => beast.health > 0);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,42 @@
...
+      }
+    }
+  }
 }
 
+Animal.alive = [];
```
illiarerun, line 38: let's move it closer to Animal class

illiarerun (changes requested):

src/herbivoresAndCarnivores.js
Zelinskyi-Serhii: lint
лінтер нас не розуміє.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(nameOfAnimal) {
```
OleksiiBlezniuk, line 25: strange variable name

OleksiiBlezniuk, line 25:
```suggestion
  bite(animal) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    if (nameOfAnimal instanceof Herbivore && nameOfAnimal.hidden === false) {
```
OleksiiBlezniuk, line 26:
```suggestion
    if (nameOfAnimal instanceof Herbivore && !nameOfAnimal.hidden) {
```

OleksiiBlezniuk (approved):

PR #1516, opened 2023-02-28
animals were created

rashyd-hasratov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
```
gretoq, line 31: Why did you make 'if' inside another one 'if'? It is a bad idea. You can use only one level 'if' statement

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
gretoq, line 28: In herbivore.hidden you already have boolean value.

gretoq (changes requested):

src/herbivoresAndCarnivores.js
rashyd-hasratov: If herbivore is hidden I don't even have to check its health, so I don't think it is a good idea to create the second 'if' after the first one.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    Animal.alive.push(this);
```
illiarerun, line 8: would be greate to create static method for this, e.g. birth. usually you want to change static values only via static methods

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive = Animal.alive.filter(animal => animal !== herbivore);
```
illiarerun, line 32: and `death` here

illiarerun (approved):

PR #1519, opened 2023-03-01
Herbivores and Сarnivores were created

serge-kontakt:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  health = 100;
+  hidden = false;
```
anstsot: only Herbivore has to have this property. it would be better to add it only in that class

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,37 @@
...
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast.hidden === false && beast instanceof Herbivore === true) {
```
anstsot, line 20: you can simplify it to:
```suggestion
    if (!beast.hidden && beast instanceof Herbivore) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,37 @@
...
+      beast.health -= 50;
+
+      for (let i = 0; i < Animal.alive.length; i++) {
+        if (Animal.alive[i].health === 0) {
+          delete Animal.alive[i];
+        }
+      }
```
anstsot, line 27: it's better not to change static properties outside of the Animal class. What about creating a specific static method for that?
and there is no need to use for loop here. you can always check the `beast` you've gotten in `bite` method.

anstsot (changes requested):
Looks good but let's improve your code a little bit

PR #1523, opened 2023-03-02
HandC

Yevhen-Ivanchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    if (!target.health) {
+      Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```
anstsot, line 32: as a suggestion, it would be great to create separate static method in `Animal` class and use it here instead of changing another class in `Carnivore`

anstsot (approved):
Great job, approved!

PR #1525, opened 2023-03-02
Herbivores and Carnivores - Solution

Yevhenii0536:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(anotherAnimal) {
+    if (anotherAnimal instanceof Herbivore && !anotherAnimal.hidden) {
+      anotherAnimal.health -= 50;
+    }
+
+    if (anotherAnimal.health <= 0) {
+      Animal.alive = Animal.alive.filter((enimal) => enimal.health > 0);
+    }
```
NikitaStambul, line 31: this if statement needs to be checked only if previous one is correct, so if animal will be an Carnivore or hidden Herbivore, it wont even check it

NikitaStambul (approved):
good job

src/herbivoresAndCarnivores.js
Yevhenii0536: nice, thanks you!

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(anotherAnimal) {
```
OleksiiBlezniuk, line 24: What does `another` mean?

OleksiiBlezniuk: Give more specified names

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,36 @@
...
+    if (anotherAnimal instanceof Herbivore && !anotherAnimal.hidden) {
+      anotherAnimal.health -= 50;
+      Animal.alive = Animal.alive.filter((enimal) => enimal.health > 0);
```
OleksiiBlezniuk, line 27:
```suggestion
      Animal.alive = Animal.alive.filter(({ health }) => health > 0);
```

OleksiiBlezniuk, line 27: Let's do this only if we know that there are animals that has less or equal 0 health

OleksiiBlezniuk (changes requested):

sergii-nosachenko (approved):
Good work 👍
Commented

PR #1526, opened 2023-03-02
solution

kateryna-babina:


Yekatery (approved):
Well done!)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore || animal.hidden)) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(({ health }) => health > 0);
```
OleksiiBlezniuk, line 30: Better to check if current animal health is lower or equal 0 to avoid unnecessary filtering

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore || animal.hidden)) {
```
OleksiiBlezniuk, line 26:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

OleksiiBlezniuk: It's just looks strange, let's just check condition when we can perform this action

OleksiiBlezniuk (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1531, opened 2023-03-02
Solution

MaksymMaliuk:


nazarovoleksiy (approved):
Good Job =)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(({ health }) => health > 0);
```
OleksiiBlezniuk, line 31: Let's do this only when we know that some animal has 0 or less health

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index > -1) {
```
Andr1yK, line 13:
```suggestion
    if (index >= 0) {
```

Andr1yK (approved):

PR #1534, opened 2023-03-02
Animals

andriibezkorovainyi:


sergey-lipikhin (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
OleksiiBlezniuk, line 30: Your code will be more scalable if you'll check that animal health is lower than or equal 0
What if we'll change damage or animals health and became lower than 0?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+      Animal.alive
+        = Animal.alive.filter(beast => beast !== animal);
```
OleksiiBlezniuk, line 32: Comparing objects is bad idea. Let's instead check if beast health is greater than 0 here

OleksiiBlezniuk (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1535, opened 2023-03-02
add task solution

Peti51:


polosanya (changes requested):
Almost done!

Peti51:
I only need the bite1 because without that the constructor is redundent and doesn't  work any suggestion?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(
+    health,
+    name) {
+    super(health, name);
```
Volodymyr-Melnyk-Fron10der, line 18:
```suggestion
  constructor(health, name) {
    super(health, name);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(
+    health,
+    name,
+    bite1) {
```
```suggestion
  constructor(health, name) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    super(health, name);
+  }
+
+  bite(prey) {
+    if (prey.hidden === false && prey instanceof Herbivore) {
```
Volodymyr-Melnyk-Fron10der, line 36:
```suggestion
    if (!prey.hidden && prey instanceof Herbivore) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+      prey.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(function(item) {
+      return item.health > 0;
+    });
```
Volodymyr-Melnyk-Fron10der, line 42: you can simplify it
```suggestion
    Animal.alive = Animal.alive.filter(({ health }) => health);
```

Volodymyr-Melnyk-Fron10der (changes requested):
a few more fixes I see

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(health, name, bite1) {
+    super(health, name);
+  }
```
polosanya, line 28: If it says that your constructor is redundant, then just remove it)
Everything will work the same without it

polosanya (changes requested):
Almost perfect! One small fix left

Volodymyr-Melnyk-Fron10der (approved):
Well done

PR #1536, opened 2023-03-02
Solution

nazarovoleksiy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+Animal.alive = [];
+
+Animal.checkIsAlive = () => {
+  for (const animal of Animal.alive) {
+    if (animal.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
+    }
+  }
+};
+
```
Lilu580, line 19: You dont need to add new method in animal,you can check for alive animals in method bite use filter (checklist.2)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+
+    Animal.alive.push(this);
+  }
```
Lilu580, line 39: you can write this without constructor and supper just adding Animal.alive.push(this); in constructor in your class Animal and remove constructor from Hebrivore and add for Animal hidden === false

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+    this.name = name;
+    this.health = health;
+  }
 }
 
+Animal.alive = [];
+
+Animal.checkIsAlive = () => {
+  for (const animal of Animal.alive) {
+    if (animal.health === 0) {
```
kuznetsova-anastasiia, line 14: what if the task will implement the bite with different damage? let's check if animals have 0 or less health

src/herbivoresAndCarnivores.js
kuznetsova-anastasiia: agree, you can use filter

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+};
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+
+    this.hidden = hidden;
+
+    Animal.alive.push(this);
```
kuznetsova-anastasiia, line 26: you can move this to the Animal constructor

src/herbivoresAndCarnivores.js
kuznetsova-anastasiia: agree, you can avoid using constructor here

kuznetsova-anastasiia (changes requested):

nazarovoleksiy:
Thanks for double penetration...

Lilu580 (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.name = name;
+    this.health = health;
+
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
+Animal.checkIsAlive = () => {
```
OleksiiBlezniuk, line 14:
```suggestion
Animal.removeDead = () => {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      Animal.checkIsAlive();
```
OleksiiBlezniuk, line 35: Let's do it only if we know that there is an animal with 0 or less health

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
sergii-nosachenko, line 6: You don't need to pass health to a constructor. Just set it to 100.
```suggestion
  constructor(name) {
    this.name = name;
    this.health = 100;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+Animal.alive = [];
+
+Animal.removeDead = () => {
+  Animal.alive = Animal.alive.filter(({ health }) => health > 0);
+};
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
```
sergii-nosachenko, line 20: The animal needs only a name to be created.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+    this.hidden = hidden;
```
sergii-nosachenko, line 22: It's false on every creation.

sergii-nosachenko (changes requested):
Quick fixes are needed.

sergii-nosachenko (approved):
Good work 👍

PR #1540, opened 2023-03-04
Develop js_herbivores_and_carnivores

sergey-lipikhin:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+    if (value >= 0) {
+      this._health = value;
+    } else {
+      this._health = 0;
+    }
+  }
+
+  get health() {
+    return this._health;
+  }
```
mstrilec, line 21: you don't need it at all.

mstrilec (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  bite(prey) {
+    if (
+      prey instanceof Herbivore
+      && !prey.hidden
+    ) {
+      prey.health -= 50;
+
+      if (prey.health === 0) {
+        Animal.alive = Animal.alive.filter((animal) => animal !== prey);
+      }
```
OleksiiBlezniuk, line 35: Let's avoid nested if's

OleksiiBlezniuk: And your solution will be more scalable if you'll check if health is greater or equal than 0
What if we'll change damage or health?

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,56 @@
...
+
+    Animal.alive.push(this);
+  }
+
+  set health(value) {
+    if (value >= 0) {
+      this._health = value;
+    } else {
+      this._health = 0;
+    }
```
sergii-nosachenko, line 16: Good place for ternary operator.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1545, opened 2023-03-05
solution for animals

kuzmi491:


kuznetsova-anastasiia (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+
+      if (herbivore.health === 0) {
```
OleksiiBlezniuk, line 29: Your code will be more scalable if you'll check if health is greater or equal then 0 instead of equality
What if we'll change damage or health and our animal will have less then 0 health?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+        Animal.alive = Animal.alive.filter(({ health }) => health > 0)
+      }
```
OleksiiBlezniuk, line 31: Let's avoid nesting if's

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
```
sergii-nosachenko, line 15: When an animal is created its health always equals 100. Also, it's always visible until you don't call hide().
You can leave only name argument, others are useless.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, hidden = false) {
+    super(name);
+    this.hidden = hidden;
+  }
```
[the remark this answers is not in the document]
kuzmi491: I see what you mean=)

PR #1549, opened 2023-03-05
Solution v1

yaroslav-vasilyev:


Vlados3611 (approved):
Good work

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    Animal.alive = Animal.alive.filter(animal => animal !== this);
+  }
```
OleksiiBlezniuk, line 12: I don't think that instances should have this method

OleksiiBlezniuk: Let's make static method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+
+      if (target.health <= 0) {
+        target.die();
+      };
```
OleksiiBlezniuk, line 35: Let's avoid nested if's

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  static die(animal) {
+    Animal.alive = Animal.alive.filter(a => a !== animal);
```
sergii-nosachenko, line 11: Bad practice to compare objects by references. You can filter by health value. Or assign to every animal a unique id and use it in the filter.

sergii-nosachenko: Also, what is `a`?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+
+      target.health <= 0 && Animal.die(target);
```
sergii-nosachenko, line 33: Ninja code. Bad practice, hard to debug.
Use regular if instead.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1550, opened 2023-03-05
solution

valeriiayarzhemska:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
BozhenaBodnar, line 26: WOW how it works

BozhenaBodnar (approved):
nice job:)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast.hidden !== true && beast instanceof Herbivore) {
+      beast.health -= 50;
+    }
+
+    if (beast.health < 50) {
```
OleksiiBlezniuk, line 35: But if beast has 40 health - it still alive, isn't it? Wild nature is cruel, but let's give it a chance

OleksiiBlezniuk: Check that animal has less or 0 health to remove it from the list

OleksiiBlezniuk (changes requested):

src/herbivoresAndCarnivores.js
valeriiayarzhemska: > 

hidden = false set like a default and when we call Herbivore.hide() it becomes true

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static deleteBeast() {
```
sergii-nosachenko, line 11:
```suggestion
  static deleteDeadBeasts() {
```

sergii-nosachenko (approved):
Good work 👍
Commented on naming

PR #1551, opened 2023-03-06
task completed

ArthurKrepko:


valeriiayarzhemska (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore
+      && animal.hidden === false
+      && animal.health > 0
+    ) {
```
OleksiiBlezniuk, line 29:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(victim => victim.health !== 0);
```
OleksiiBlezniuk, line 33: Why should we filter this array even if there are no dead animals?

OleksiiBlezniuk (changes requested):

illiarerun (approved):

PR #1552, opened 2023-03-06
herbivores_and_carnivores

MaksymHuriev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
```suggestion
    if (!animal.health) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+      Animal.alive.pop(animal);
```
Volodymyr-Melnyk-Fron10der, line 32: It won't works actually. You should remove specific animal, not just last one

`The pop() method removes the last element from an array and returns that value to the caller.`

Volodymyr-Melnyk-Fron10der (changes requested):
Keep it up) Check my comments pls

DarkMistyRoom (approved):

PR #1553, opened 2023-03-06
herbivoresAndCarnivores done

Stepan-TS:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.addAnimal();
+  }
+
+  addAnimal() {
+    Animal.alive.push(this);
+  }
```
OleksiiBlezniuk, line 12: We don't need this method in instances, let's make it static

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+
+  removeAnimal() {
+    Animal.alive = Animal.alive.filter(({ health }) => health > 0);
+  }
```
OleksiiBlezniuk, line 16: Same here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health === 0) {
```
OleksiiBlezniuk, line 39: Your code will be more scalable if you'll consider that health or damage can be changed and change `===` to `<=`

OleksiiBlezniuk (changes requested):

Stepan-TS:
fixed comments)

Andr1yK (approved):

PR #1556, opened 2023-03-06
Solution

BozhenaBodnar:


milenadiamonds1 (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
-  // write your code here
+  static add(animal) {
+    this.alive.push(animal);
+  }
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.add(this);
+    Animal.notDeath();
```
OleksiiBlezniuk, line 12: Why should we do it every time when we creating our instances?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  }
+
+  static notDeath() {
```
OleksiiBlezniuk, line 15:
```suggestion
  static removeDead() {
```

OleksiiBlezniuk: Function name should represent action that this function doing

OleksiiBlezniuk (changes requested):
GJ, but let's remove unnecessary actions

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
sergii-nosachenko, line 29:
```suggestion
    this.hidden = true;
```
Just set to true.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
sergii-nosachenko, line 35:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

sergii-nosachenko (changes requested):

src/herbivoresAndCarnivores.js
BozhenaBodnar: okey, i will fix:)

Andr1yK (approved):

PR #1559, opened 2023-03-06
Solution

YaroslavDzyadiv:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
```
DenysDubas, line 17: it will be better, because we are lazy programmers)
```
 constructor(name) {
    super(name);
    this.hidden = false;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive = Animal.alive.filter(({ health }) => health > 0);
+      }
```
OleksiiBlezniuk, line 31: Better to avoid nested if's

OleksiiBlezniuk (approved):

PR #1561, opened 2023-03-06
Solution

sasha-dranuta:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static animalDeath(deathOfAnimal) {
+    Animal.alive = Animal.alive
+      .filter(animal => animal.name !== deathOfAnimal.name);
+  }
```
OleksiiBlezniuk, line 13: I guess it's enough to compare health, but you may leave it as it is now

PR #1567, opened 2023-03-07
task done

tecHamster:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
tetiana-ivanchenko, line 32: use filter instead

tetiana-ivanchenko (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
polosanya, line 26: image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        animal.die();
```
polosanya, line 36: Animal health can't be less than zero. Also, you can make die method static, as it just filters all animals and is not connected with the exact animal

polosanya (changes requested):
Almost done!

anstsot (approved):
Great job, approved!

PR #1569, opened 2023-03-08
Solution

DaveBeetle:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    };
+
+    Animal.alive = Animal.alive.filter(elem => elem.health > 0);
```
anstsot, line 30: it would be better to create an additional method in `Animal` class and place filtering there. because it's not a good practice to change class outside. 
and you can just check current animal and remove it from `Animal.alive`. there is no need to filter the whole array each time

anstsot (approved):
Good job, approving!
you can check my comment and improve your code

PR #1572, opened 2023-03-09
Solution

IvanVaverchak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
IvanFesenko, line 7: hidden prop should be added only for Herbivore

IvanFesenko (changes requested):
Good job. Fix Herbivore constructor

VitaliyBondarenko1982 (approved):

PR #1576, opened 2023-03-10
1st version

Grechana:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    super(name, health);
+  }
+
+  bite(food) {
+    if (!(food instanceof Carnivore || food.hidden)) {
+      food.health -= 50;
+    }
+
+    if (!food.health) {
+      Animal.alive.pop(food);
```
nazarmatsevych, line 39: Try not to manipulate with array in this way, you can use method filter instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    }// write your code here
```
nazarmatsevych, line 40: Remove comments

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  } // write your code here
```
nazarmatsevych, line 25: Remove comments

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    if (this.health) {
+      Animal.alive.push(this);
+    }
+  }
   // write your code here
```
nazarmatsevych, line 14: Remove comments

nazarmatsevych (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+  bite(food) {
+    if (!(food instanceof Carnivore || food.hidden)) {
+      food.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(x => (x.health > 0));
```
ivbutsykin, line 33: Hm, I suggest to add this condition before: `if (food.health <= 0)`. So your code will be like this:
```suggestion
    if (food.health <= 0) {
      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
    }
```

Why? After bite you will iterate by your alive animals always. Even if your animal not die

ivbutsykin (changes requested):

src/herbivoresAndCarnivores.js
Grechana: oh, I understand

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 class Animal {
-  // write your code here
-}
+  static alive = []
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
 
+    if (this.health) {
```
agniya-i, line 10: I guess this check can be redundant, because `this.health` always contains some value

agniya-i (approved):

PR #1581, opened 2023-03-12
Herbivores & Carnivores v.0.1 - Solution

NedobezhkinOleksandr:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  die() {
+    Animal.alive = Animal.alive.filter(animal => animal !== this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
Volodymyr-Melnyk-Fron10der, line 18: health has default value in Animal class, so it's not required to do it again in child class

Volodymyr-Melnyk-Fron10der (approved):
Good👍

PR #1587, opened 2023-03-14
Solution

Mar4enkofff:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
OleziO, line 4: Remove `health = 100` from constructor props, you don't use it,

OleziO: Or start using it :)

OleziO (changes requested):

SerhiiKirik (approved):
Well done 🚀

PR #1589, opened 2023-03-15
Initial commit

RomanHrytsiuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
Volodymyr-Melnyk-Fron10der, line 14: you have set default in the base class
```suggestion
  constructor(name, health, hidden = false) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
Volodymyr-Melnyk-Fron10der, line 27: 1. simplify
2. add check is it really instance of herbivore
```suggestion
    if (!herbivore.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(x => x.health !== 0);
```
Volodymyr-Melnyk-Fron10der, line 31: do it only if health <= 0 and filter items only if health<=0 as well

Volodymyr-Melnyk-Fron10der (changes requested):
Almost there!

OleziO (approved):

PR #1590, opened 2023-03-15
Solution

andrii-kovalskyi02:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive.pop();
```
Volodymyr-Melnyk-Fron10der, line 31: how would you know that last element is exactly animal you need to delete?) change this logic

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
Volodymyr-Melnyk-Fron10der, line 30: let's make sure that health is not positive value
```suggestion
    if (animal.health <= 0) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, good job, almost there)

PR #1592, opened 2023-03-17
Solved

silcki:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,69 @@
...
+    }
+
+    animal.bitten(this.biteHealth);
+
+    if (animal.isDead()) {
+      const key = Animal.alive.findIndex(x => x === animal);
+
+      if (key) {
+        Animal.alive.splice(key, 1);
+      }
```
IvanFesenko, line 65: You can use Array.filter instead

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,69 @@
...
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+
+    return this.hidden;
```
IvanFesenko, line 24: I think it's useless
```suggestion
```

IvanFesenko (approved):

PR #1600, opened 2023-03-22
created animal class

prokivskaaa:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,54 @@
...
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+
+    Animal.alive.push(this);
```
Esteban4769, line 18: Dont need that here, u already use `push` in Animal constructor.

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,54 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+
+    Animal.alive.push(this);
```
Esteban4769, line 30: Same right here, also dont really need a constructor here.

Esteban4769 (changes requested):
Well done, but need some fixes

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,54 @@
...
+
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.filter(
+        (storedAnimal) => storedAnimal.name !== animal.name
+      );
+    }
+  }
 }
 
+Animal.alive = [];
```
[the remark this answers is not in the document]
prokivskaaa: Thanks) But shouldn't we follow the current eslint rules of the project? Probably, the task would provide us with the requirement if we had to do that.

SerhiiKirik (approved):
GJ 🔥

PR #1601, opened 2023-03-22
ClassTask

IvanFront-end:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(
+    health,
+    name,
+    hidden = false,
+  ) {
+    super(health, name);
```
Volodymyr-Melnyk-Fron10der, line 22: first argument is name as I see in your Parent class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => {
+      if (animal instanceof Herbivore && animal.health === 0) {
+        return false;
+      }
+
+      return true;
+    });
```
Volodymyr-Melnyk-Fron10der, line 43: do some check, is health < or = then 0.
And there is no need to make this condition `animal instanceof Herbivore && animal.health === 0`

Volodymyr-Melnyk-Fron10der (changes requested):
keep it up👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => {
+      if (animal.health <= 0) {
```
OleziO, line 38: Just return `animal.health >= 0`, you don't need all these returns, just do it in one line.

OleziO (approved):

PR #1604, opened 2023-03-24
Solution

LBondarchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.health >= 50
+      && animal.hidden === false) {
```
polosanya, line 28:
```suggestion
    if (animal instanceof Herbivore && animal.hidden === false) {
```
Actually, there is no need to check the health as you remove creatures with 0 health from an array of animals

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+      animal.health -= 50;
+
+      Animal.alive = Animal.alive.filter(a => a.health !== 0);
```
```suggestion
      if (animal.health === 0) {
        Animal.alive = Animal.alive.filter(a => a.health !== 0);
      }
```
It is better to check it here to avoid redundant method calls

polosanya (changes requested):
Almost done! Let's improve your code a bit)

src/herbivoresAndCarnivores.js
LBondarchuk: Hello. I did that from the beginning. But then reworked to remove that if. Because these ifs are constantly given comments

anstsot (approved):
Great job, approved!

PR #1610, opened 2023-03-28
Solution

BogdanParubchenk01:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    this.name = name;
+    this.health = health;
+    this.hidden = hidden;
```
anstsot, line 7: `hidden` property should be only related to `Herbivore`. Please add it only there

anstsot (changes requested):
Good job, but there are few fixes to improve your code.

PR #1611, opened 2023-03-28
Solution

AnyaLyaLya:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
jstmpelowycz, line 4: `health` is not configurable, 100 by default for all animals
```suggestion
  constructor(name) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+    this.name = name;
+    this.health = health;
+    this.alive = [];
```
jstmpelowycz, line 7: `alive` is static, you don't need this property here
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+
+    this.hidden = hidden;
+  }
```
jstmpelowycz, line 17: `hidden` is not configurable as well
```suggestion
  constructor(name) {
    super(name, health);

    this.hidden = false;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    const isHerbivore = animal instanceof Herbivore;
+
+    if (isHerbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
```
jstmpelowycz, line 31: it won't get less than 0, just 0
```suggestion
      if (animal.health === 0) {
```
or 
```suggestion
      if (!animal.health) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+        Animal.alive = Animal.alive.filter(animalToDelete => {
+          return animalToDelete.name !== animal.name;
+        });
```
jstmpelowycz, line 34: just a suggestion — you can avoid `return`
```suggestion
        Animal.alive = Animal.alive.filter(animalToDelete => (
          animalToDelete.name !== animal.name
        ));
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+      }
+    }
+  }
 }
 
+Animal.alive = [];
```
jstmpelowycz, line 40: let's put it closer to `Animal`

jstmpelowycz (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
```
jstmpelowycz, line 16: :harold:
```suggestion
  constructor(name) {
    super(name);
```

jstmpelowycz (approved):

PR #1612, opened 2023-03-28
solution

Oleh-Rybachuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victum) {
+    if (victum.hidden === false) {
```
vlad-demchuk, line 27: you also should check if animal is instanceof Herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+      victum.health -= 50;
+    }
+
+    if (victum.health === 0) {
+      Animal.alive.pop();
```
vlad-demchuk, line 32: better to use `filter` method instead of `pop`

vlad-demchuk (changes requested):

Viktor-Kost (approved):
GJ!

PR #1618, opened 2023-03-29
wish you lacky day

glots-a:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    this.hidden = true;
+  };
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
```
makandriival, line 30: you don't really need all this stuff in the constructor - since the predator is not hiding and we don't require to know its name

makandriival (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    // i have to pass this prop here because i had some eslint error.
```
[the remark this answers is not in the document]
SerhiiKirik: agree, remove comment

src/herbivoresAndCarnivores.js
SerhiiKirik: agree, just remove this

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  bite(animal) {
+    if (animal.hidden || animal instanceof Carnivore) {
```
SerhiiKirik, line 33:
```suggestion
    if (animal.hidden || animal instanceof Herbivore) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+      return;
+    }
+    animal.health -= 50;
+
+    if (animal.health === 0) {
```
SerhiiKirik, line 38: Use beast.health <= 0 instead of beast.health === 0 to account for cases when the health goes below 0.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+      Animal.alive = Animal.alive.filter(beast => {
+        return beast.name !== animal.name;
+      });
+    }
```
SerhiiKirik, line 42: you can make it shortest
```suggestion
      Animal.alive = Animal.alive.filter(beast => (
        beast.name !== animal.name;
      ));
    }
```

SerhiiKirik (changes requested):
sync

src/herbivoresAndCarnivores.js
glots-a: it crush my logic!

liliia-livandovskaya (approved):

PR #1621, opened 2023-03-30
func working

BondarenkoDmitriy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.alive = [];
```
SerhiiKirik, line 7: you already do it in 42 line

SerhiiKirik (approved):
Great 🔥

PR #1626, opened 2023-03-30
Solution

Zibi95:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  contructor(name) {
```
DorotaLeniecDev, line 6: Typo here should be
```suggestion
  constructor(name) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+  }
+
+  static updateAlive(target) {
+    const index = this.alive.indexOf(target);
+
+    if (target.health <= 0) {
+      this.alive.splice(index, 1);
+    } else {
+      this.alive[index].health = target.health;
+    }
```
DorotaLeniecDev, line 18: I think this **else** statement is redundant. As **target.health** is already set **Carnivore.bite()** method

DorotaLeniecDev (changes requested):
Looking good but small changes needed :)

src/herbivoresAndCarnivores.js
Zibi95: Ohhh that's why it wasn't working as i tought it should work with inheritance. Now it all makes sense :D

DorotaLeniecDev (changes requested):
Yeah now it is looking good with inheritance in place. One more performance improvement comment

DorotaLeniecDev (approved):
Approved

PR #1633, opened 2023-03-31
Solution

danylo-dobrovolskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.alive = [];
```
ivadimko, line 7: The `alive` property should be a static property on the `Animal` class, not an instance property. Move it to after the class definition and initialize it as an empty array.

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,46 @@
...
+  bite(beast) {
+    const isHerbivore = beast instanceof Herbivore;
+
+    if (!isHerbivore || beast.hidden) {
+      return;
+    }
+
+    beast.health -= 50;
+
+    if (beast.health === 0) {
```
ivadimko, line 34: Use `beast.health <= 0` instead of `beast.health === 0` to account for cases when the health goes below 0.

IvanTkachenko3301 (approved):

src/herbivoresAndCarnivores.js
SerhiiKirik, line 7: remove this, bc you already add static property on the Animal class
```suggestion
```

src/herbivoresAndCarnivores.js
SerhiiKirik: Agree

package.json
```diff
@@ -17,7 +17,7 @@
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "^7.13.0",
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.9.14",
+    "@mate-academy/scripts": "^1.2.8",
```
SerhiiKirik, line 20: In the future - Do not add extra files to PR

SerhiiKirik (changes requested):
Well done 🚀 
Let's fix comments 🧑‍💻

liliia-livandovskaya (approved):

PR #1639, opened 2023-03-31
solve task

kukla1989:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
```
Anastasia2403, line 7: I think this property must be only in the Herbivore class

Anastasia2403 (approved):

src/herbivoresAndCarnivores.js
liliia-livandovskaya: Agree, move it to the Herbivore class

liliia-livandovskaya (approved):

PR #1640, opened 2023-03-31
solution

Goto-Jesus:


Hamuud (approved):
If we add your cat🐱 to Carnivore, how much health❤️‍🩹 will it take from a bite🤕?

You created two static methods and it inspired me to rework my own code🧑‍💻, looks great, keep up the good work

liliia-livandovskaya (approved):

PR #1641, opened 2023-03-31
solution

Olekk17:


DimaDvm (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim.hidden === false) {
```
SerhiiKirik, line 25: Also, you should check this 
image

```suggestion
    if (!victim.hidden) {
```

SerhiiKirik (changes requested):
Great 🔥 
Let's make some improvements 🧑‍💻

PR #1646, opened 2023-04-01
herbivoresAndCarnivores

ymys1:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
AP-Solution, line 6: you can specify 100 as the default value of the health function argument

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+    }
+
+    if (victim.health === 0) {
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
AP-Solution, line 32: It works, but I think the condition with animal.name will be better practice in this case

AP-Solution (approved):
Good job, nice solution!

liliia-livandovskaya (approved):

PR #1663, opened 2023-04-03
Solution

TanyaS133:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,46 @@
...
+  bite(animal) {
+    const isHerbivore = animal instanceof Herbivore;
+
+    if (!isHerbivore || animal.hidden) {
+      return;
+    }
+
+    animal.health -= 50;
+
+    if (animal.health <= 0) {
```
liliia-livandovskaya, line 35:
```suggestion
    if (!animal.health) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,46 @@
...
+      Animal.alive = Animal.alive.filter((beast) =>
+        (beast.name !== animal.name));
+    }
+  }
 }
 
+Animal.alive = [];
```
liliia-livandovskaya, line 42: put it closer to the `Animal` class

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.alive = [];
```
liliia-livandovskaya, line 7:
```suggestion
```

liliia-livandovskaya: you already have a static array `alive`, why do you need it?

liliia-livandovskaya (changes requested):
Good job! Lets work on improvement

PR #1669, opened 2023-04-06
Solution

meltourme:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,38 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
loralevitska, line 29: Also check here instance of animal according task point
image

```suggestion
    if (!animal.hidden) {
```

loralevitska (changes requested):
Almost done) Let's improve

.eslintrc.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   extends: '@mate-academy/eslint-config',
-  parser: "babel-eslint",
```
SerhiiKirik, line 3: In the future - Do not add extra files to PR

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
SerhiiKirik, line 30: also, you can use <= 0 to handle case when animal has health less than 0
```suggestion
    if (animal.health <= 0) {
```

SerhiiKirik (approved):
Great 🔥

PR #1681, opened 2023-04-13
done

victor-buldenko:


package.json
```diff
@@ -17,7 +17,7 @@
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "^7.13.0",
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.9.14",
+    "@mate-academy/scripts": "^1.2.8",
```
SerhiiKirik, line 20: In the future - Do not add extra files to PR

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
SerhiiKirik, line 16: move this to the Animal Class, bc if you do like that - you don't need to create in all classes
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+  bite(target) {
+    const status = target instanceof Herbivore;
+
+    if (target.hidden === false && status) {
```
SerhiiKirik, line 31: You can make it short, bc `hidden` it's a boolean type
```suggestion
    if (!target.hidden && status) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+      target.health -= 50;
+    }
+
+    if (!target.health) {
```
SerhiiKirik, line 35: Just check if health <= 0
image
```suggestion
    if (target.health <= 0) {
```

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your solution 🧑‍💻

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+  bite(target) {
+    const status = target instanceof Herbivore;
```
SanyaBratashchuk, line 28:
```suggestion
    const isHerbivore = target instanceof Herbivore;
```

SanyaBratashchuk (approved):

PR #1683, opened 2023-04-13
Solution

MaksymKos:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
DarkMistyRoom, line 4: Please set static property outside the class to pass the tests.
this way: `Animal.alive = [];`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  }
+
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
+      herbivore.health = herbivore.health - 50;
+    }
+
+    if (herbivore.health <= 0) {
+      herbivore.die();
+    }
```
DarkMistyRoom, line 40:
```suggestion
    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
      herbivore.health = herbivore.health - 50;
      
      if (herbivore.health <= 0) {
        herbivore.die();
      }
    }
```

DarkMistyRoom: no need to check it without bitting

DarkMistyRoom (changes requested):

VitaliyBondarenko1982 (approved):

PR #1687, opened 2023-04-14
Develop

Dka888:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
kasiakubisiak, line 27:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+      animal.health -= 50;
+      Animal.alive = Animal.alive.filter(aliveAnimal => aliveAnimal.health > 0);
```
kasiakubisiak, line 29: Lets move this to Animal class and put inside new method with name e.g. 'die' and then reuse it here 
```suggestion
    animal.die();
```

kasiakubisiak (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -6,6 +6,17 @@ class Animal {
     this.health = 100;
     Animal.alive.push(this);
   }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+      animal.die();
+    }
```
kasiakubisiak, line 14: Almost there, just move bite method to Carnivore class. After you do that, you can also move animal.die() outside of if statement.

kasiakubisiak (changes requested):

kasiakubisiak (approved):

PR #1692, opened 2023-04-17
Solution

mateusz-michalowski:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+
+  bite(prey) {
+    if (prey.hidden === false && prey instanceof Herbivore) {
+      prey.health -= 50;
+    }
+
+    if (prey.health <= 0) {
+      Animal.alive.splice(
+        Animal.alive.indexOf(
+          animal => Animal.alive.animal.name === prey.name), 1);
```
natalia-klonowska, line 37: checklist point 2 - Try to avoid nesting loops (indexOf + splice), use filter instead

natalia-klonowska (changes requested):

d-wasilewski (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,37 @@
...
 
 class Carnivore extends Animal {
   // write your code here
+
+  bite(prey) {
+    if (prey.hidden === false && prey instanceof Herbivore) {
+      prey.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
d-wasilewski, line 34: Actually, let's move this to Animal class and put inside new method e.g. 'killAnimal' and then reuse it here like 
```suggestion
    animal.killAnimal();
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,45 @@
...
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+    Animal.alive.push(this);
+  }
+
+  static killAnimal(name) {
+    const index = Animal.alive.indexOf(name);
+
+    Animal.alive.splice(index, 1);
```
d-wasilewski, line 14: use filter to filter out animals with health <= 0

d-wasilewski: and you can then remove that condition from bite

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1698, opened 2023-04-17
solution

YBM3:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  bite(animal) {
+    const isHerbivore = animal instanceof Herbivore;
+
+    if (!isHerbivore || animal.hidden) {
+      return;
+    }
+
+    animal.health -= 50;
+
+    Animal.alive = Animal.alive.filter(beast => beast.health !== 0);
```
kasiakubisiak, line 36: Lets move this to Animal class and put inside new method e.g. die and then reuse it here 
```suggestion
    animal.die();
```

kasiakubisiak (changes requested):

d-wasilewski (approved):

PR #1699, opened 2023-04-18
first solution

Ozwiena:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+  bite(creature) {
+    if (creature instanceof Herbivore && creature.hidden !== true) {
```
kasiakubisiak, line 31:
```suggestion
    if (creature instanceof Herbivore && !creature.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+      creature.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(x => x.health > 0);
```
kasiakubisiak, line 34: Use better naming here
```suggestion
    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```

kasiakubisiak: After adding better naming move this to Animal class and put inside new method e.g. die and then reuse it here 
```suggestion
    animal.die();
```

kasiakubisiak (changes requested):

PR #1700, opened 2023-04-18
solution

natalia-klonowska:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,33 @@
 
 class Animal {
   // write your code here
+  static alive = [];
```
kasiakubisiak, line 5:
```suggestion
  Animal.alive = [];
```
let's change it like this and move outside Animal class :)

kasiakubisiak (changes requested):

kasiakubisiak (approved):
nicely done!

PR #1701, opened 2023-04-18
Solution

NadieinOleh:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(health, name);
+  }
+
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
+      herbivore.health -= 50;
+      Animal.alive = Animal.alive.filter(item => item.health !== 0);
```
polosanya, line 31: Better check if animal health is 0 before to avoid extra filtering

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(health, name);
+  }
+
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
polosanya, line 29: It would be great to double check if it's really a herbivore

polosanya (approved):
Good job!

PR #1702, opened 2023-04-18
solution

sylcym:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
+    Animal.alive.push(this);
+  }
```
kasiakubisiak, line 8:
```suggestion
  constructor(name) {
    this.name = name;
    this.health = 100;
    Animal.alive.push(this);
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden) {
+      return;
+    }
+
+    animal.health -= 50;
+
+    if (!animal.health) {
+      Animal.alive = Animal.alive.filter((herbivore) => herbivore.health > 0);
```
kasiakubisiak, line 32: Let's move it to a new method inside Animal class, e.g. 'killAnimal' and then call it here: `Animal.killAnimal();`

kasiakubisiak (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
+    super(name, health);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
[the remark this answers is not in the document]
sylcym: done

src/herbivoresAndCarnivores.js
sylcym: I am not sure what you mean. I have removed the if statemenet, but I'm not sure about moving it to a new static method. According to the instructions `Carnivore has a 'bite' method, which takes a herbivore object and decreases the object's health by 50`, therefore I think this code is part of the bite method

kasiakubisiak: Yes, the solution I provided will actually work the same as yours but will be more understandable because logic for removing a dead animal will be contained in a single method inside Animal class.

So you just create new method inside Animal:
```  
static killAnimal() {
    Animal.alive = Animal.alive.filter((herbivore) => herbivore.health > 0);
  }
```

and inside Carnivore just call this method:
```
animal.health -= 50;
Animal.killAnimal();
```

sylcym: I understand, thank you :)

PR #1704, opened 2023-04-18
solution

DanyloPasko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health, biten) {
+    super(name, health);
+
+    this.biten = biten;
+  }
+
+  bite(object) {
+    if (object.hidden === false) {
```
vlad-demchuk, line 33:
```suggestion
    if (!object.hidden) {
```

also, you should check `instanceof Herbivore`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+      object.health -= 50;
+    }
+    if (object.health <= 0) {
+      const index = Animal.alive.indexOf(object);
+      Animal.alive.splice(index, 1);
```
vlad-demchuk, line 38: use `filter` method instead

vlad-demchuk (changes requested):

IvanFesenko (approved):

PR #1706, opened 2023-04-19
add solution

PufAnimalsDev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 
 class Carnivore extends Animal {
   // write your code here
+  bite(bitten) {
+    if (!bitten.hidden && bitten instanceof Herbivore) {
+      bitten.health -= 50;
+
+      if (!bitten.health) {
+        Animal.alive = Animal.alive.filter((herbivore) => herbivore.health > 0);
+      }
```
kasiakubisiak, line 36: Lets move this to Animal class and put inside new method e.g. 'killAnimal' and then reuse it here 
```suggestion
    animal.killAnimal();
```
also, avoid nesting - putting if inside another if.

kasiakubisiak (changes requested):

kasiakubisiak (approved):

PR #1713, opened 2023-04-22
Fix

marshalek37:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  die() {
```
d-wasilewski, line 10:
```suggestion
  kill() { or killAnimal() {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (!(animal instanceof Herbivore) || animal.hidden) {
```
d-wasilewski, line 34: It's better because it will work as intended in case we have more classes
```suggestion
    if (animal instanceof Carnivore || animal.hidden) {
```

d-wasilewski (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  kill() {
+    Animal.alive = Animal.alive.filter(animal => animal !== this);
```
d-wasilewski, line 11: you should check if the animal's health is greater than 0 in this place

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    if (animal instanceof Carnivore || animal.hidden) {
+      return;
+    }
+
+    animal.health -= 50;
+
+    if (animal.health <= 0) {
+      animal.kill();
+    }
+  }
```
d-wasilewski, line 43: it's better to just do 
```
if(positive condition) {
// your code inside this if
```
You will avoid returning nothing (you don't need it) and the code will be more readable

d-wasilewski (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  doDamage(damage) {
+    this.health -= damage;
+
+    if (this.health <= 0) {
+      Animal.alive = Animal.alive.filter(animal => animal !== this);
+    }
```
d-wasilewski, line 15: It's better than comparing the animal to this
```suggestion
    Animal.alive = Animal.alive.filter(aliveAnimal => aliveAnimal.health > 0);
```

d-wasilewski (changes requested):

PR #1716, opened 2023-04-24
solve

ctgdprzyb:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
```
d-wasilewski, line 26: As mentioned before, it's better to push the animals to the Animal.alive array in the Animal constructor to avoid duplicating the code

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
 }
 
+Animal.alive = [];
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
d-wasilewski, line 15: It's better to push the animals to the Animal.alive array in the Animal constructor, so you can avoid duplicating the code in both Herbivore and Carnivore constructors

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+
+      if (target.health < 1) {
+        Animal.alive = Animal.alive.filter(x => x.health > 0);
```
d-wasilewski, line 34:
```suggestion
        Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```

d-wasilewski, line 34: you can move this part of the code to a function, for example kill ()

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #1718, opened 2023-04-25
solution

yuron-maker:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
polosanya, line 16: It is better to move it to the Animal constructor so as not to duplicate it for Herbivore and Carnivore, as both of them expand Animal

polosanya (changes requested):
Almost done! Let's improve your code a bit)

AlisaVlasova (approved):

PR #1720, opened 2023-04-25
herbivores & carnivores - first try

RufatRM:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
 
 class Animal {
   // write your code here
+
+  constructor(name) {
```
```suggestion
  constructor(name) {
```
Better remove the comments and empty lines

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
...
+  };
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(object) {
```
```suggestion
  bite(target) {
```
Object is a reserved name in JS, better find another name in order not to confuse other developers

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
...
+    if (object.hidden === false) {
+      object.health -= 50;
+    }
+
+    if (object.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(object), 1);
```
polosanya, line 36: image

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,40 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(object) {
+    if (object.hidden === false) {
```
polosanya, line 31: It would be great to double check if it's Herbivore too

polosanya (changes requested):
Almost done! Let's improve your solution a bit)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animalToBite) {
+    if (animalToBite.hidden === false
+      && animalToBite instanceof Herbivore === true) {
```
anstsot, line 28:
```suggestion
    if (!animalToBite.hidden && animalToBite instanceof Herbivore) {
```

anstsot (approved):
Good job, approved!

PR #1722, opened 2023-04-26
initial commit

44shaggy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (!beast.hidden && beast instanceof Herbivore) {
+      beast.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(creature => creature.health > 0);
```
Anastasiia-Svintsova, line 30: Better check if animal health is 0 before to avoid extra filtering

Anastasiia-Svintsova (approved):

PR #1724, opened 2023-04-27
Solution

AnastasiiaVorobets:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static deleteAnimal(dead) {
+    Animal.alive = Animal.alive.filter(animal => animal.name !== dead.name);
```
anstsot, line 11: it's better to check the `health === 0` because name isn't unique variable

anstsot (changes requested):
It’s almost done! Let’s improve a little

slavanosatskyi (approved):

PR #1731, opened 2023-05-03
herbivores&carnivores

vladimirshub:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
agniya-i, line 26:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive = Animal.alive.filter(a => a.health !== 0);
```
agniya-i, line 31:
```suggestion
      Animal.alive = Animal.alive.filter(herbivore => herbivore.health !== 0);
```

`a` is not really good name, better to call it more descriptive, for example `animal`/`herbivore`

agniya-i (approved):

PR #1732, opened 2023-05-03
solutin1

zvir91:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        animal.die();
```
agniya-i, line 39: You can also use the `filter` method here instead of `indexOf` +` splice` for better performance and filter `Animal.alive`: 
`Animal.alive.filter(beast => beast.health > 0);`

agniya-i (approved):

PR #1735, opened 2023-05-04
Solution

92BS:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
```
polosanya, line 7:
```suggestion
```
Remove this property in the Animal class. Only Herbivores have it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+
+    Animal.alive.push(this);
+  }
+
+  die() {
```
polosanya, line 12: 😢

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
polosanya, line 17: image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+  }
+
+  decreaseHealth(amount) {
+    this.health = Math.max(0, this.health - amount);
```
polosanya, line 21: I think there is no need to use `amount` parameter, as health is always decreased by 50 according to the task description

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(h) {
```
polosanya, line 38: Let's use more descriptive names than just `h`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    if (h.hidden || h instanceof Carnivore) {
+      return;
+    }
+    h.decreaseHealth(50);
```
polosanya, line 42:
```suggestion
    if (!h.hidden && h instanceof Carnivore) {
      h.decreaseHealth(50);
    }
```
Just decrease the health only if conditions are true to avoid useless return

polosanya (changes requested):
Good job! Let's improve your solution a bit)

VitaliyBondarenko1982 (approved):

PR #1736, opened 2023-05-04
solution

antonina-klishch:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim.hidden === false && victim instanceof Herbivore) {
```
OleziO, line 29: Replace `victim.hidden === false` with `!victim.hidden`

OleziO (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (!victim.hidden && victim instanceof Herbivore) {
+      victim.health -= 50;
+    }
+
+    if (victim.health === 0) {
+      const deadAnimal = Animal.alive.indexOf(victim);
```
DarkMistyRoom, line 34:
```suggestion
      const deadAnimalIndex = Animal.alive.indexOf(victim);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
+      Animal.alive.splice(deadAnimal, 1);
```
DarkMistyRoom, line 36: another way is to use `filter` method. Just want you to know)

DarkMistyRoom (approved):
Good job!

PR #1739, opened 2023-05-06
Solution

the-tique:
I wasn't able to use static property Animal.alive inside Animal class because of linter. So I moved it outside of class. The other way was to disable linter rule, but I choose the 1st variant.

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && prey.hidden !== true) {
```
SanyaBratashchuk, line 31:
```suggestion
    if (prey instanceof Herbivore && !prey.hidden) {
```

SanyaBratashchuk (approved):

PR #1745, opened 2023-05-08
Solution

yaroslav5481:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      const index = Animal.alive.indexOf(animal);
+      if (index > -1) {
+        Animal.alive.splice(index , 1);
```
tetiana-ivanchenko, line 32: use filter instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
```
tetiana-ivanchenko, line 11:
```suggestion

  constructor(name) {
    this.name = name;
    this.health = 100;
    Animal.alive.push(this);
  }
}

Animal.alive = [];
```

tetiana-ivanchenko: it will helps to pass the tests

tetiana-ivanchenko (changes requested):

PR #1751, opened 2023-05-11
add herbivoresAndCarnivores solution

DidenkoDenis306:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static removeAnimal() {
```
dinaminahi, line 11: I would rather name it something like 'removeDiedAnimals' or 'filterDiedAnimals'

dinaminahi (approved):
Good job 👍

Issue #952, opened 2022-08-10
ESlint rejects static props and methods


sergii-nosachenko:
When trying to declare static props get an ESlint error: `Parsing error: Unexpected token =`.
Fixed only by adding additional configuration line to `.eslintrc.js` file above `extends` entry: `parser: "babel-eslint",`.

Dmytro-Matsehora:
I had the same problem, thanks a lot for sharing your solution!

sergii-nosachenko:
@Dmytro-Matsehora Also you can add static property after class declaration:
```
class ClassName {
  ...
}

ClassName.staticProperty = value;
```

PR #1752, opened 2023-05-11
solution

generatedman:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.name = name;
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.sort((a, b) => a.health - b.health);
```
33Orange, line 35: Instead of sorting and removing the animal with 0 health, you can use the filter method to remove the dead animal from the Animal.alive array

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    super(name);
+    this.name = name;
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
```
33Orange, line 34: The health check should be in the Animal constructor to remove the animal from the alive array whenever its health reaches 0 or below, regardless of the action that caused it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
33Orange, line 4: Let's fix problem with eslint https://github.com/mate-academy/js_herbivores_and_carnivores/issues/952 

You can add static property after class declaration
```
class ClassName {
  ...
}

ClassName.staticProperty = value;
```

33Orange (changes requested):
Some comments:
- Let's fix problem with eslint https://github.com/mate-academy/js_herbivores_and_carnivores/issues/952 
- Instead of checking the health of animal when bitten, lets create a cleanup method `removeDiedAnimals` in the Animals class. And use this method.
- Use method `filter` instead `sort` + `shift`.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.name = name;
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
naziamoff, line 34: let's use shortcuts for booleans
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+    Animal.alive.push(this);
+  }
+
+  static removeDiedAnimals() {
```
naziamoff, line 10:
```suggestion
  static removeDeadAnimals() {
```

naziamoff (approved):
looks good 👍

PR #1759, opened 2023-05-15
Solution

4ykis:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
+
+  decreaseHealth(amount = 50) {
+    this.health -= amount;
+
+    if (this.health <= 0) {
+      this.die();
+    }
+  }
+
+  die() {
```
volodymyr-soltys97, line 19: die method should be private, consider renaming it to _die.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    Animal.alive.push(this);
+  }
+
+  decreaseHealth(amount = 50) {
```
volodymyr-soltys97, line 11: decreaseHealth method should be private, consider renaming it to _decreaseHealth.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
```
volodymyr-soltys97, line 36: Herbivore's hide method should toggle the hidden state instead of always setting it to true.

volodymyr-soltys97 (changes requested):
Great 👍 
But let's make your code more sophisticated

nazarmatsevych (approved):

PR #1760, opened 2023-05-15
Sacrifice

BogdanParubchenk01:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(sacrifice) {
+    if (sacrifice instanceof Herbivore && !sacrifice.hidden) {
+      sacrifice.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(beastie => beastie.health > 0);
```
Viktor-Kost, line 29: no needs to filter alive animals each time after a bite. Add a condition like `if (sacrifice <= 0) {}`

Viktor-Kost (changes requested):
Great job!
A minor improvement needed

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(sacrifice) {
+    if (sacrifice instanceof Herbivore && !sacrifice.hidden) {
+      sacrifice.health -= 50;
+    }
+
+    if (sacrifice.health === 0) {
```
tetiana-ivanchenko, line 30:
```suggestion
    if (sacrifice.health <= 0) {
```

tetiana-ivanchenko (approved):

PR #1762, opened 2023-05-16
add task solution

maksym-nemera:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,46 @@
...
+  }
+
+  static addAnimal(animal) {
+    this.alive.push(animal);
+  }
+
+  static removeAnimal(animal) {
+    const index = Animal.alive.indexOf(animal);
+
+    Animal.alive.splice(index, 1);
```
33Orange, line 19: `[CODE PERFOMANCE] - Try to avoid nesting loops (indexOf + splice), use filter instead`

33Orange (changes requested):
Well done!

Just a one moment with `removeAnimal` method. Try to use `.filter()` instead 2 loops (indexOf and splice).

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+    this.hidden = false;
+
+    Animal.addAnimal(this);
+  }
+
+  static addAnimal(animal) {
+    this.alive.push(animal);
+  }
+
+  static removeAnimal(animal) {
```
dinaminahi, line 16: here you are comparing object to object(reference to reference), either we can't compare their names, because we can have animals with same names.  
Let's rename this method to `removeDeadAnimals`, and filter animals which have health > 0

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore
+          && animal.hidden === false) {
```
dinaminahi, line 30: `animal.hidden === false`    ->    `!animal.hidden`

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
dinaminahi, line 34: Better check here would be `animal.health <= 0`

dinaminahi (changes requested):

dinaminahi (approved):
Well done 👍

PR #1763, opened 2023-05-16
1st commit

ihor-halkin:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
```
polosanya, line 25:
```suggestion
  bite(target) {
```
Let's use more meaningful names

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    if (obj.hidden === false) {
+      obj.health -= 50;
+    }
+
+    const filteredAnimal = Animal.alive.filter(({ health }) => health > 0);
+
+    Animal.alive = filteredAnimal;
```
polosanya, line 32:
```suggestion
    Animal.alive = Animal.alive.filter(({ health }) => health > 0);
```
Also, it would be great to add a check if target's health is equal to zero to avoid extra method calls

polosanya (changes requested):
Almost perfect

PR #1769, opened 2023-05-21
solution

IamLoren:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
```
TarasHoliuk, line 14:
```suggestion
  constructor(name, health, hidden = false) {
```
I guess `hidden` shouldn't be in the constructor
Anyway not critical here, so approved

TarasHoliuk (approved):

PR #1773, opened 2023-05-22
solution

HYPER2307:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+      ? this.hidden = false
+      : this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+    this.name = name;
```
etojeDenys, line 31: The assignment to this.name is redundant since it's already done in the parent class

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #1775, opened 2023-05-24
Solution

NickBrodyuk:


.eslintrc.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   extends: '@mate-academy/eslint-config',
-  parser: "babel-eslint",
```
volodymyr-soltys97, line 3: The 'parser' property is missing. If you are using Babel, add 'parser: "babel-eslint"' to the configuration.

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal.health === 0), 1);
```
volodymyr-soltys97, line 30: The splice method should remove the dead animal, not the first animal with health equal to 0. Replace 'animal.health === 0' with 'animal' in the indexOf method.

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,39 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
volodymyr-soltys97, line 29: Consider checking if the animal's health is less than or equal to 0 instead of only equal to 0, to handle cases where health might become negative.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

.eslintrc.js
NickBrodyuk: >  If you are using Babel,
Am I?

volodymyr-soltys97 (approved):
Awesome 🚀

PR #1777, opened 2023-05-25
solution

okosohor:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+
+      Animal.alive = Animal.alive.filter(checkAnimal => checkAnimal.health > 0);
```
33Orange, line 29: Let's move the `delete` method into a instance of the `Animal` class
And add a check that if the animal you bite has health 0 or less, then we delete it

```
if (animal.health <= 0) {
}
```

33Orange (changes requested):
Good job.

Just a one moment

dinaminahi (approved):
Well done 👍

PR #1778, opened 2023-05-25
Solution

julshvets:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (!beast.hidden && !(beast instanceof Carnivore)) {
+      beast.health -= 50;
+    }
+
+    if (beast.health === 0) {
+      Animal.alive = Animal.alive.filter(creature => creature !== beast);
```
33Orange, line 33: And change condition
Let's move the `delete` method into a instance of the `Animal` class

```
if (animal.health <= 0) {
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    if (!this.hidden) {
+      this.hidden = true;
+    }
+  };
+};
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (!beast.hidden && !(beast instanceof Carnivore)) {
```
33Orange, line 28:
```suggestion
    if (!beast.hidden && beast instanceof Herbivore) {
```

33Orange (changes requested):
Well done !

A few moments to fix

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  static removeFromAlive(animal) {
+    this.alive = this.alive.filter(creature => creature !== animal);
```
dinaminahi, line 11: here you are comparing object to object(reference to reference), either we can't compare their names, because we can have animals with same names.
Let's rename this method to `removeDeadAnimals`, and filter animals which have health > 0

dinaminahi (changes requested):
Good job! 👍 
But still one fix needed

src/herbivoresAndCarnivores.js
julshvets: Done.

dinaminahi (approved):
Well done 👍

PR #1779, opened 2023-05-25
add task solution

KatyaZaremska:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(el => el.health !== 0);
```
33Orange, line 34: Let's move the `delete` method into a instance of the `Animal` class
And add a check that if the animal you bite has health 0 or less, then we delete it

```
if (animal.health <= 0) {
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
33Orange, line 15: You don't need to accept hidden as an argument in the constructor
```suggestion
  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    super(name, health);
+    this.hidden = hidden;
```
33Orange, line 17:
```suggestion
    this.hidden = false;
```

33Orange (changes requested):
Well done!

A few moments to fix

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.removeFromAlive(animal);
+    }
+  }
```
33Orange, line 22: Why you move `bite` method to instance of `Animal` ?
Only `Carnivore` can bite

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static removeFromAlive(animal) {
+    Animal.alive = Animal.alive.filter((el) => el.health > 0);
+  }
```
33Orange, line 12: Almost done.
But why do we need the animal parameter in this method if it is not used?
Use more descriptive names in callbacks

```suggestion
  static removeFromAlive() {
    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
  }
```

33Orange (changes requested):
A few moments to fix

PR #1786, opened 2023-05-29
solution

Yuliya1907:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (!animal.health) {
+      const index = Animal.alive.indexOf(animal);
```
naziamoff, line 32: what if we add some methods which gonna give not 50, but let's say 27 damage? 
If you'll get `-3` health on some animal - they won't be removed, as negative values are `truthy`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
```
naziamoff, line 27: why `!(animal instanceof Carnivore)`, but not `animal instanceof Herbivore`?

naziamoff (approved):
well done 🔥 
A couple of small suggestions:

PR #1787, opened 2023-05-29
add task solution

EugeneSherepa:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    Animal.alive.push(this);
+  }
+
+  static restInPeace() {
```
33Orange, line 11: That's funny.
But let's use more descriptive names

33Orange (approved):
Good job

PR #1788, opened 2023-05-29
solution

artem-ivashchenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
mykhalenych, line 21: The hide method should toggle the hidden property, not just set it to true.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
mykhalenych, line 27:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

mykhalenych (changes requested):

sTorba24 (approved):
GJ 🚀

PR #1789, opened 2023-05-29
Solution

amashutikov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+};
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
```
naziamoff, line 23: is this method supposed to toggle visibility?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && !(animal instanceof Carnivore)) {
```
naziamoff, line 30: it makes sense to check what the animal is rather than what it isn't

Also, you are trying to check whether the animal is hidden even before getting to know you're dealing with Carnivore/Herbivore.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
+Animal.removeDeadAnimal = () => {
```
naziamoff, line 13: by its name to me it seems that this method should accept some animal and remove it from the list, not filtering the animals

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && !(animal instanceof Carnivore)) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
```
naziamoff, line 33: what if later we add some method which deals 27 damage?

naziamoff (changes requested):

naziamoff (approved):
GJ 👍

PR #1790, opened 2023-05-29
Solution

bohdan-mykhailenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
+    }
+
+    if (prey.health === 0) {
```
naziamoff, line 32: what if later we add some method that deals 27 damage?

naziamoff (approved):
looks good 🔥 
One suggestion though

PR #1791, opened 2023-05-29
add solution

ulositska:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+    }
+
+    if (victim.health === 0) {
+      Animal.alive = Animal.alive.filter(animal => animal !== victim);
+    }
```
naziamoff, line 29: what if health is less than 0? I mean yeah, now we have only one method for giving a damage and it's 50, so there's not chance it's gonna be less than 0. But what if we add some later?

naziamoff (approved):
looks good 👍 
One small suggestion

PR #1792, opened 2023-05-29
add solution

growingananas:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
+
+  set health(value) {
+    this._health = value;
+    Animal.removeDeadAnimals();
+  }
```
33Orange, line 31: A setter is used to set a value, or to process a received value and set it to an object.
Why you use `Animal.removeDeadAnimals()` here?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    Animal.removeDeadAnimals();
```
33Orange, line 44: Let's call this method inly if `animal.health <= 0`
```suggestion
    if (animal.heath <=0) {
      Animal.removeDeadAnimals();
    }
```

33Orange (changes requested):
Good job.

A few moments

src/herbivoresAndCarnivores.js
growingananas: I agree, it's not necessary

33Orange (approved):
Well done !

PR #1793, opened 2023-05-29
Animals created

olehmahera:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (!(target instanceof Carnivore) && target.hidden !== true) {
```
naziamoff, line 27: why you check whether the animal is not carnivore? To me it makes sense to check whether it is the Herbivore

naziamoff: also, why `!== true`? Why not `===` false?

Anyway, let's use shortcut for this boolean, like `if (boolean)` of `if (!boolean)`

naziamoff (changes requested):
good job, but there's a couple of things to polish

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore) {
```
naziamoff, line 27: can be inside a single `if`

naziamoff (approved):
well done 👍

PR #1794, opened 2023-05-29
Solution

Shevtsooov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && !beast.hidden) {
+      beast.health -= 50;
+    }
+
+    if (beast.health <= 0) {
+      Animal.alive = Animal.alive.filter(animal => animal.name !== beast.name);
```
33Orange, line 32: Let's move this logic to method in instance of `Animal` class. 
You can simply clear the array in this filter from animals that have health <= 0

```suggestion
      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```

33Orange (changes requested):
Good job.

Just a one moment

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  removeAnimal() {
```
dinaminahi, line 10:
```suggestion
  removeDeadAnimals() {
```

dinaminahi (approved):
Good job! 👍

src/herbivoresAndCarnivores.js
Shevtsooov: Thanks!

PR #1796, opened 2023-05-29
add solution

niaYaro:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.animalsCreated.push(this);
+  }
 }
 
+Animal.animalsCreated = [];
```
naziamoff, line 16: does `created` in the name of the variable give us any necessary context?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static get alive() {
```
naziamoff, line 4: I don't think this should be a getter, that's just the method we use inside our class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim.constructor.name === 'Herbivore' && victim.hidden !== true) {
```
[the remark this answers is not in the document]
naziamoff: `victim.hidden !== true` - looks like a crutch 😄

naziamoff (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden !== true) {
```
naziamoff, line 24: let's use shortcuts for boolean
```suggestion
    if (victim instanceof Herbivore && victim.hidden !== true) {
```
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
```

naziamoff (approved):
👍

PR #1797, opened 2023-05-29
Solution

annholovko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden !== true) {
```
naziamoff, line 30: probably you mean this? 😄 
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

naziamoff (approved):
looks good 👍

PR #1798, opened 2023-05-29
Herbivores and Сarnivores solution

PANCH1811:


33Orange (changes requested):
Good job.

But a property hidden exit only in Herbivores.
Right now you Carnivores have property hidden too
Let's fix it and we're good

33Orange (approved):
Well done !

PR #1800, opened 2023-05-29
Solution

ivankhrsh:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    this.health = 100;
+    this.hidden = false;
+    Animal.animalBorn(this);
+  }
+
+  static alive = [];
+
+  static animalBorn(animal) {
+    this.alive.push(animal);
+  }
```
naziamoff, line 15: does not sound like a name for the method, but rather for the variable

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  static animalDie(animal) {
```
naziamoff, line 17: what does this method do? I can't guess by its name

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    Animal.alive = Animal.alive.filter(element => element !== animal);
```
naziamoff, line 18: element?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    const canBite = (animal instanceof Herbivore) && animal.hidden === false;
+
+    animal.health = canBite
+      ? animal.health - 50
+      : animal.health;
```
naziamoff, line 34: do you need this action? I guess you should perform actions only if `canBite`, no other actions required if `!canBite`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
33Orange, line 7: Only Herbivores have `hidden` flag.
Right now your Carnivores will be have it too.

`Create a Herbivore class. Herbivore has a method of hide, which changes the hidden property of the beast and helps to hide from carnivores.`

33Orange (changes requested):
Almost done

Just a one moment to fix

naziamoff (approved):
🔥

PR #1801, opened 2023-05-29
add task solution

opiquem:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
naziamoff, line 31: is this a herbivore expected here? I don't think so, as you check it in the if statement. Let's use other name for this variable

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    if (!herbivore.hidden
+      && herbivore instanceof Herbivore) {
```
naziamoff, line 33: if we move the if condition to the new line - we do it like that:
```suggestion
    if (
      !herbivore.hidden
      && herbivore instanceof Herbivore
    ) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+
+    Animal.alive.push(this);
+  }
+
+  static checkLifeStatus() {
```
naziamoff, line 11: what does this method do? By its name I don't expect it to change the state somehow

naziamoff (changes requested):
not bad, but there are a couple of code style suggestions

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (
+      !beast.hidden
+      && beast instanceof Herbivore
```
naziamoff, line 34: I believe would be better to switch their position, as here you have redundant check for hidden if the Carnivor is passed

naziamoff (approved):
good job!

PR #1802, opened 2023-05-29
solution

serentinos:


serentinos:
Invalid checks because of linter
image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  static newAnimal(animalObj) {
+    this.alive.push(animalObj);
+  }
+
+  static animalDeath(animalObj) {
```
naziamoff, line 10: is this a name for method, or for the variable?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  static newAnimal(animalObj) {
```
naziamoff, line 6: is this a name for method, or for the variable?

src/herbivoresAndCarnivores.js
naziamoff, line 10: don't use data type in the variable name

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+
+    this.hidden = false;
+  }
+
+  hide() {
```
naziamoff, line 33: is `hide` method supposed to be toggled, or it is meant just for hiding?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    const bitePower = 50;
+    const isHerbivore = Object.hasOwnProperty.call(victim, 'hidden');
```
naziamoff, line 41: there's a special method for checking what instance our object is. Checking for some exact property is a crutch.
What if we add the `hidden` property to the Carnivores?

naziamoff (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static addToAlive(animalObj) {
+    this.alive.push(animalObj);
+  }
+
+  static removeFromAlive(animalObj) {
```
naziamoff, line 8: - let's use more descriptive names for your variables. It should be maximally clear what your method does by its name
```suggestion
  static removeDeadAnimal(animalObj) {
```

naziamoff: also, you should not use data type in the name of your variables, as it does not give us any additional information. It should have clear naming even without explicit mention of data type

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    const bitePower = 50;
+    const isHerbivore = victim instanceof Herbivore;
+
+    if (!victim.hidden && isHerbivore) {
```
naziamoff, line 43: better to switch the places. If Carnivore is passed you have redundant check for `hidden`

src/herbivoresAndCarnivores.js
serentinos: Thank you for suggestion!

33Orange (approved):
Well done !

naziamoff (approved):

PR #1804, opened 2023-05-29
solution

snuber-w:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    Animal.alive.push(this);
+  }
+
+  bite(beast) {
+    if ((beast instanceof Herbivore) && (beast.hidden === false)) {
```
naziamoff, line 33: - do you really need round braces here?
- let's use shortcuts for booleans `boolean || !boolean`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+      beast.health -= 50;
+      Animal.alive = Animal.alive.filter(organism => organism.health > 0);
```
naziamoff, line 35: why organism, but not animal?

naziamoff (approved):
good job! A couple of small suggestions though

src/herbivoresAndCarnivores.js
snuber-w: I know that these round brackets are not necessary, I thought they would improve readability

src/herbivoresAndCarnivores.js
snuber-w: did not want to repeat the names of variables and classes and invented new

PR #1805, opened 2023-05-29
made task js herbivores and carnivores

vladyslav-lapshyn:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim.hidden !== true && victim instanceof Herbivore) {
```
naziamoff, line 36: why you check for the hidden first? If we pass Carnivore - they don't have such a property

naziamoff (changes requested):
not bad, but there's a couple of things to polish

naziamoff (approved):
👍

PR #1806, opened 2023-05-29
Solution

MaksimKad:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static murderousSort() {
+    this.alive = this.alive.filter(animal => animal.health > 0);
+  }
```
naziamoff, line 12: wow, what a name for the method 😄 
Why not just `filterDeadAnimals`?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
```
naziamoff, line 30: what if we pass Carnivore? You will firstly check whether it is hidden, but it doesn't have such a property. Typescript would bring you an error in such case

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
naziamoff, line 34: What if later we add new method with different damage, like 23? This way animal health can be < 0 and the animal won't be filtered

As well there's a shorthand syntax for booleans

naziamoff (changes requested):

dinaminahi (approved):
Good job 👍

naziamoff (approved):

PR #1810, opened 2023-05-29
Solution

Mykola-Maik:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
+Animal.removeAnimal = function(objToRemove) {
+  delete this.alive[Animal.alive.indexOf(objToRemove)];
```
naziamoff, line 14: better to store something in the constants, as it now does not appear to be readable

naziamoff: also, do you believe `objToRemove` is a good name for a variable?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
+    if (!obj.hidden && !(obj instanceof Carnivore)) {
```
naziamoff, line 29: obj?

naziamoff: also, it it better to check what the animal is, rather than what it isn't

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+      obj.health -= 50;
+
+      if (obj.health === 0) {
```
naziamoff, line 32: what if we later add some method which would deal `27` damage?

naziamoff (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
+const indexOfDeadAnimal = function(animalToRemove) {
```
naziamoff, line 13: is there any reason to assign a function declaration to the variable? It can be stored just as a variable, it seems to me

PR #1811, opened 2023-05-29
solution

SofiaLiubchenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
```
naziamoff, line 30: what if later we add method which deals let's say `27` damage. This way it becomes possible that the `health` is less than 0. We should take into the account that case

naziamoff (approved):
good job!

PR #1812, opened 2023-05-29
Solution

Yuliia-kucheriavenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    const decreaseHealthBy = 50;
+
+    if (!animal.hidden && !(animal instanceof Carnivore)) {
```
naziamoff, line 26: does this make sense to firstly check whether the animal is hidden? What if we pass Carnivore there? Typescript would bring you an error in such case

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    const decreaseHealthBy = 50;
```
naziamoff, line 24: sounds more like a name for a method, but not for a variable. Let's think of something more descriptive

src/herbivoresAndCarnivores.js
naziamoff: also, why you don't just check whether the animal is instanceof Herbivore, but you check whether they are not instanceof Carnivore?

naziamoff (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= 50;
+
+      Animal.alive = Animal.alive
+        .filter(value => value.health > 0);
+    }
```
33Orange, line 29: Would be better to filter alive array only when the animal heath `<= 0`
    
```suggestion
    if (!animal.hidden && animal instanceof Herbivore) {
      animal.health -= 50;
    }
    
    if(alimal.heath <= 0) {
      Animal.alive = Animal.alive.filter(value => value.health > 0);
    }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
```
33Orange, line 24: You need to check if animal is instanceof Herbivore and only after this check `hidden` flag.

```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

33Orange (changes requested):
Almost done

PR #1813, opened 2023-05-29
solution

vitvadym:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, damage = 50) {
+    super(name);
+    this.damage = damage;
+  }
```
33Orange, line 34: I do not see in the conditions of the task that the damage of a predator can be different. Let's do everything according to the conditions
```suggestion

```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+
+  bite(animal) {
+    animal instanceof Herbivore && !animal.hidden
+      ? animal.health -= this.damage
+      : animal.health = this.health;
```
33Orange, line 39: Why you use ternary operator here?
We use ternary operator only if we do smth in every case, but here you do smth only if condition `true`.

```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
      animal.health -= this.damage
    }
```

33Orange (changes requested):
Good job.

A few moments to fix

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  static removeDiedAmimals() {
```
silchyn, line 10:
```suggestion
  static removeDeadAmimals() {
```

silchyn (approved):

PR #1814, opened 2023-05-30
Solution

maryanmats:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
+Animal.removeDeadAnimal = () => {
```
naziamoff, line 13: by its name it seems to me that this method should expect some animal as an arg and remove it from the list

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && !(animal instanceof Carnivore)) {
```
naziamoff, line 30: - you check whether the animal is hidden even before getting to know who are you dealing with
- let's check what the animal is instead of checking what it isn't

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+      animal.health -= 50;
+
+      if (animal.health === 0) {
```
naziamoff, line 33: what if later we add some method which deals 27 damage?

naziamoff (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
+Animal.removeDeadAnimal = (animal) => {
+  Animal.alive = Animal.alive.filter(beast => beast !== animal);
+};
```
dinaminahi, line 15: Now you are comparing object to object(reference to reference)
I think @naziamoff was expecting to change the name of the method to `removeDeadAnimals` or `filterDeadAnimal`

dinaminahi (changes requested):
Almost done, let's fix one thing

naziamoff (approved):

PR #1815, opened 2023-05-30
Add Solution

vchervinko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
```
naziamoff, line 30: would be more logical to check whether the animal is herbivore firstly, as Carnivores don't have hidden property

naziamoff (approved):
looks good 👍

PR #1816, opened 2023-05-30
added task solution

holytimy4:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && !beast.hidden) {
+      beast.health -= 50;
+    }
+
+    if (!beast.health) {
```
dinaminahi, line 35: What if beast health will be less than 0?
```suggestion
    if (beast.health <= 0) {
```

dinaminahi (changes requested):
Well done  👍 

A few moments

33Orange (approved):
Well done !

PR #1818, opened 2023-05-30
Develop

domkab:


src/herbivoresAndCarnivores.js
```diff
@@ -1,19 +1,48 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor (name, health = 100, lives = true) {
```
```suggestion
  constructor (name, health = 100) {
```
Why do you need `lives` boolean? It's actually redundant

src/herbivoresAndCarnivores.js
```diff
@@ -1,19 +1,48 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor (name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide () {
+    this.hidden = !this.hidden;
+  }
```
polosanya, line 21: image

src/herbivoresAndCarnivores.js
```diff
@@ -1,19 +1,48 @@
...
+    super(name, health, hidden);
+  }
+
+  bite (animal) {
+    if (animal instanceof Carnivore) {
+      return;
+    }
+
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
```
polosanya, line 35:
```suggestion
  bite (animal) {
    if (animal instanceof Herbivore && !animal.hidden) {
      animal.health -= 50;
```
This check is actually redundant. You will do nothing if animal is Carnivore anyway

polosanya (changes requested):

src/herbivoresAndCarnivores.js
domkab: I remove this check and I get error for Carnivore tests

witflash (approved):
👍🏼

PR #1819, opened 2023-05-30
Solution

dimoshen:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
+      if (animal.health === 0) {
+        const animalIndex = Animal.alive.indexOf(animal);
+
+        Animal.alive.splice(animalIndex, 1);
+      }
+    }
+  }
 }
 
+Animal.alive = [];
```
33Orange, line 36: Will be better to move this immediately after the declaration of the Animal class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
33Orange, line 11: Do we need to add property `hidden` as an argument to the constructor?
The hidden flag must always be `false` on creation
```suggestion
  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (Object.getPrototypeOf(animal) !== Carnivore.prototype
+    && animal.hidden !== true) {
```
33Orange, line 24: Use `instanceof` to check instead your construction.
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Operators/instanceof

You can simplify the check for the hidden flag

```suggestion
    if (animal instanceof Herbivores && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
+      animal.health -= 50;
+
+      if (animal.health === 0) {
+        const animalIndex = Animal.alive.indexOf(animal);
+
+        Animal.alive.splice(animalIndex, 1);
```
33Orange, line 30: Move this code to separated method in instance of `Animal` class and use it here.
Let's change this to filter. Checklist p2.

`[CODE PERFOMANCE] - Try to avoid nesting loops (indexOf + splice), use filter instead`

33Orange (changes requested):
Good job.

A few moments to fix

src/herbivoresAndCarnivores.js
dimoshen: fixed

33Orange (approved):

PR #1821, opened 2023-05-30
solution

thienlaoo:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
33Orange, line 4: https://github.com/mate-academy/js_herbivores_and_carnivores/issues/952#issuecomment-1219513446
You can add static property as in issue comment to avoid test fails

```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      const index = Animal.alive.indexOf(animal);
+
+      Animal.alive.splice(index, 1);
```
33Orange, line 31: Let's change it to method filter and clear `alive` array from animals with health `<= 0`
Move this logic to separated method in instance of `Animal` class and use it here as method of Animal

33Orange (changes requested):
Good job.

A few moments to fix

33Orange (approved):
Well done

PR #1824, opened 2023-05-31
Solution

vlkzmn:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
+        Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
polosanya, line 32: image

polosanya (changes requested):
Almost done!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+
+      Animal.alive = Animal.alive.filter(item => item.health > 0);
```
polosanya, line 31: It would be great to check if the animal's health is zero before to avoid extra calls

polosanya (approved):
👍

PR #1826, opened 2023-06-01
Solution of task JS practise Herbivores and Carnivores

vanya-kalyenichenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
```
loralevitska, line 26: It is recommended to use 'herbivore.health = Math.max(0, herbivore.health - 50)' to prevent the health from going below 0.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    }
+
+    if (herbivore.health === 0) {
```
loralevitska, line 29: The condition should be 'herbivore.health <= 0' to handle cases where the health goes below 0.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+      Animal.alive = Animal.alive.filter((animal) => animal !== herbivore);
```
loralevitska, line 30: Filter by `health` value

loralevitska (changes requested):
Almost perfect) Let's improve

loralevitska (approved):
Good job!

PR #1827, opened 2023-06-01
Solution

vitalii-fedusov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden && !(herbivore instanceof Carnivore)) {
+      herbivore.health -= 50;
```
VitaliyBondarenko1982, line 26: The bite method should decrease the health of the herbivore only if the herbivore is an instance of Herbivore. Replace !(herbivore instanceof Carnivore) with (herbivore instanceof Herbivore).

VitaliyBondarenko1982 (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
```
etojeDenys, line 12: The constructor of the Herbivore class should have a 'name' parameter and pass it to the 'super()' call, like this: 'super(name);'

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
etojeDenys, line 24: The constructor of the Carnivore class should have a 'name' parameter and pass it to the 'super()' call, like this: 'super(name);'

etojeDenys (approved):

PR #1831, opened 2023-06-05
Solution

myplekan:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
```
anastasiiavorobiova, line 21: The parameter name should be replaced with something like 'target' to avoid confusion with the name property of an animal.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+    if (name.hidden === false && name instanceof Herbivore) {
+      name.health -= 50;
+    }
+
+    if (name.health === 0) {
+      const index = Animal.alive.indexOf(this);
+
+      Animal.alive = Animal.alive.splice(index, 1);
```
anastasiiavorobiova, line 29: Use Animal.alive.splice(index, 1) instead of Animal.alive = Animal.alive.splice(index, 1) to remove the dead animal from the array.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
+    if (name.hidden === false && name instanceof Herbivore) {
+      name.health -= 50;
+    }
+
+    if (name.health === 0) {
+      const index = Animal.alive.indexOf(this);
```
anastasiiavorobiova, line 27: You need to find the index of the target in Animal.alive, not 'this'.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
anastasiiavorobiova, line 7: The 'hidden' property should be in the Herbivore class only, not in Animal class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
+    if (name.hidden === false && name instanceof Herbivore) {
+      name.health -= 50;
+    }
+
+    if (name.health === 0) {
```
anastasiiavorobiova, line 26: You should check if the target's health is less than or equal to 0 instead of equal to 0.

anastasiiavorobiova (changes requested):
Great job
Needs a bit of improvement

PR #1832, opened 2023-06-05
solution

Gosyav:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health === 0) {
+      const index = Animal.alive.indexOf(this);
+
+      Animal.alive = Animal.alive.splice(index, 1);
```
IvanFesenko, line 34: use `Array.filter()` instead

IvanFesenko (changes requested):
Good job. Let's add smal improvement

Viktor-Kost (approved):

PR #1834, opened 2023-06-05
Solution

NVlasenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100, hidden) {
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
```
polosanya, line 8: Not all animals have `hidden` property. It appears only for Herbivores

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
```
polosanya, line 20: Let's find a better name for animal than `obj`

polosanya (changes requested):
Almost done!

demetrxx (approved):
Well done 👍 Go ahead!

PR #1842, opened 2023-06-07
Solution

Sergey8793:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden !== true && animal instanceof Herbivore) {
```
OleziO, line 27: Replace `animal.hidden !== true` with `!animal.hidden`. It will be the same

OleziO (changes requested):

OleziO (approved):

PR #1847, opened 2023-06-11
animals

diana-tuz:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    }
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
+      animal.health -= 50;
+    }
```
nazarmatsevych, line 33:
```suggestion
    if (!animal.hidden && animal instanceof Herbivore) {
      animal.health -= 50;
    }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+    this.hidden = hidden;
+  }
+  hide() {
+    if (this.hidden === true) {
+      this.hidden = false;
+    } else {
+      this.hidden = true;
+    }
+  }
```
```suggestion
  hide() {
    this.hidden = true;
  }
```

This method should only set hidden value to true

nazarmatsevych (changes requested):
Almost done!
Pls check comments)

nazarmatsevych (approved):
👍

PR #1848, opened 2023-06-11
animals function

pos1bl:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
```suggestion
  hide() {
    this.hidden = true;
  }
```

This method should hide animal, not toggle hidden value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  bite(animal) {
+    if (animal.hidden !== true && !(animal instanceof Carnivore)) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
+        const id = Animal.alive.indexOf(animal);
+        
+        Animal.alive.splice(id, 1);
+      }
```
nazarmatsevych, line 37: Don't manipulate with array in this way, it's better to use filter method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (animal.hidden !== true && !(animal instanceof Carnivore)) {
```
```suggestion
    if (!animal.hidden && animal instanceof Herbivore) {
```

nazarmatsevych (changes requested):
Good work!

Pls check comments to improve your solution

nazarmatsevych (changes requested):
* test

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
```
volodymyr-soltys97, line 19: The hide method should be able to toggle the 'hidden' property of the Herbivore instance. Consider changing the implementation to 'this.hidden = !this.hidden;'

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/herbivoresAndCarnivores.js
pos1bl: image
Ви знущаеєтесь?)

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1850, opened 2023-06-12
Solution

Sergej161995:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
mykhalenych, line 21: For better flexibility, toggle the hidden property state instead of setting it to true every time hide() method is called.

mykhalenych (approved):

PR #1851, opened 2023-06-12
Rules of beasts

Oksana-Maruniak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    Animal.alive.push(this);
+
+    this.name = name;
+    this.health = health;
```
OleziO, line 8: Push it after setting fields.

OleziO (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  constructor(name, health) {
+    super();
+  }
+
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health === 0) {
```
SerhiiKirik, line 36: It Would be better, bc here you handle all cases
```suggestion
    if (herbivore.health <= 0) {
```

SerhiiKirik (approved):
Nice 🔥

PR #1852, opened 2023-06-13
Add animals

AmadeuszAndroid:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    if (Animal.alive === undefined) {
```
darokrk, line 8: You don't need to check if Animal.alive is undefined. You can initialize it as a static property directly.

`static alive = [];` and add it before `constructor()`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
darokrk, line 39: Think we don't want to filter out all animals but only each bited animal, so think this method should be rather: 

```
if (herbivore.health <= 0) {
const index = Animal.alive.indexOf(herbivore);
    if (index !== -1) {
      Animal.alive.splice(index, 1);
    }
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
darokrk, line 4:
```suggestion
  constructor(name) {
  this.name = name;
  this.health = 100
```

Think you dont need this health value as parameter bcs every animal will start with 100 health by default

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+
+    if (health > 0) {
+      Animal.alive.push(this);
+    }
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
darokrk, line 19: you can also remove `health` from here and from `super()` bcs it will be inherited from Animal class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
darokrk, line 30: Same as above on Herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.name = name;
+    this.health = health;
+
+    if (Animal.alive === undefined) {
+      Animal.alive = [];
+    }
+
+    if (health > 0) {
```
darokrk, line 12: then if we setted up always health animal by 100 we always want to push this animal. Remove this if and move `Animal.alive.push(this);` to the constructor right after `this.health = 100;`

```
this.health = health;
Animal.alive.push(this);
```

darokrk (changes requested):
It's good but small fixes needed :)

src/herbivoresAndCarnivores.js
AmadeuszAndroid: I set health to 100 in body.

src/herbivoresAndCarnivores.js
AmadeuszAndroid: I set health to 100 in Animal constructor body.

src/herbivoresAndCarnivores.js
AmadeuszAndroid: I set health to 100 in Animal constructor body.

src/herbivoresAndCarnivores.js
AmadeuszAndroid: Linter on GitHub [does not allow static fields](https://github.com/mate-academy/js_herbivores_and_carnivores/actions/runs/5256564884/jobs/9498072968?pr=1852#step:5:15). I tried also use `Animal.alive` as property and this throws Linter error locally.

Radoslaw-Czerniawski (approved):
Done :)

PR #1853, opened 2023-06-13
add task solution

Vitalii-Maistrenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
```
polosanya, line 30: Isn't it better to check if the animal has an instance of Herbivore instead of checking the objection of being a Carnivore?

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
+      animal.health -= 50;
+      Animal.removeDeadAnimals();
```
polosanya, line 32: It would be great to check if animal's health is equal to zero before calling the method to avoid extra calls

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
+    }
+  }
 }
 
+Animal.removeDeadAnimals();
```
polosanya, line 37: What's the purpose of calling this method twice?

polosanya (changes requested):
Almost perfect!

PR #1859, opened 2023-06-19
solution

vshepel:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
+  decreaseHealth(num) {
+    const posOfAll = Animal.alive.indexOf(this);
+
+    this.health -= num;
+
+    if (this.health <= 0 && posOfAll > -1) {
+      Animal.alive.splice(posOfAll, 1);
+    }
+  }
```
IvanFesenko, line 19: Move this logic into Carnivore bite method.
Also use Array.filter instead of splice

IvanFesenko (changes requested):

src/herbivoresAndCarnivores.js
vshepel: @IvanFesenko Can you describe why splice method is not appropriate for this?

IvanFesenko: If you have any questions - ask them in fe_chat

IvanFesenko (changes requested):
Not fixed

mykhalenych (approved):

PR #1860, opened 2023-06-20
Solution

s3rg4sh:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hasOwnProperty('hidden') && !herbivore.hidden) {
```
Viktor-Kost, line 24: Replace the 'hasOwnProperty' check with an instanceof check to make sure the passed object is an instance of Herbivore. For example: 'if (herbivore instanceof Herbivore && !herbivore.hidden)'.

Viktor-Kost (approved):
Nice!
Check a comment I left for you

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
Viktor-Kost, line 28: add if condition before filtering 
if (animal.health <= 0)

Viktor-Kost (changes requested):
.

etojeDenys (approved):
LGTM

PR #1867, opened 2023-06-22
Solution

reseneweb:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
 'use strict';
 
+const alive = [];
```
agniya-i, line 3: Make it a static field of Animal
```suggestion
Animal.alive = [];
```

agniya-i (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden) {
+      return;
+    }
+    animal.decreaseHealth(50);
```
witflash, line 45: You can do it vice versa:
```
if (!(animal instanceof Carnivore) && !animal.hidden) {
  animal.decreaseHealth(50);
}
```

witflash (approved):
Good job!

PR #1878, opened 2023-06-25
add task herbivores and carnivores

AnastasiiaKhemych:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  }
+
+  decreaseHealth(amount) {
+    const index = Animal.alive.indexOf(this.name);
+
+    if (!this.hidden) {
+      this.health -= amount;
+
+      if (this.health === 0) {
+        Animal.alive.splice(index, 1);
```
YegorKorenkov, line 30: You can filter by health instead of taking index and splice

YegorKorenkov (approved):
GJ

PR #1879, opened 2023-06-25
Solution

anya-shults:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.name = name;
+  }
+
+  bite(beast) {
+    if (beast.hidden === false && beast instanceof Carnivore === false) {
+      beast.health -= 50;
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
YegorKorenkov, line 36: You can firstly check `if (beast.health === 0)` only after that call the filter method. This is necessary in order to prevent unnecessary calls to the filter method

YegorKorenkov (approved):
GJ! Please pay attention on comment

PR #1881, opened 2023-06-26
add task solution

AlyonaSarapina:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.health = health || 100;
```
ManoylovAC, line 5: you can use value by default
```suggestion
  constructor(name, health = 100) {
    this.health = health;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden !== true && herbivore instanceof Herbivore) {
```
ManoylovAC, line 26: please use more appropriate and clear name
you name this argument as `herbivore` but it also can be `carnivore` or any else `animal` - it's a bit confusing

ManoylovAC (approved):
Hey, good work 👋
Please pay attention to the names - good names are an important part of coding

PR #1887, opened 2023-06-26
solution

ihorhladkov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.carnivore = true;
```
qoa11a, line 27: You don't need this property.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  }
+
+  bite(animalName) {
+    if (!animalName.hidden && !animalName.carnivore) {
```
qoa11a, line 31:
```suggestion
    if (!animalName.hidden && !animalName.carnivore) {
```
Here you can check if the animal is instaceOf Herbivore.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+      animalName.health -= 50;
+    }
+
+    if (animalName.health === 0) {
```
qoa11a, line 35: If the health is below zero the animal should die as well.

qoa11a (changes requested):

And678 (approved):
Approved :heavy_check_mark:

PR #1888, opened 2023-06-26
add task solution

irinabolbat:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
qoa11a, line 16: Move it to the Animal's constructor.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(bitedAnimal) {
```
qoa11a, line 30: It may not be bitten if it's a Carnivore. Not a good name for an argument.

qoa11a (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health, hidden);
+    Animal.alive.push(this);
```
qoa11a, line 27: You don't need this because now every instance of Animal will be pushed to the Animal.alive by default. But now it's being pushed twice.

qoa11a (changes requested):

qoa11a (approved):

anstsot (approved):

PR #1891, opened 2023-06-26
Solution

melnyksergey7:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+    Animal.alive.push(this);
+  }
+  static removeDead() {
+    Animal.alive = Animal.alive.reduce((acc, animal) => {
+      if (animal.health > 0) {
+        acc.push(animal);
+      }
+
+      return acc;
+    }, []);
```
ManoylovAC, line 17: you can do this much easier by using the filter method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden) {
+      return 'The bite method does not work.';
```
ManoylovAC, line 37: you no need to return any string
you can use only 'return'-statement or make invert condition and decrease here animal health

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    Animal.alive.push(this);
+  }
+  static removeDead() {
```
ManoylovAC, line 10: actually method will remove all dead animals, so it's better to use an appropriate name like `removeAllDead`

ManoylovAC (approved):
Good work
Please check my comments and improve your code

PR #1894, opened 2023-06-26
solution

AnatoliiLykhohodenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
 'use strict';
-
 class Animal {
-  // write your code here
+  static alive(animals) {
```
ManoylovAC, line 3: Maybe you mean `getAlive`, that would be more descriptive.
I don't see where you are using this method. If you don't use it, then delete it to avoid confusion.

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(bited) {
```
ManoylovAC, line 28: not very clear name
you can use something like `target` or `animal`
```suggestion
  bite(animal) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+    if (!bited.hidden && bited.hasOwnProperty('hidden')) {
+      bited.health -= 50;
+    }
+
+    if (bited.health === 0) {
```
ManoylovAC, line 33: sometimes in other scenarios it will not work
so fo sure it is better check in this way
```suggestion
    if (animal.health <= 0) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+      Animal.alive = Animal.alive.filter(animal => animal !== bited);
+    }
+  }
 }
 
+Animal.alive = [];
```
ManoylovAC, line 39: it is better to move this property right after `Animal` class or place it inside `Animal` class

ManoylovAC (changes requested):
Good start
Please check my comments I left for you and make some improvements

mgrinko (approved):

PR #1896, opened 2023-06-26
relationships between animal

ivanbilotserkivskiy:


qoa11a (changes requested):
At the moment, if I create a herbivore it'll be pushed to the `Animal.alive` array twice. Find a mistake and fix it.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+    }
+
+    if (victim instanceof Herbivore && victim.health <= 0) {
```
ManoylovAC, line 30: You already check if the victim is a `Herbivore` before you bite, so it's not necessary in this case
```suggestion
    if (victim.health <= 0) {
```

ManoylovAC (approved):
Good work 👍

PR #1897, opened 2023-06-26
Solution

eleonora-demikh:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
qoa11a, line 16: Move it to the Animal's constructor.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
```
qoa11a, line 27: Same here.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
qoa11a, line 21: The method `hide` should do what it says. At the moment, it can unhide an animal.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if ((animal instanceof Carnivore) === false
+      && animal.hidden === false) {
```
qoa11a, line 32:
```suggestion
    if (!(animal instanceof Carnivore) && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      animal.health -= 50;
+    };
+
+    Animal.alive = Animal.alive
+      .filter(an => an.health > 0);
```
qoa11a, line 37: Do not filter the array every time without a reason. You should check if the animal's health equals or less than a zero and only then you should delete an animal from the array.

qoa11a (changes requested):

tetiana-ivanchenko (approved):

PR #1899, opened 2023-06-26
Solution

perfectAcedia:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    super(name, health);
+    this.isCarnivore = true;
+  }
+
+  bite(animal) {
+    if (!animal.hidden && !animal.isCarnivore) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(an => an.health > 0);
```
ManoylovAC, line 38: you will do filter method even if all animals is alive
It is better to do this only if you need - please think about it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
+    } else {
+      Animal.alive.push(this);
+    }
```
ManoylovAC, line 12: you will check this condition every time
but you can avoid this if set static `alive` property as empty array by default

ManoylovAC (approved):
Good work good 👍 
And you can do better, please check my comments that I left for you

PR #1900, opened 2023-06-26
add solution

MarianVolytskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
ManoylovAC, line 26: what about if animal is `Carnivore`?
you should bite only `Herbivore`
image

ManoylovAC (changes requested):
Hey, please check your `bite` method
Everything else looks fine 👌

src/herbivoresAndCarnivores.js
MarianVolytskyi: Done

PR #1902, opened 2023-06-26
Solution

dianadiv:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
qoa11a, line 16: It's better to move it to the Animal's constructor.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden !== true) {
```
qoa11a, line 31:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

qoa11a (approved):
You should improve your code according to the suggestions.

meforoma (approved):
{technical approve}

PR #1903, opened 2023-06-26
Solution

FlameInBrains:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
+    }
```
ManoylovAC, line 10: you will check this condition every time, but you can set this value by default

ManoylovAC (approved):
Good work!

PR #1906, opened 2023-06-26
solution

ovynarchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
```
ManoylovAC, line 10: please leave a blank line between different code for better readability

the same applies to other parts of the code

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(object) {
+    if (object.hidden === false) {
+      object.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
ManoylovAC, line 27: pay attentions you will run this filter every time, but you can do it only when some animal loose health

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(object) {
+    if (object.hidden === false) {
```
ManoylovAC, line 24: what about if animal is `Carnivore`?
you should bite only `Herbivore`
image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+    this.hidden = hidden;
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(object) {
```
ManoylovAC, line 23: name `object` is not very clear, you can use `animal` for example
```suggestion
  bite(animal) {
```

ManoylovAC (changes requested):
Good start
Please review my comments and make some improvements
Good luck 🍀

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false && animal instanceof Herbivore) {
+      animal.health -= 50;
+      Animal.alive = Animal.alive.filter(creature => creature.health > 0);
```
ManoylovAC, line 28: pay attentions you will run this filter every time, but you can do it only when animal loose health totally
```suggestion

      if (animal.health <= 0) {
        Animal.alive = Animal.alive.filter(creature => creature.health > 0);
      }
```

ManoylovAC (approved):
Good work

PR #1909, opened 2023-06-26
added task solution

vlopatko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
qoa11a, line 25:
```suggestion
    if (!animal.hidden) {
```
You should check if the Animal is instance of Herbivore || not instance of Carnivore.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+      animal.health -= 50;
+    };
+    Animal.alive = Animal.alive.filter(el => el.health > 0);
```
qoa11a, line 28: Do not filter an array without a reason. Do it only when animal.health is less than or equal to zero.

tetiana-ivanchenko (approved):

PR #1912, opened 2023-06-26
add task solution

parmenid89:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
ManoylovAC, line 20: think about:
if `Herbivore` is already hidden, you will unmask it in this case

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
ManoylovAC, line 25: please make more appropriate and clear name
you name this argument as `herbivore` but it also can be `carnivore` or any else `animal` - it's a bit confusing

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    if (herbivore.hidden === false && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      const beastIndex = Animal.alive.findIndex(el => el.health <= 0);
+
```
ManoylovAC, line 32: you check current `herbivore` health but you can find another one with 0 health also (for example it loosed health in another way)

so you need search really current `herbivore` you check in condition or filter all animals with no health to remove them from array

ManoylovAC (changes requested):
Not bad and it's works
I left some comments for you to improve some potential cases

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false && animal instanceof Animal) {
```
ManoylovAC, line 26: you can remove health only if animal instanceof `Herbivore` but not any Animal

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+    // eslint-disable-next-line
+      const beastIndex = Animal.alive.findIndex(el => el.health <= 0 && el.name === animal.name);
```
ManoylovAC, line 32: you already check if animal health is less 0 in condition above
try this one or something like that
```suggestion
      const beastIndex = Animal.alive.findIndex(el => animal);
```

ManoylovAC (changes requested):
Please check my new suggestions

src/herbivoresAndCarnivores.js
ManoylovAC: you forgot about this condition
image

ManoylovAC (changes requested):
You missed my point about checking the victim - please fix it 
image

src/herbivoresAndCarnivores.js
```diff
@@ -23,7 +23,7 @@ class Herbivore extends Animal {
 
 class Carnivore extends Animal {
   bite(animal) {
-    if (animal.hidden === false && animal instanceof Animal) {
+    if (animal.hidden === false && animal instanceof Herbivore) {
```
parmenid89, line 26: Is it correct now? 
I checking if the animal is hiding and also if the animal is an instance of Herbivore

ManoylovAC: yes, now it's ok)

ManoylovAC (approved):
Good work

PR #1917, opened 2023-06-27
Solution

TamaraShukalo:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
ManoylovAC, line 30: pay attentions you will run this filter every time, but you can do it only when some animal loose health

ManoylovAC (approved):
Good work 👍

PR #1918, opened 2023-06-27
Solution

kylyvnyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
```
qoa11a, line 27:
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+      victim.health -= 50;
+      Animal.alive = Animal.alive.filter(live => live !== victim);
```
qoa11a, line 29: Do not filter an array without a reason. Only do it when victim.health is less than or equal to zero.

qoa11a (approved):

meforoma (approved):
{technical approve}

PR #1920, opened 2023-06-27
Solution

zoriana-melnychuk:


YegorKorenkov (approved):
GJ!

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health === 0) {
+      Animal.alive = Animal.alive.filter(animal => animal !== herbivore);
+    }
+  }
 }
 
+Animal.alive = [];
```
ManoylovAC, line 39: it will be better to move this right after `Animal` class or place it inside `Animal` class as static property

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(health = 100, name) {
+    super(name, health);
+  }
+
+  bite(herbivore) {
```
ManoylovAC, line 28: you name this argument as `herbivore` but it also can be `carnivore` or any else `animal` - it's a bit confusing
`animal` name will be more truthy

ManoylovAC (approved):
Good work!
Also, please check my comments that I have left for you for the future

PR #1921, opened 2023-06-27
js_herbivores_and_carnivores solution

NataliGru:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
-  // write your code here
+  bite(animal) {
+    const DAMAGE_OF_BITE = 50;
+    const isHerbivore = (animal instanceof Herbivore);
+    const isEnoughHealth = (animal.health >= DAMAGE_OF_BITE);
+    const isVisible = (animal.hidden === false);
+
+    if (isHerbivore && isVisible && isEnoughHealth) {
+      animal.health -= DAMAGE_OF_BITE;
+    }
```
qoa11a, line 34: The animal should be bitten even if it has not enough health. It'll die anyway, right? ;(

qoa11a (changes requested):

src/herbivoresAndCarnivores.js
NataliGru: Yeah, my bad) 
I thought it wouldn't make sense for an animal with zero health to be bitten, but I agree that Carnivore can eat the rests of an animal. 
Thank you for the correction))

qoa11a: But you were not checking if the animal has zero health. You were checking if the health is bigger than or equal to the bite damage. So if the animal has 20 health and the bite damage is 1000 the animal wouldn't die.

qoa11a (approved):

src/herbivoresAndCarnivores.js
NataliGru: Oh, indeed) I initially compared it to zero, and at the end, for some reason, I switched to comparing it with bite damage)  that's why I still thought I was comparing it to zero)

meforoma (approved):
{technical approve}

PR #1928, opened 2023-06-30
solution

dipinchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    const isHerbivore = animal instanceof Herbivore;
+    const isHidden = animal.hidden === true;
+
+    animal.health = isHerbivore
+      ? (isHidden ? animal.health : animal.health - 50) : animal.health;
```
mgrinko, line 30: NEVER format code like this (with double ?:)! 
It is impossible to read

The next one is much simpler

```suggestion
    if (isHerbivore && !animal.hidden && animal.health > 0) {
      animal.health -= 50;
    }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+    if (animal.health === 0) {
```
mgrinko, line 32: Because 30 - 50 === -20 

```suggestion
    if (animal.health <= 0) {
```

mgrinko (approved):

PR #1933, opened 2023-07-04
add alive animal

dubyuliia:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
-}
+  constructor() {
+    super();
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
```
mykhalenych, line 23: The hide method always sets the hidden property to true, which means that once a Herbivore is hidden, it can never unhide. Consider changing the implementation to toggle the hidden state, or perhaps add a separate method for unhiding.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+};
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(someone) {
+    if (someone instanceof Herbivore && !someone.hidden) {
+      someone.health -= 50;
+    }
+    Animal.updateAliveAnimals();
```
mykhalenych, line 32: You're updating the alive animals array every time a bite method is called. Consider updating the array when the health of an animal changes to 0 or less, instead of after every bite.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  };
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
-}
+  constructor() {
+    super();
```
mykhalenych, line 19:
```suggestion
    super(name);
```

mykhalenych (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
-}
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
+  }
+  showUp() {
+    this.health = false;
```
anastasiiavorobiova, line 26:
```suggestion
    this.hidden = false;
```

anastasiiavorobiova (changes requested):
Awesome solution! Added a tiny comment to fix

TarasHoliuk (approved):
LGTM :+1:

PR #1935, opened 2023-07-05
Solution

ShamievDavid:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
```
Viktor-Kost, line 4: constructor of `Animal` class does not take `health` as an argument according to task. It should be set to 100 by default.

Viktor-Kost (changes requested):
GJ!
A few minor fixes needed

etojeDenys (approved):
lgtm

PR #1936, opened 2023-07-06
Solution

vova-sherstov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
OleziO, line 27: Replace with `this.hidden = true;`

OleziO (changes requested):

PR #1938, opened 2023-07-06
Solution

1v43k:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  decrease(damage) {
+    this.health -= damage;
```
mykhalenych, line 11: The 'decrease' method is a specific behavior related to the 'Carnivore' class. You should move this method to the 'Carnivore' class.

.eslintrc.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   extends: '@mate-academy/eslint-config',
-  parser: "babel-eslint",
```
mykhalenych, line 3: why did you delete it?

mykhalenych (changes requested):

.eslintrc.js
1v43k: That wasn’t me, i didn’t even open this file, should be some mistake here, sorry(

etojeDenys (approved):
LGTM

PR #1939, opened 2023-07-06
Solution RM

RuslanMelnykFE:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore) {
```
Viktor-Kost, line 30: You should return when animal is hidden as well
```suggestion
    if (animal instanceof Carnivore || animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+      return;
+    }
+
+    if (!animal.hidden) {
+      animal.health -= 50;
+    }
```
Viktor-Kost, line 36:
```suggestion
      animal.health -= 50;
```

Viktor-Kost (changes requested):

nazarmatsevych (approved):

PR #1944, opened 2023-07-11
add solution

prostonadobobb:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
volodymyr-soltys97, line 28: The hide method should change the hidden property, not just set it to true. It should be toggled each time hide method is called.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1946, opened 2023-07-12
Solution

SerhiiVoitiuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
TarasHoliuk, line 24: since the method name is `hide` it's a bit unexpected that it toggles `hidden` value. according to the name, it should work only in one way (only set `true`)

TarasHoliuk (changes requested):
GJ!
Left one minor comment

Viktor-Kost (approved):

PR #1947, opened 2023-07-12
Solution

katiailchenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  bite(object) {
+    if (
+      Object.getPrototypeOf(object) !== Carnivore
+      && object.hidden === false
+    ) {
+      object.health -= 50;
+    }
+
+    if (object.health <= 0) {
+      Animal.alive = Animal.alive.filter((element) => element.health > 0);
```
ivadimko, line 39: This code would remove all animals with health <= 0 whenever any animal is bitten, not just the animal being bitten. Should filter out the specific animal when its health reaches 0.

consider using `Animal.alive.filter((element) => element !== object);

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(object) {
+    if (
+      Object.getPrototypeOf(object) !== Carnivore
```
ivadimko, line 32: Incorrect check for object instance type. GetPrototypeOf is not the correct way to check the instance of a class.

consider usgin `object instanceOf Carnivore`

ivadimko (approved):

PR #1948, opened 2023-07-13
Solution

Luk2asz:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    if (Animal.alive === undefined) {
+      Animal.alive = [];
+    };
```
DorotaLeniecDev, line 10: You could initiate **alive** like this
```js
static alive = [];
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
DorotaLeniecDev, line 26: Default value is set in Animal calss, no need for duplicating
```suggestion
  constructor(name, health) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    super(name, health);
+  }
+
+  bite(bitten) {
+    if (bitten.hidden === false) {
```
DorotaLeniecDev, line 31: Check if bitten is instance of **Herbivore** before decrising health. It should only work on Herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      bitten.health -= 50;
+    }
+
+    if (bitten.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(x => x === bitten), 1);
```
DorotaLeniecDev, line 36: Make use of **.filter** array method here
```suggestion
      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```

Checklist: _[CODE PERFOMANCE] - Try to avoid nesting loops (indexOf + splice), use filter instead_

DorotaLeniecDev (changes requested):
Few changes needed

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health) {
+    this.name = name;
+    this.health = 100;
```
darokrk, line 7: The `health` parameter is ignored in the Animal constructor, it always sets `this.health` to 100. If you want to set a default value if `health` is not passed, you can use `this.health = health || 100;`

darokrk (approved):
Good work 👍🏽  give approval to not block you, but fix my one mention before moving next task

PR #1949, opened 2023-07-13
Solution

ArtemVelychko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false && !(herbivore instanceof Carnivore)) {
```
```suggestion
    if (!herbivore.hidden && !(herbivore instanceof Carnivore)) {
```
Also, you can just check if it has instanceof Herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
polosanya, line 26: We don't know if it's certainly herbivore, so it's better to change the name
```suggestion
  bite(animal) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    if (herbivore.hidden === false && !(herbivore instanceof Carnivore)) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(herbivore), 1);
```
polosanya, line 32: image

polosanya (changes requested):
Almost done!

lerastarynets (approved):
You're doing amazing, keep it up!

PR #1951, opened 2023-07-14
Solution

AnnaViolin23:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(a => a.health > 0);
```
anastasiiavorobiova, line 30: The filter operation for the 'alive' array should not be inside the 'bite' method. It will be more efficient if it is done after an animal's health is decreased to 0 to maintain the 'alive' array more accurately.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
```
anastasiiavorobiova, line 9: Consider creating the health check function here for handling the array of animals

anastasiiavorobiova (changes requested):
Great solution! Added tiny comments to fix

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+}
+
+function checkHealth() {
+  Animal.alive = Animal.alive.filter(animal => animal.health > 0);
 }
```
anastasiiavorobiova, line 14: The function checkHealth() is a global function which is supposed to operate on data of Animal class. It would be better to include it as a static method inside the Animal class.

anastasiiavorobiova (changes requested):
Almost done! Added a tiny comment to fix

anastasiiavorobiova (approved):
Awesome!

PR #1952, opened 2023-07-14
add task solution

w315:
have some code duplication (this.health are set separately), I have no idea about fixing that without an error of ESLint (useless constructor in Carnivore)

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.health = 100;
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.health = 100;
+  }
+
+  bite(prey) {
+    if (prey.hidden !== undefined && prey.hidden !== true) {
```
```suggestion
    if (!prey.hidden) {
```

I guess it should work as well

nazarmatsevych (approved):
Good work, left minor comment

nazarmatsevych:
> have some code duplication (this.health are set separately), I have no idea about fixing that without an error of ESLint (useless constructor in Carnivore)

I think you can declare this.health in Animal class

w315:
> > є певне дублювання коду (this.health встановлюється окремо), я не знаю, як виправити це без помилки ESLint (некорисний конструктор у Carnivore)
> 
> Я думаю, ви можете оголосити this.health у класі Animal



> > have some code duplication (this.health are set separately), I have no idea about fixing that without an error of ESLint (useless constructor in Carnivore)
> 
> I think you can declare this.health in Animal class

I did so from the beginning, in carnivores in the constructor without an additional field (only with super) it says that the constructor is useless

src/herbivoresAndCarnivores.js
w315: then subtracted health in carnivores

PR #1954, opened 2023-07-17
Solution

AllaSerhiienko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static die(animal) {
+    if (animal.health === 0) {
```
loralevitska, line 11:
```suggestion
    if (!animal.health) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+      const index = Animal.alive.indexOf(animal);
+
+      Animal.alive.splice(index, 1);
```
loralevitska, line 14: This solution is more obvious and will allow you to filter out all the animals that already have a health level of zero or below zero. Rewrite this solution with `filter` method.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
loralevitska, line 38:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

loralevitska (changes requested):
Almost perfect) Let's improve

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static die() {
+    Animal.alive = Animal.alive.filter(elememt => elememt.health > 0);
+  }
```
sTorba24, line 12: use this as just filter in bite() function. It's not logically a little, we don't have method `die` in real life :))

sTorba24 (changes requested):
Almost good! Lets make it perfect

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
DarkMistyRoom, line 14: no need to set a default value for `health` here because you've done it in the Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
DarkMistyRoom, line 25: same

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+      Animal.alive = Animal.alive.filter(elememt => elememt.health > 0);
```
DarkMistyRoom, line 32: no need to run the `filter` method if `animal` health is > 0. It's about performance

PR #1958, opened 2023-07-20
My solution

misha-shevchuk:
All features were implemented.

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -27,8 +27,6 @@ class Herbivore extends Animal {
 
 class Carnivore extends Animal {
   bite(victim) {
-    // const victimIndex = Animal.alive.findIndex(animal => animal === victim);
-
     if (victim instanceof Herbivore && !victim.hidden) {
       victim.health -= 50;
       Animal.removeKilled();


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    const victimIndex = Animal.alive.findIndex(animal => animal === victim);
+
+    if (victim.hasOwnProperty('hidden') && victim.hidden === false) {
+      victim.health -= 50;
```
anastasiiavorobiova, line 33: You decrease the victim's health, but there is no check if the victim's health is already 0 or less. It would be better to add a condition to check if the victim's health is more than 0 before decreasing it.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    } else {
+      this.hidden = true;
+    }
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    const victimIndex = Animal.alive.findIndex(animal => animal === victim);
```
anastasiiavorobiova, line 30: You're finding the index of the victim but it could be more efficient to remove the animal from the alive array within the Animal class itself when its health reaches 0. This way you avoid searching for the animal's index every time a bite occurs.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+    if (victim.hasOwnProperty('hidden') && victim.hidden === false) {
```
anastasiiavorobiova, line 32: You are checking if the victim is a Herbivore by checking the 'hidden' property. However, it would be better to check the instance of the object to ensure the victim is a Herbivore. You can use 'instanceof' operator to check if the victim is an instance of Herbivore class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    if (this.hidden) {
```
anastasiiavorobiova, line 20: The 'hide' method can be simplified by using the 'not' operator to toggle the boolean value. So instead of an if-else statement, you can write 'this.hidden = !this.hidden'.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
-  // write your code here
+  bite(victim) {
+    const victimIndex = Animal.alive.findIndex(animal => animal === victim);
+
+    if (victim.hasOwnProperty('hidden') && victim.hidden === false) {
+      victim.health -= 50;
+    }
+
+    if (victim.health <= 0) {
+      Animal.alive.splice(victimIndex, 1);
```
anastasiiavorobiova, line 37: Consider creating a function in the Animal class for this purpose

anastasiiavorobiova (changes requested):
Great start! Added comments to fix

anastasiiavorobiova (approved):
Awesome!

PR #1959, opened 2023-07-20
solution

tndeineka:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health === 0) {
+      const index = Animal.alive.indexOf(herbivore);
+
+      Animal.alive.splice(index, 1);
```
loralevitska, line 32: Better use `filter` method here

loralevitska (changes requested):
Almost done) Let's improve

VitaliyBodnar (approved):
Good job!

PR #1960, opened 2023-07-20
js_herbivores_and_carnivores

KarolinaWiktoria:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
DorotaLeniecDev, line 24: According to the checklist: _if the function has a name hide this function should just hide but not toggle_
```suggestion
    this.hidden = true;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+      Animal.removeKilled();
```
DorotaLeniecDev, line 32: In my opinion it would be good to check here `victim.health` and if it is `<=0` then you could call the `Animal.removeKilled` function. There is no need to filter Animals when there is no need to. Imagine if we will have some more amout of data that might cost a lot of performance.

- [ ] Conditionaly call `Animal.removeKilled()`

DorotaLeniecDev (changes requested):
Check my comment for performance optimalization

src/herbivoresAndCarnivores.js
KarolinaWiktoria: Sure, I've forgotten this point of the checklist.

src/herbivoresAndCarnivores.js
KarolinaWiktoria: okay thanks. Good suggestion :)!

Radoslaw-Czerniawski (approved):
All good :)

PR #1961, opened 2023-07-22
Solution

Illia-L:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive.splice(Animal.alive.findIndex(a => a === animal), 1);
```
Anastasiia-Svintsova, line 31: Using Array.prototype.splice() with Array.prototype.findIndex() to remove an item from an array can be inefficient for large arrays. Consider using Array.prototype.filter() instead.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
Anastasiia-Svintsova, line 27:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

Anastasiia-Svintsova (changes requested):

nazarmatsevych (approved):

PR #1964, opened 2023-07-23
completed task

Horizon-git:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,39 @@
...
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive = Animal.alive.filter((item) => item.health > 0);
```
volodymyr-soltys97, line 34: The code is checking and removing dead animals only when a Carnivore bites a Herbivore. This could potentially leave dead animals in the 'alive' array if the animal's health reaches zero by some other means.

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,39 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
   // write your code here
+  constructor(name, health = 100, hidden = false) {
```
volodymyr-soltys97, line 17: No need to pass 'hidden' as a parameter to the Herbivore constructor, it should always be initialized to false.

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,39 @@
...
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
```
volodymyr-soltys97, line 33: Health cannot have a negative meaning

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

TarasHoliuk (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,36 @@
...
 class Herbivore extends Animal {
   // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
Viktor-Kost, line 24: this method should only hide
```suggestion
  hide() {
    this.hidden = true;
  }
```

Viktor-Kost (approved):
Nice!
Check a minor fix to update your code

PR #1965, opened 2023-07-24
Solution

Yuliia-Hrinchenko1:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
```
vlad-demchuk, line 14: you can use filter method here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+    }
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super();
```
vlad-demchuk, line 23: specify the properties you should inherit and then use them instead of super.die();

vlad-demchuk (changes requested):

vlad-demchuk (changes requested):
test

volodymyr-soltys97 (approved):
Good job 👍

PR #1967, opened 2023-07-25
task solution

AnVBondar:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,57 @@
...
+    if (target instanceof Herbivore && target.health === 0) {
+      return;
+    }
+
+    if (target instanceof Herbivore && target.hidden !== true) {
+      target.health -= 50;
+      findTarget.health -= 50;
+
+      if (findTarget.health === 0) {
+        Animal.alive.splice(targetIndex, 1);
```
vlad-demchuk, line 49: it will be better to use filter method

vlad-demchuk (approved):

src/herbivoresAndCarnivores.js
AnVBondar: @vlad-demchuk , Thank you for advise. Fixed.

PR #1976, opened 2023-08-02
Add tusk solution.

MaksymBevskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
anastasiiavorobiova, line 26: Consider checking the instance of the animal before biting it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+      herbivore.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(pets => pets.health > 0);
```
anastasiiavorobiova, line 30: Consider adding this functionality to the animal class instead of filtering in the Carnivore class

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #1983, opened 2023-08-04
herbivores And Carnivores

NikolaienkoN:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-}
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
+
+    Animal.alive.push(this);
+  }
+  die() {
+    if (this.health === 0) {
+      const index = Animal.alive.indexOf(this);
```
etojeDenys, line 12: let's just use [filter](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) here

etojeDenys (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
-  // write your code here
-}
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
 
+    Animal.alive.push(this);
+  }
+  die() {
+    Animal.alive = Animal.alive.filter(element => element.health !== 0);
```
SanyaBratashchuk, line 11:
```suggestion
    Animal.alive = Animal.alive.filter(element => element.health >= 0);
```

SanyaBratashchuk (approved):

PR #1990, opened 2023-08-08
solution

Sancz0pansa:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden === false) {
+      target.health -= 50;
+
+      if (target.health <= 0) {
+        Animal.alive = Animal.alive.filter(animal => animal !== target);
```
JoyCoffeeAddict, line 29: You can move this line into `Animal` class and make it a static method, just for practice.

JoyCoffeeAddict (approved):
Approve with a suggestion.

PR #1996, opened 2023-08-09
Solution

deandre25:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
denys-danyliuk, line 4:
```suggestion
  constructor(name) {
```
it should take only name as param
Знімок екрана 2023-08-10 о 10 39 30

denys-danyliuk (approved):

PR #1997, opened 2023-08-10
task solution

HubertKlamann:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
```
choeqq, line 9: please cosider reducing health and dying in Animal class - every Animal can have it's health reduced or die

choeqq (approved):

PR #1998, opened 2023-08-10
Solution

klaudia-klis:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
choeqq, line 26:
```suggestion
    if (animal instanceof Herbivore && animal.hidden {
```
when it comes to boolean values you can just check for that value - you don't have to check if it equas something

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+    this.health = health || 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
```
choeqq, line 14:
```suggestion
  constructor(name, health) {
```
I think hidden is not necessary here, that belongs only to Herbivore class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+    super(name, health);
+    this.hidden = false;
```
choeqq, line 16: also you aren't even using hidden argument 😄

choeqq (changes requested):

choeqq (approved):

PR #1999, opened 2023-08-10
Animal hunter

WaveOfDandelions:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    if (!obj.hidden && obj instanceof Herbivore) {
+      obj.health -= 50;
+    }
+
+    Animal.alive.map((animal, index) => {
+      if(animal.health <= 0) {
+          Animal.alive.splice(index, 1);
+        }
+      return animal;
+    })
```
choeqq, line 41: filter method is perfect for that

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.name = name;
+  }
+
+  bite(obj) {
```
choeqq, line 31: I think you can do little better on that obj name

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
```
choeqq, line 9: please consider adding reduceHealth and die methods to Animal class, then you could just call them when biten

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.name = name;
+    this.hidden = false;
+  }
```
choeqq, line 18: Herbivore should also have health
```suggestion
  constructor(name, health) {
    super(name, health);
    this.name = name;
    this.hidden = false;
  }
```

choeqq (changes requested):

asakevych (approved):
Great job!! 🚀 🔥 👏 

you can also remove comments `// write your code here` to make file look better, there is already your code 😉

PR #2000, opened 2023-08-10
solution

cardinal1991:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,43 @@
...
+    this.name = name;
+  };
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
   // write your code here
+  constructor(name, hidden = false) {
+    super(100, name);
```
choeqq, line 16: please make Herbivore constructor take name and health, instead of hardcoding 100, also you don't have to take/pass hidden property
```suggestion
    super(100, name);
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(health, name) {
+    super(100, name);
+  }
+
+  bite(herbivoreobj) {
```
asakevych, line 31: please choose better names -> in example you can see that your argument can be also carnivore, so probably it should be `bite(animal) {` ?

PR #2001, opened 2023-08-10
Develop

anpiet:


choeqq (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 class Animal {
-  // write your code here
+  static alive = [];
+
+  health = 100;
+
+  constructor(name) {
+    this.name = name;
+    Animal.alive.push(this);
+  }
```
asakevych, line 11: it task description there is:
All instances of the `Animal` class must have properties health and a name. Health is 100 by default.

asakevych (changes requested):
Good job!! small changes left 😉

src/herbivoresAndCarnivores.js
anpiet: Good point. I have changed this.
I just don't understand why should I add reduceHealth and die methods to Animal if it's only needed in herbivore. I have added this, but I think that now code is more complex. Could you explain it to me a little bit more?

anpiet: I know that I could add Animal.live outside the class, but the task was:
All alive animals should be in the static `Animal.alive` array. I think that word 'static' says that it should be added inside the class.

choeqq (approved):
Looks good to me 🚀

PR #2005, opened 2023-08-10
add task solution

kuzgan:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
```
choeqq, line 27: you can do little better on that name 
```suggestion
  bite(betterVariableName) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    if (prey instanceof Carnivore || prey.hidden === true) {
```
choeqq, line 28: you don't have to check if boolean value are true/false
```suggestion
    if (prey instanceof Carnivore || prey.hidden) {
```

choeqq (changes requested):

choeqq (approved):
Looks good to me 🚀

PR #2006, opened 2023-08-10
add task solution

ogospadzipalina:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,56 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    Animal.alive = Animal.alive.filter(animal => animal !== this);
+  }
+
+  afterBite(decreaseInHealth) {
```
choeqq, line 15:
```suggestion
  reduceHeatlh(decreaseInHealth) {
```
I think that's a little better name for that method

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,56 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
choeqq, line 31: that's already done when Animal constructor is exectued

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,56 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    Animal.alive.push(this);
```
choeqq, line 42: that's already done when Animal constructor is exectued

PR #2007, opened 2023-08-10
init commit

domalewski:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false && herbivore instanceof Herbivore) {
```
choeqq, line 26: you don't have to use equals to check for boolean values

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    // is better to use !herbivore.hidden && !(herbivore instanceof Carnivore)?
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      Animal.alive = Animal.alive.filter(animal => animal !== herbivore);
+    }
```
choeqq, line 33: this belong to every Animal, so it should be on animal class

choeqq (changes requested):

choeqq (changes requested):
Waiting for any updates @domalewski 👀

PR #2009, opened 2023-08-11
Solution

Reject-4444:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    if (target.health <= 0) {
+      const index = Animal.alive.indexOf(target);
```
mykhalenych, line 29: Potential issue with array manipulation. If the target is not found (index == -1), splice will remove the last item in the array, which is incorrect. Better to use filter

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+    }
+
+    if (target.health <= 0) {
```
mykhalenych, line 28: Checking for health <= 0 should be part of the bite method code block to prevent an exception if the target is not a Herbivore.

mykhalenych (changes requested):

Andriy-Shymkiv (approved):

PR #2010, opened 2023-08-11
solution

Nightycake:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,53 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
asakevych, line 32: you can add it to your Animal constructor so it wouldn't be necessary to repeat in every class that inherits Animal

asakevych (changes requested):
Great job!! I really liked your solution 😍 🚀,  just 1 small change is needed 😉

choeqq (approved):
Looks good to me 🚀

PR #2012, opened 2023-08-11
add task solution

jakubMikolajczykTech:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
asakevych, line 6: it task description there is:
All instances of the Animal class must have properties health and a name. Health is 100 by default.

asakevych (changes requested):
Good job!! just small changes left 😉

choeqq (changes requested):
Waiting for any updates @jakubMikolajczykTech 👀

PR #2013, opened 2023-08-11
add task solution

arturwisnia:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
asakevych, line 6: it task description there is:
All instances of the Animal class must have properties health and a name. Health is 100 by default.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    Animal.alive.push(this);
+  }
+
+  static removeFromAlive(animal) {
+    const index = Animal.alive.indexOf(animal);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
+  }
```
asakevych, line 17: It's a perfect case to use filter 😉

asakevych (changes requested):
Good job! just some small changes left 😉 remember about reading checklist

PR #2014, opened 2023-08-11
Solution

Vira-V:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+    this.hidden = true;
+  }
 }
-
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+      Animal.removeDeadAnimals();
```
asakevych, line 28: there is no sense to do filter every time when you decrease health, you can invoke it only when target.health is less or equal 0

asakevych (changes requested):
Great job!! 🚀  just one small change left 😉

choeqq (approved):
Looks good to me 🚀

PR #2015, opened 2023-08-12
add task solution

voievodik:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden !== true) {
+      target.health -= 50;
+
+      if (target.health === 0) {
+        const index = Animal.alive.indexOf(target);
+
+        Animal.alive.splice(index, 1);
```
polosanya, line 32: Better filter the animals and leave only animals with positive health
image

polosanya (changes requested):
Almost done! Pay attention to the checklist

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.health = 100;
```
BudnikOleksii, line 5: Looks like a magic number, the same as 50

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden !== true) {
+      target.health -= 50;
+
+      if (target.health === 0) {
+        Animal.alive = Animal.alive.filter(animal => animal !== target);
+      }
+    }
```
BudnikOleksii, line 32: Try to avoid nested "if" statements.
Also, `target.hidden !== true` you simplify `!target.hidden`

BudnikOleksii (changes requested):
Good job!
Let's get rid of magic numbers.

dffUqp (approved):
GJ✔️

PR #2017, opened 2023-08-13
solution

SviatoslavVasyletskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,57 @@
...
+    }
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
+  }
```
choeqq, line 25: that's the perfect usecase for filter method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,57 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  // eslint-disable-next-line no-useless-constructor
+  constructor(name) {
+    super(name);
+  }
```
choeqq, line 34: you probably don't even have to call the constructor here 🤔

choeqq (changes requested):

choeqq (approved):

PR #2018, opened 2023-08-14
add task solution

MateuszSlezak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false && herbivore instanceof Herbivore) {
```
choeqq, line 27: if value you're checking against is boolean you don't have to compare it to anything, it's just true/false by itself
```suggestion
    if (herbivore.hidden && herbivore instanceof Herbivore) {
```

choeqq (approved):

PR #2022, opened 2023-08-14
Solution

Vitalii-Fediaiev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
 }
 
+  Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
```
lerastarynets, line 16: The `hidden` property should not be in the constructor parameters, as it is not meant to be set upon creation of an object. It should only be changed through the `hide` method.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health = animal.health - 50;
+    }
+
+    if (animal.health <= 0) {
```
lerastarynets, line 32: The check for animal health and removal of dead animals from the alive array should not be inside the `bite` method. It should be in the Animal class and get called whenever an animal's health changes.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
```
lerastarynets, line 4: where is `health` in Animal constructor?

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit more

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+
+  bite(beast) {
+    if (beast.hidden === false && !(beast instanceof Carnivore)) {
+      beast.health = beast.health - 50;
```
anastasiia-tilikina, line 29: you can use shorthand here
```suggestion
      beast.health -= 50;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    }
+
+    if (beast.health === 0) {
+      Animal.alive = Animal.alive.filter(v => v !== beast);
```
anastasiia-tilikina, line 33: here you should delete beasts whose health is equal 0
that means you should change your condition for the filter method

anastasiia-tilikina (changes requested):
let's check your solution again and fix some issues)

also pay attention to[this](https://github.com/mate-academy/js_herbivores_and_carnivores/pull/2022#discussion_r1293531227) comment from previous review. looks like it is still not fixed)

agniya-i (approved):

PR #2025, opened 2023-08-15
add task solution

Wesses:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    if (!herbivore.hidden && !(herbivore instanceof Carnivore)) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      const index = Animal.alive.findIndex(animal => animal === herbivore);
+
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
+      }
```
Moroz-Dmytro, line 36: Use `filter` instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
Moroz-Dmytro, line 26: Try to make up a better name, because it can be another `сarnivore`, not `herbivore`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    if (!herbivore.hidden && !(herbivore instanceof Carnivore)) {
```
Moroz-Dmytro, line 27:
```suggestion
    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```

Moroz-Dmytro (changes requested):
Good job 👍 
But let's improve it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter(el => el.health > 0);
```
Moroz-Dmytro, line 30: But even better would be do this filtering `if(animal.health <= 0)`. To avoid useless filtering

Moroz-Dmytro (approved):
LGTM 👍 
Only small comment

PR #2028, opened 2023-08-15
add task solution

Koliras:


Moroz-Dmytro (approved):
👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
```
Moroz-Dmytro, line 16: But `health` is useless here, parent constructor do nothing with it

PR #2029, opened 2023-08-16
solution

patryk177m:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-}
 
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+}
+Animal.alive = [];
```
mvjl000, line 12: I think, you should first check whether this array exists and then push `this`

mvjl000 (changes requested):
Just one small change ;)

src/herbivoresAndCarnivores.js
patryk177m: That's because lint doesn't pass tests. This is a lint workaround.

choeqq (approved):

PR #2030, opened 2023-08-16
Develop

alina-kabanets:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
+    Animal.alive.push(this);
+  }
+
+  dead() {
+    const index = Animal.alive.indexOf(this);
```
lerastarynets, line 11: There is no check to see if the animal instance is actually in the 'alive' array before attempting to remove it. It would be safer to only proceed with the 'splice' operation if 'index' is not -1.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide(hidden) {
```
lerastarynets, line 25: The 'hide' method should not take any arguments. It should just set the 'hidden' property of the instance to true.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
```
lerastarynets, line 6:
```suggestion
  constructor(name, health=100) {
    this.name = name;
    this.health = health;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    super(name);
+    this.hidden = false;
+  }
+
+  hide(hidden) {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
```
lerastarynets, line 30: where is constructor in Carnivore class?

lerastarynets (changes requested):

src/herbivoresAndCarnivores.js
alina-kabanets: constructor(name) {
    super(name);
  }

Linter says it's 'useless constructor'. What shoud I add here?

Anastasiia-Svintsova (approved):

PR #2032, opened 2023-08-16
Solved task

AndriiKuz:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+const MAX_HEALTH = 100;
+
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = MAX_HEALTH) {
+    this.health = health;
+    this.name = name;
+    this.hidden = false;
```
Moroz-Dmytro, line 11: This field should be available only in class `Herbivore`

Moroz-Dmytro (changes requested):
Good progress 👍 
Some comments below
Also something is wrong with tests

Moroz-Dmytro (approved):
LGTM 🔥

PR #2040, opened 2023-08-17
the best solution, maybe

AndrMar1939:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
 'use strict';
 
+const aliveAnimalsList = [];
```
Moroz-Dmytro, line 3: Better to create this array as static field - `Animal.alive = [];`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
Moroz-Dmytro, line 27:
```suggestion
    this.hidden = true;
```

Moroz-Dmytro: I guess, we should not show the animal if we call method `hide` twice. To show animal we would need to create another method or rename this one

Moroz-Dmytro (changes requested):
Good work 👍 
But some comments

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+    this.health = HEALTH;
+    Animal.aliveAnimalsList.push(this);
+  }
+
+  static get alive() {
+    return Animal.aliveAnimalsList.filter(animal => animal.health > 0);
+  }
 }
 
+Animal.aliveAnimalsList = [];
```
Moroz-Dmytro, line 18:
```suggestion
Animal.alive = [];
```

Moroz-Dmytro: image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = HEALTH;
+    Animal.aliveAnimalsList.push(this);
+  }
+
+  static get alive() {
+    return Animal.aliveAnimalsList.filter(animal => animal.health > 0);
+  }
```
Moroz-Dmytro, line 15: You can change this to method `die` which will filter static `alive` array. And call this method on `victim` in method `bite` of `Carnivore`

Moroz-Dmytro (changes requested):
Some comments

src/herbivoresAndCarnivores.js
AndrMar1939: What a silly mistake( sorry

Moroz-Dmytro (approved):
LGTM 👍

PR #2046, opened 2023-08-17
Solution

OlhaMomot:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = (!this.hidden);
```
Moroz-Dmytro, line 24:
```suggestion
    this.hidden = true;
```

Moroz-Dmytro: I guess, method `hide` should now show animal when we call this method twice

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
```
Moroz-Dmytro, line 30:
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+      victim.health -= HEALTH_TAKEN_BY_BITE;
+    }
+
+    if (victim.health === DEAD_ANIMAL_HEALTH) {
+      const index = Animal.alive.indexOf(victim);
+
+      delete Animal.alive[index];
```
Moroz-Dmytro, line 37: Better to use `filter`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= HEALTH_TAKEN_BY_BITE;
+    }
+
+    if (victim.health === DEAD_ANIMAL_HEALTH) {
```
Moroz-Dmytro, line 34:
```suggestion
    if (victim.health <= DEAD_ANIMAL_HEALTH) {
```

Moroz-Dmytro (changes requested):
Good work 🚀 
But let's improve your solution

tiserett (approved):

PR #2047, opened 2023-08-17
add task solution

NazarVotkanych:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
```
Moroz-Dmytro, line 7: This property should be available only for `Herbivore`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    Animal.alive.push(this);
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
Moroz-Dmytro, line 13: This method should be available only for `Herbivore`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
```
Moroz-Dmytro, line 18: Why do we need argument `health` if we don't use it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    super(name);
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
```
Moroz-Dmytro, line 24: Same here

Moroz-Dmytro (changes requested):
Good progress 👍 
But let's fix this a bit

src/herbivoresAndCarnivores.js
NazarVotkanych: Sure.
I understand.
I have removed the constructor from Carnivore, and he still inherits the constructor from its parent class.

src/herbivoresAndCarnivores.js
NazarVotkanych: Ok.
I passed health to super.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
Moroz-Dmytro, line 4: Also would be nice to create constants for "magic values"

Moroz-Dmytro, line 4: You should get only `name`, `health` is useless here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
```
Moroz-Dmytro, line 13: You should get only `name`. `health` and `hidden` are useless here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
```
Moroz-Dmytro, line 26: Magic value

Moroz-Dmytro (changes requested):
GJ 👍 
Fix small comments and its perfect

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.health = FULL_HEALTH;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
```
Moroz-Dmytro, line 18:
```suggestion
  constructor(name) {
    super(name);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
+const FULL_HEALTH = 100;
+const INJURED_HEALTH = 50;
```
Moroz-Dmytro, line 4:
```suggestion
const BITE_DAMAGE = 50;
```

Moroz-Dmytro (approved):
GJ 👍 
But check comments

PR #2051, opened 2023-08-18
Solution

dmytro-popovych:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.health = health || 100;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
etojeDenys, line 15: let's just use filter method

loralevitska (approved):
Well done)

PR #2062, opened 2023-08-20
Solution

AntonHerasymov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
Moroz-Dmytro, line 27: But it also can be a `Carnivore`. Try to make up better name

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,41 @@
...
+    if (!(herbivore instanceof Carnivore) && !herbivore.hidden) {
+      herbivore.health -= BITE_DAMAGE;
+    }
+
+    Animal.alive = Animal.alive.filter((animal) =>
+      animal.health > HEALTH_AT_DEATH);
```
Moroz-Dmytro, line 33: You don't need to do useless filtering if `animal.health > HEALTH_AT_DEATH`

Moroz-Dmytro: You can add 1 more condition

Moroz-Dmytro (approved):
LGTM 🔥

PR #2065, opened 2023-08-20
add task solution

irahos:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
Moroz-Dmytro, line 25: But it also can be a `Carnivore`. Try to make up better name

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
```

Moroz-Dmytro (changes requested):
Good progress 👍 
But some comments below

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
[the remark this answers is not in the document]
irahos: sorry for reviewing your comments so late( I don't really understand why is that useless, I used this filtering to sort out the dead animals. should I do it completely differently or should I just change the filtering condition?

PR #2066, opened 2023-08-20
js_herbivores_and_carnivores

vatatan:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
Moroz-Dmytro, line 4: Magic value `100`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static alive = [];
```
Moroz-Dmytro, line 10: For better readability place it at the beginning of the class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  die() {
+    Animal.alive = Animal.alive.filter(animal => animal !== this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
```
Moroz-Dmytro, line 20:
```suggestion
    super(name);
    
    this.hidden = false;
```

src/herbivoresAndCarnivores.js
Moroz-Dmytro: But argument `health` is useless. We don't need it here

Moroz-Dmytro (changes requested):
GJ 👍 
Check comments below

PR #2068, opened 2023-08-20
solution herbivores and carnivores

ihorid:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
+const FULL_HEALTH = 100;
+const HALF_HEALTH = 50;
```
Moroz-Dmytro, line 4:
```suggestion
const BITE_DAMAGE = 50;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if ((animal instanceof Herbivore) && (animal.hidden === false)) {
```
Moroz-Dmytro, line 31:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

Moroz-Dmytro: It can be simplified

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+      animal.health -= HALF_HEALTH;
+
+      Animal.alive = Animal.alive.filter(({ health }) => health > DEAD);
```
Moroz-Dmytro, line 34: You don't need to do useless filtering if `animal.health > DEAD`. You can add 1 more condition for this

Moroz-Dmytro (changes requested):
Good work 👍 
But check comments below

dffUqp (approved):
LGTM🚀

PR #2069, opened 2023-08-20
add task solution

dpidlutska:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
Moroz-Dmytro, line 8: Magic value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
```
Moroz-Dmytro, line 14: `health` and `hidden` are useless arguments, you don't really need them

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+      if (victim.health === 0) {
```
Moroz-Dmytro, line 29:
```suggestion
      if (victim.health <= 0) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.name = name;
+    this.health = INIT_ANIMAL_HEALTH;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
```
dffUqp, line 18: looks like we don't pass name to the parent constructor.

```js
const newAnimal = new Herbivore("test")

console.log(newAnimal) // { name: undefined, health: 100, hidden: false }
```

dffUqp (changes requested):
Almost done:smiley:

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
```
roman-mirzoian, line 29: The first part of the condition can be transferred to a separate variable.

roman-mirzoian (approved):
Well done 👍

PR #2071, opened 2023-08-20
Solution

Vitalii-Petruk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivoreVictim) {
```
Moroz-Dmytro, line 25: But it also can be a `Carnivore`. Better to name it just `victim`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    if (herbivoreVictim instanceof Herbivore && !herbivoreVictim.hidden) {
+      herbivoreVictim.health -= 50;
```

Moroz-Dmytro (changes requested):
Good work ⭐ 
Some comments

BudnikOleksii (approved):
Nicely done!

PR #2077, opened 2023-08-21
Solution

valet2102:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
roman-mirzoian, line 26: If we call `hide()` twice in a row, `this.hidden` will be `fаlse`, and should always be `true` after calling.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden) {
+      return;
+    }
+
+    animal.health -= CARNIVORE_DAMAGE;
+
+    if (animal.health === 0) {
+      Animal.alive = Animal.alive.filter(({ name }) => name !== animal.name);
+    }
```
roman-mirzoian, line 39:
```suggestion
    if (animal.health <= 0) {
      Animal.alive = Animal.alive.filter(({ health }) => health > 0);
    }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden) {
```
roman-mirzoian, line 31: The first part of the condition can be transferred to a separate variable.

roman-mirzoian (changes requested):

roman-mirzoian (approved):
Well done 👍

PR #2078, opened 2023-08-21
Solution

Yevheniia-Sid:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  health = 100;
+  constructor(name) {
```
roman-mirzoian, line 6: Description from task:
`All instances of the Animal class must have properties health and a name. Health is 100 by default.`
So you should have `health` in `constructor`. Health default value can be separated to constant.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
+  hide() {
```
roman-mirzoian, line 17:
```suggestion

  hide() {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.hidden = true;
+  }
 }
 
+const BITE_STRENGTH = 50;
```
roman-mirzoian, line 22: It is better to get constants at the beginning of the code.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if ((animal instanceof Herbivore) && (animal.hidden === false)) {
```
roman-mirzoian, line 31: The first part of the condition can be transferred to a separate variable, 
and the second part is a little easier to write: `!animal.hidden`.

roman-mirzoian (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+
+  bite(animal) {
+    const isHerbivore = animal instanceof Herbivore;
+
+    if (isHerbivore && !animal.hidden) {
+      animal.health -= BITE_STRENGTH;
+
+      if (animal.health <= 0) {
+        Animal.alive = Animal.alive.filter(({ health }) => health > 0);
+      }
```
roman-mirzoian, line 40: This `if` could not be nested inside another.

roman-mirzoian (approved):
Well done 👍

PR #2079, opened 2023-08-21
add task solution

ivannaromanovych:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
+
+      if (prey.health === 0) {
+        const index = Animal.alive.indexOf(prey);
+
+        if (index !== -1) {
+          Animal.alive.splice(index, 1);
+        }
+      }
```
roman-mirzoian, line 34: Large nesting should be avoided.
Here you can get away with two `if` for the entire `bite` method.

roman-mirzoian (changes requested):

PR #2081, opened 2023-08-21
solution

Taras2705:


src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,42 @@
 'use strict';
 
+const FULL_HEALTH = 100;
+const ANIMAL_DIED = 0;
```
roman-mirzoian, line 4:
```suggestion
const DIED_VALUE = 0;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,42 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= BITE_DEMAGE;
+    }
+
+    if (animal.health === ANIMAL_DIED) {
+      Animal.alive = Animal.alive.filter((beast) => beast.health > ANIMAL_DIED);
```
roman-mirzoian, line 35: Use destructuring on the object.

src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,42 @@
...
+    }
+  }
 }
+Animal.alive = [];
```
roman-mirzoian, line 39:
```suggestion

Animal.alive = [];
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!animal.hidden && animal instanceof Herbivore) {
+      animal.health -= BITE_DEMAGE;
+    }
+
+    if (animal.health === ANIMAL_DIED) {
```
roman-mirzoian, line 34: If `animal.health` is less than 0, the condition will not be fulfilled, but it should be.

roman-mirzoian (changes requested):

PR #2089, opened 2023-08-22
Solution

stunnerboyone:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,51 @@
...
+    if (herbivore.health === 0) {
+      Animal.removeByName(herbivore.name);
+    }
+  }
 }
 
 class Herbivore extends Animal {
   // write your code here
+
+  constructor(name, health = 100, hidden = false) {
```
denys-danyliuk, line 27:
```suggestion
  constructor(name) {
```
it should take only name to create an instance of a class

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,51 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  constructor(name, health = 100) {
```
denys-danyliuk, line 39: same here

PR #2090, opened 2023-08-23
herbivoresAndCarnivores

Dorosh90:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= DAMAGE;
+    }
+
+    if (prey.health === DIED) {
```
Anastasiia-Svintsova, line 34: The current implementation will only remove the animal from the 'alive' array when its health is exactly 0. If the health drops below 0, the animal will not be removed.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+      const diedAnimal = Animal.alive.indexOf(prey);
+
+      Animal.alive.splice(diedAnimal, 1);
```
Anastasiia-Svintsova, line 37: Use the filter method instead

witflash (approved):
Good job! 👍🏻

PR #2091, opened 2023-08-23
solution

olegpaunov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
-}
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
```
alexander-ignatow, line 7: Per task:

> All instances of the Animal class must have properties health and a name. Health is 100 by default. 
> Create a Herbivore class. Herbivore has a method of hide, which changes the hidden property of the beast and helps to hide from carnivores.

Probably some of these properties should be in another class

alexander-ignatow (changes requested):
Almost done!

Lets make one small change

src/herbivoresAndCarnivores.js
olegpaunov: yes sure, no problem

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+  static removeDeadAnimals() {
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
```
roman-mirzoian, line 22:
```suggestion

    this.hidden = false;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
roman-mirzoian, line 4: It would be good to put the "magic values" in separate constants.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
```
roman-mirzoian, line 37: 50 is also "magic value".

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    this.health = health;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
+    }
+    Animal.alive.push(this);
+  }
+
+  static removeDeadAnimals() {
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
roman-mirzoian, line 15: Use destructuring on the object.

roman-mirzoian (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,43 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    const VICTIM = victim instanceof Carnivore;
```
roman-mirzoian, line 27:
```suggestion
  bite(aim) {
    const isVictim = aim instanceof Carnivore;
```

roman-mirzoian (approved):
Well done 👍

PR #2096, opened 2023-08-25
Solution

YuraMisura:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivoreAnimal) {
+    if (herbivoreAnimal.hidden === false) {
+      herbivoreAnimal.health -= 50;
+    }
+
+    if (herbivoreAnimal.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(herbivoreAnimal, 0), 1);
```
IvanFesenko, line 31: use `Array.filter()` instead

IvanFesenko (changes requested):
Good job. One improvement required

etojeDenys (approved):
lgtm

PR #2099, opened 2023-08-28
solution

mykyta01:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
+    Animal.alive.push(this);
+
+    Animal.alive.push({
+      name: this.name,
+      health: this.health,
+      hidden: this.hidden,
+    });
```
loralevitska, line 16: For what do push twice?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+  }
+
+  static removeFromAlive(animal) {
+    const index = Animal.alive.indexOf(animal);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
loralevitska, line 24: Use `filter` array method

loralevitska (changes requested):
Almost done) Let's improve

mykyta01:
fixed:) thanks for review)

anastasiiavorobiova (approved):
Great work!

PR #2105, opened 2023-08-30
add task solution

MishaLehotskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
TarasHoliuk, line 21: There is no need to do it here
Animal will be pushed when you do `super` call

The same below

TarasHoliuk (changes requested):

OleziO (approved):

PR #2112, opened 2023-09-07
Add task solution

spojrzenie:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore) {
```
darokrk, line 22: The 'bite' method should not decrease the health of the herbivore if it is hiding. The current code will decrease the health regardless of whether the herbivore is hiding or not. Use a proper condition to prevent this from happening.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+    this.hidden = false;
```
darokrk, line 9: The 'hidden' property should not be part of every Animal instance, because it is specific to Herbivores only. Move this property to the Herbivore class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore) {
+      herb.health = herb.hidden === false ? herb.health - 50 : herb.health;
+    }
+
+    if (herb.health <= 0) {
```
darokrk, line 26: The code for removing the dead herbivores should be inside the Animal class. When the health of any animal reaches 0, it should be removed from the alive array. You can create a separate method for this or include this logic in the setter of the health property. 

Try simplify this whole Carnivore class code to be just:

```
 bite(target) {
    if (target.hidden === false) {
      target.health -= 50;
    }

  };
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
```
darokrk, line 14: You should use also use a constructor method to inherit values, e.g:

```
constructor(name, health) {
    super(name, health);

    this.hidden = false;
  }
```

darokrk (changes requested):
Looks good, left few comments for improvements 👨🏽‍💻 🛠️

src/herbivoresAndCarnivores.js
spojrzenie: Thanks for comments. Please check code if is ok. Thanks.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
```
OleziO, line 40: Add constructor method.

OleziO (changes requested):

src/herbivoresAndCarnivores.js
spojrzenie: Where and why? I don't need constructor method at Carnivore class.

darokrk (approved):
Looks good, there is no need to add constructor to Carnivore 👍🏽  approved 🏆

PR #2118, opened 2023-09-11
Solution

Q1W2E3R4T5Y6U7I8a:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        const index = Animal.alive.indexOf(animal);
+
+        if (index > -1) {
+          Animal.alive.splice(index, 1);
+        }
+      }
```
nazarmatsevych, line 34: Try not to manipulate with array in this way, use filter method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
nazarmatsevych, line 19:
```suggestion
    this.hidden = true;
```

hide method should only change hidden value to true and not toggle it

nazarmatsevych (changes requested):
Good work, please check my comments below

Q1W2E3R4T5Y6U7I8a (commented):
it is right?

Q1W2E3R4T5Y6U7I8a:
0523cdadee97c3bd0c93c09a3f7888e79acc9c45
is the second commit correct?

src/herbivoresAndCarnivores.js
Q1W2E3R4T5Y6U7I8a: I changed the code, please check

Q1W2E3R4T5Y6U7I8a:
it is right? 89050dc27b102f8c68667a4f5fbeaef31ba86ddb

Viktor-Kost (approved):

PR #2119, opened 2023-09-12
ttask animal

kalkvova:


lerastarynets (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
```
maxim2310, line 5:
```suggestion
```

maxim2310 (approved):
Good work!!!

PR #2125, opened 2023-09-15
Solution

HachevskyyMarkiyan:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    Animal.alive.push(this);
+  }
+
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
+    }
+
+    if (prey.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(prey), 1);
```
loralevitska, line 38: In this case better use filter array method

loralevitska (changes requested):
Almost perfect) Let's improve

src/herbivoresAndCarnivores.js
HachevskyyMarkiyan: Done ✅

loralevitska (approved):
Good job!

PR #2127, opened 2023-09-17
solution

helenmarth:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    super();
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
```
maxim2310, line 25: need to add constructor

maxim2310 (changes requested):

etojeDenys (changes requested):
still not fixed
image

src/herbivoresAndCarnivores.js
helenmarth: Screenshot 2023-09-18 at 21 17 03

when i add the constructor i receive following eslint error

helenmarth: when i add the constructor i receive following eslint error
Screenshot 2023-09-18 at 21 21 58

etojeDenys (approved):
awesome

PR #2130, opened 2023-09-20
adding solution

bormaxvel:
I will be glad if you tell me why my commented version did not work

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  // static alive = []; Parsing error: Unexpected token =
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+  }
 }
 
+Animal.alive = []; // For some reason, the 4th line didn't work for me
+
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  // static alive = []; Parsing error: Unexpected token =
```
lerastarynets, line 4: The error you're encountering, "Parsing error: Unexpected token =", is likely because the static class field assignment syntax using = is not supported by the JavaScript version you are using or the environment where you are running this code.

Static class field assignments using = are part of more recent JavaScript specifications (such as ECMAScript 2019 and later) and may not be supported in older JavaScript engines or environments.

To resolve this issue, you can initialize the alive property inside a constructor or a static method instead of using static class field assignment.
```
class Animal {
  constructor() {
    if (!Animal.alive) {
      Animal.alive = [];
    }
  }
}

// Usage
const animal = new Animal();
console.log(Animal.alive); // []
```

lerastarynets (approved):
Well done! Short and clear, I like it)

PR #2131, opened 2023-09-20
solution

Liubomyr19:
**Read [the guideline](https://github.com/Liubomyr19/js_task-guideline/blob/master/README.md) before start**

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
+    Animal.alive.push(this);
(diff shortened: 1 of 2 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
lerastarynets, line 7: The 'hidden' property should belong to the Herbivore class, not the Animal class, as it's specific to Herbivores. Move this property to the Herbivore class constructor.

lerastarynets (changes requested):

sTorba24 (approved):
It's guideline :) not needed to replace smth in url) 
image


Approved

PR #2136, opened 2023-09-21
Herbivores_and_Cornivores

VasylKurtiak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100,) {
```
etojeDenys, line 4:
```suggestion
  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
etojeDenys, line 15: you can just use filter method

etojeDenys (changes requested):

plodnikate (changes requested):
Check all comments please

etojeDenys (approved):
lgtm

PR #2137, opened 2023-09-21
Animal solution

hyrych-max:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
maxim2310, line 6: this is not necessary, could remove it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+
+    if (this.health > 0) {
+      Animal.alive.push(this);
+    }
```
maxim2310, line 10: this is redundant, because here you have health 100 as default

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
```
maxim2310, line 31: also need to add constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
-  // write your code here
+  bite(animal) {
+    if (
+      !(animal instanceof Carnivore)
+      && (animal instanceof Herbivore && !animal.hidden)
+    ) {
```
maxim2310, line 36:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

maxim2310 (changes requested):

maxim2310 (changes requested):
.

sTorba24 (approved):

PR #2138, opened 2023-09-21
Solution

mrogozhinskaya:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.health = health || 100;
+    this.name = name;
```
maxim2310, line 6:
```suggestion
  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+    super(health, name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
```
maxim2310, line 24: also need to add constructor

maxim2310 (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
-  // write your code here
+  bite(herb) {
+    if (herb.hidden === false) {
```
maxim2310, line 26: also need to check if its a herbivore

maxim2310 (changes requested):

etojeDenys (approved):
good job

PR #2151, opened 2023-09-26
Develop

GIroSTarK:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    this.hidden = false;
+    Animal.alive.push(this);
```
VitaliyBodnar, line 16: `Animal.alive.push(this);` should be only in Animal class, don't need to duplicate it

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #2154, opened 2023-09-28
solution, eslint parser fix

donizer:
Looks like after `npm i` the parser property disappeared from .eslintrc.js, so I kept getting an error when trying to create a static var in Animal.

So I added it manually.


donizer:
And GitHub installs dependencies as well, so it was useless

VitaliyBodnar (approved):
Good job!

PR #2155, opened 2023-09-28
add solution Herbivores and Сarnivores

A1eeex:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(health, name) {
+    this.health = 100;
```
Helen-arch, line 5: we don't receive prop health via constructor, so you can declare it directly after class Animal and just once

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
+
+    if (animal.health <= 0) {
+      const getAliveAnimal = Animal.alive.filter((item) => item.health > 0);
+
+      Animal.alive = getAliveAnimal;
+    }
+  }
 }
 
+Animal.alive = [];
```
Helen-arch, line 36: I think it should be static

Helen-arch (changes requested):
Great job! just a few suggestions

src/herbivoresAndCarnivores.js
A1eeex: ESLint doesn't allow you to do it any other way

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      const getAliveAnimal = Animal.alive.filter((item) => item.health > 0);
```
ArtemTeslenko, line 29: you search along the entire length of the array after each bite, it is better to splice one element by index, `findIndex` of animal with 0 health and splice. Also if you use filter it is better to name variables by meaning, not `item`.

ArtemTeslenko (approved):
Good job)) There is a comment, please consider next time

PR #2156, opened 2023-09-28
add classes

Helen-arch:


A1eeex (approved):
Nice!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
polosanya, line 4: Let's move it outside the class to pass the tests
`Animal.alive = []`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+
+  health = 100;
```
polosanya, line 6: Move it to the constructor

polosanya (changes requested):
Nice! Let's improve your code a bit)

StasSokolov (approved):
Great job!

PR #2160, opened 2023-09-30
solution

OlegPopovych:
Hi! The tests run successfully, although I have doubts about my solution. When I tried to use `static alive = [];` I took a linter error, and then I used another way to define static prop. I tested it, instances of Animal don't have access to `Animal.alive`.

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore) {
+      return;
+    }
+
+    if (!animal.hidden) {
+      animal.health -= 50;
+    }
+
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
mpohorenyi, line 21: 3. [CHECKLIST] - if the function has a name hide this function should just hide but not toggle

mpohorenyi (approved):
Good Job!
Please note the following comments ↓

src/herbivoresAndCarnivores.js
OlegPopovych: Thanks for noticing that! I didn't find a link to the checklist at first in readme :D

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore) {
+      return;
+    }
+
+    if (!animal.hidden) {
+      animal.health -= 50;
+    }
```
StasSokolov, line 32: let's move this into one block 
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
      animal.health -= 50;
    }
```

StasSokolov (changes requested):
Good job, let's improve your code a bit more!

src/herbivoresAndCarnivores.js
OlegPopovych: Thanks, it looks better

ArtemTeslenko (approved):
Good job!

PR #2164, opened 2023-10-01
add task solution

adioskv:


sintax1s (approved):
Good Job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+
+    // Linter is not allowing us to use static =(
+
+    if (!Animal.alive) {
+      Animal.alive = [];
```
StasSokolov, line 11: Let's move it outside the class `Animal.alive = []`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
+        Animal.alive = Animal.alive.filter(beast => beast.name !== animal.name);
```
StasSokolov, line 34: What if a few animals have the same name? I think it's better to filter by health

StasSokolov (changes requested):
Good job, only few comments left to fix!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
```
ArtemTeslenko, line 20: if you once hid this beast, you could not change it in the future. Think how to make it work both ways

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
+        Animal.alive = Animal.alive.filter(beast => beast.health !== 0);
```
ArtemTeslenko, line 30: you search along the entire length of the array after each bite, it is better to splice one element by index, findIndex of animal with 0 health and splice

ArtemTeslenko (changes requested):
Good job) But you can not change property `hidden` after first change, fix it.

src/herbivoresAndCarnivores.js
adioskv: image
Check checklist before commenting solutions)

src/herbivoresAndCarnivores.js
adioskv: image

ArtemTeslenko: Sorry, my mistake)

ArtemTeslenko (approved):

src/herbivoresAndCarnivores.js
adioskv: i can add new function unhide or unhide beast when he was attacked, but his status was `hidden`.
Also, there is zero information about changing property `hidden` backwards, so just read the task and comment after it :)

PR #2165, opened 2023-10-01
made task solution

sintax1s:


yulabakun (approved):
Looks all good to me. 
I suppose you should add health property to Animal constructor in case user wants to create animal with custom health value, but I'm not sure about that.

StasSokolov (approved):
LGTM🔥

PR #2166, opened 2023-10-02
Classes and inheritance task

ElenaGrasovskaya:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.type = 'carnivore';
+    Animal.addAnimal(this);
+  }
+
+  bite(victim) {
+    if (victim.type === 'herbivore' && victim.hidden === false) {
```
sviatikkondur, line 42: small suggestion: use can just use NOT(!) operator instead of compare to false

sviatikkondur (approved):
Good job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+    this.health = 100;
+    this.name = name;
+  };
+
+  static addAnimal(animal) {
+    this.alive.push(animal);
+  }
+
+  static removeAnimal(animal) {
+    const index = this.alive.findIndex(el => el === animal);
```
ArtemTeslenko, line 14: specify what index are you looking for, when your class will grow up it will be easier to read, for example killedAnimalIndex, also `el` - is not good name for variable

ArtemTeslenko (changes requested):
Good job! Comparison with false is not the best practice, fix it.

src/herbivoresAndCarnivores.js
ElenaGrasovskaya: fixed... silly me  :)

src/herbivoresAndCarnivores.js
ElenaGrasovskaya: renamed

ArtemTeslenko (approved):
Good job!

PR #2169, opened 2023-10-02
add solution Animals

denzzlle171:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  die() {
+    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
+  }
```
superpooperxxx, line 13: This method is filtering out ALL animals that has `health <= 0`. This is a general action that might have side effect for us (removing other animals). I think we should remove the animal we just killed (`this`)

superpooperxxx (changes requested):
Good job🔥 Just one concerning moment

superpooperxxx (changes requested):
.

ArtemTeslenko (approved):
Good job!

PR #2172, opened 2023-10-02
Solution

Maksim114488:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
+  hide(name) {
```
o-drozzdyk, line 19: name argument is not necessary here

o-drozzdyk (approved):
Good job!

src/herbivoresAndCarnivores.js
polosanya: ➕ , good point

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
```
polosanya, line 27: No need to duplicate it here. You already have in Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden !== true) {
```
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
```
This condition can be simplified

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+      victim.health -= 50;
+    }
+
+    if (victim.health <= 0) {
+      Animal.alive = Animal.alive.filter(item => item !== victim);
```
polosanya, line 35: Would be great to use more meaningful names than `item`

polosanya (changes requested):

polosanya (approved):
Not all comments are fixed. I'm approving it, but please check them one more time)
Maybe you didn't push all the changes?

PR #2174, opened 2023-10-02
solution

OlhaKoziuk:


dolphinnq (approved):
Great job! :)

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
-  // write your code here
+  bite(animal) {
+    const DAMAGE = 50;
+
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= DAMAGE;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
StasSokolov, line 31: let's use `.filter()` instead `.splice()`

StasSokolov (changes requested):
Awesome job, let's fix only one comment

PR #2175, opened 2023-10-02
solution

vlevik:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+
+      if (victim.health <= 0) {
+        const victimIndex = Animal.alive.indexOf(victim);
+
+        Animal.alive.splice(victimIndex, 1);
```
YePolishchuk, line 32: [CODE PERFOMANCE] - Try to avoid nesting loops (indexOf + splice), use filter instead

YePolishchuk (approved):
Good job!!!

src/herbivoresAndCarnivores.js
StasSokolov: @vlevik check this point please

StasSokolov (changes requested):
Great job, let's improve only one part of application

ArtemTeslenko (approved):
Good job!

PR #2176, opened 2023-10-02
Develop

momos1703:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.name = name;
```
etojeDenys, line 27: this line is redundant cause you call super

momos1703: Than you. Done!

Esceype (approved):
GJ🚀🚀🚀

PR #2178, opened 2023-10-02
task solution

YePolishchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
```
BetterSol, line 16: it's not necessary to use a constructor here, our class extends the Animal class anyway;)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
```
BetterSol, line 27: same here

BetterSol (approved):
Awesome work! Suggested a few minor changes that could make your code even better)

StasSokolov (approved):
Awesome job🔥

PR #2180, opened 2023-10-02
SOLUTION

julwer1k:
image

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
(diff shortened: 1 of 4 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      const index = Animal.alive.indexOf(animal);
+
+      Animal.alive.splice(index, 1);
```
nazarmatsevych, line 34: Do not manipulate with array in this way, your filter method instead

nazarmatsevych (changes requested):

volodymyr-soltys97 (approved):
Great work 🔥

PR #2182, opened 2023-10-02
task solved, first iteration

BetterSol:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
 }
-
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
StasSokolov, line 25: please use `.filter()` instead `.splice()`

StasSokolov (changes requested):
Great job, let's improve your code a bit more!

src/herbivoresAndCarnivores.js
BetterSol: Thanks, looks much better now)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,28 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, hidden) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
polosanya, line 7: Only herbivores have this property, so it's better to move it to their class

polosanya (approved):
Good job! I'm approving it, but pay attention to the comment)

PR #2185, opened 2023-10-02
Solution

PaninDmytro:


Kuper18 (approved):
Good job!)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
+    Animal.alive.push(this);
```
superpooperxxx, line 8: Only `herbivore` should have property `hidden`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
superpooperxxx, line 13: Should go to `Herbivore` class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  hide() {
+    super.hide();
+  }
```
superpooperxxx, line 21: You should have a property `this.hidden = false;` and after that write a method to hide herbivore (and only herbivore)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!this.hidden && animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
```
superpooperxxx, line 28: So again, as only only Herbivore can be hidden this check is redundant `!this.hidden`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+      if (animal.health <= 0) {
+        Animal.alive = Animal.alive.filter((item) => item !== animal);
+      }
```
superpooperxxx, line 31: 👍🏻👍🏻👍🏻
I'd suggest not using such general words as `item` and moving filtering functionality to some static method

superpooperxxx (changes requested):
Good job🔥 but you can better

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive = this.stayAliveAnimals(Animal.alive, animal);
+      }
+    }
+  }
+
+  stayAliveAnimals(animals, animal) {
```
polosanya, line 35: Do you really need to pass all animals too?

PR #2186, opened 2023-10-02
animal planet JS

HardCodWorker:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  }
+
+  bite(victim) {
+    if (victim instanceof Herbivore) {
+      if (!victim.hidden) {
+        victim.health -= 50;
+      }
+    }
+
+    if (victim.health <= 0) {
```
annabasenyuk, line 36: HI) You can use === rather than <= because 'healthy' will never be lower than 0.

annabasenyuk (approved):
Hi, good job) I added a little mark)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(health, name) {
+    super(name, health);
+  }
+
+  bite(victim) {
+    if (victim instanceof Herbivore) {
+      if (!victim.hidden) {
```
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
```
You can just combine the conditions

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+        victim.health -= 50;
+      }
+    }
+
+    if (victim.health <= 0) {
+      Animal.alive = Animal.alive.filter((value) => value !== victim);
```
polosanya, line 37: Would be great to use more meaningful names than `value`

src/herbivoresAndCarnivores.js
polosanya: Not critical, but relevant comment)

polosanya (changes requested):
Almost perfect!

HardCodWorker:
Done

StasSokolov (approved):
Great job, i'm approving it🔥

PR #2190, opened 2023-10-03
add solution

HokageHEL:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    if (Animal.alive.includes(this)) {
+      Animal.alive.splice(Animal.alive.indexOf(this), 1);
+    }
```
etojeDenys, line 15: you can just use the filter method

etojeDenys (approved):
good job

PR #2192, opened 2023-10-03
solution

vladkorobka:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive = [...Animal.alive].filter(animal => {
```
VitaliyBodnar, line 40: don't need to create a copy or array, because method filter did it already 
```suggestion
        Animal.alive = Animal.alive.filter(animal => {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    if (this.hidden) {
+      this.hidden = false;
+
+      return;
```
VitaliyBodnar, line 23: don't see any sense for adding empty return here
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
```
VitaliyBodnar, line 36:
```suggestion
    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```

VitaliyBodnar (changes requested):

src/herbivoresAndCarnivores.js
vladkorobka: I've done it, because I should finish execution of method hide(). In other way, I will have this.hidden equals true again, after condition.
I could use "if.. else", but other mentors said don't do that in this way.

VitaliyBodnar: method hide should only hide element, not toggle 
so, in this case you can write it like this:
```
hide() {
  if (!this.hidden) {
   this.hidden = true;
  }
}
```

VitaliyBodnar (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #2193, opened 2023-10-03
add task solution

annabasenyuk:
Hi) Added the task solution, check it please)

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,9 +1,11 @@
 'use strict';
 
+const COMPLEATLY_HEALTHY = 100;
+const BITE_OF_ANIMAL = 50;
+const NOT_HEALTHY = 0;
+
 class Animal {
   constructor(name) {
-    const COMPLEATLY_HEALTHY = 100;
-
     this.health = COMPLEATLY_HEALTHY;
     this.name = name;
     Animal.alive.push(this);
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    const COMPLEATLY_HEALTHY = 100;
```
polosanya, line 5: Better move this variable outside of the constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(health, name);
+  }
+
+  bite(prey) {
+    const BITE_OF_ANIMAL = 50;
+    const NOT_HEALTHY = 0;
```
polosanya, line 34: Same here. Move all variables to the top

polosanya (changes requested):
Almost perfect!

src/herbivoresAndCarnivores.js
annabasenyuk: Thanks for your comment) I hope I understood correctly the comments)

polosanya (approved):
Well done! 👍

PR #2194, opened 2023-10-03
add task solution

Stim1:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+    Animal.alive.push(this);
+  }
+
+  removeFromAlive() {
+    Animal.alive = Animal.alive.filter(animal => animal.health !== 0);
```
ivaZaiets, line 11: You also can use smth like this:
``` js
removeFromAlive(animal) {
    Animal.alive = Animal.alive.filter(aliveAnimal => aliveAnimal !== animal);
  }
```
Where you compare link of the both objects

ivaZaiets (approved):
Good job! Good solution!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore
+      && !victim.hidden) {
```
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+      victim.health -= 50;
+      this.removeFromAlive();
```
polosanya, line 32: Would be great to check the victim's health before calling removeFromAlive method. It will help to avoid extra calls

polosanya (changes requested):
Almost perfect!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+    Animal.alive.push(this);
```
superpooperxxx, line 7: [non-blocker]: I'd create a separate method for this e.g. addToAlive(animal)

src/herbivoresAndCarnivores.js
superpooperxxx: It sounds more logical to me. This method should remove particular animal from alive, what if you call this method expecting, that it will filter out your desired animal, but some another animal also has health 0?

superpooperxxx (changes requested):
Good job🔥 I think we can improve one thing

ArtemTeslenko (approved):
Good job!

PR #2195, opened 2023-10-03
Solution

dsfreedom:


MVoutch (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
```
MVoutch, line 6: No need for a logical operator here

`this.health = 100;`

MVoutch (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
+    const bittenAnimal = Animal.alive.find(animal => animal === name);
```
polosanya, line 26:
```suggestion
  bite(targetAnimal) {
```
Actually, this part is redundant as bite method should receive the whole object as an argument, not just name.
Pay attention to the task's description:
image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+    if (bittenAnimal instanceof Herbivore && bittenAnimal.hidden === false) {
```
polosanya, line 28:
```suggestion
    if (bittenAnimal instanceof Herbivore && !bittenAnimal.hidden) {
```
You can simplify the condition

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = 100;
```
```suggestion
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;
```
You can pass the default parameter as follow:

polosanya (changes requested):

src/herbivoresAndCarnivores.js
dsfreedom: Did like this at first and than fchanged it and forgot to return it. Thanks for pointing that out

polosanya (approved):
Well done! 🚀

PR #2196, opened 2023-10-03
task solution

DenysShurubura:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+    this.hidden = hidden || false;
+  }
+
+  hide() {
+    if (this.hidden) {
+      this.hidden = false;
+    }
+
+    this.hidden = true;
+  }
```
ArtemTeslenko, line 33: seems like toggle, but more difficult, this method should not toggle value, only hide

ArtemTeslenko (changes requested):
Good job! There are some issues, fix setting default values and hide method))

polosanya (approved):
👍

PR #2198, opened 2023-10-03
Solution

MVoutch:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
```
superpooperxxx, line 35: [non-blocker]: I think the parameter should not be called `herbivore`, cause later this `herbivore instanceof Herbivore` looks confusing

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+
+  die() {
+    const animal = Animal.alive;
+    const index = animal.indexOf(this);
+
+    if (animal !== 0) {
+      animal.splice(index, 1);
+    }
+  }
 }
```
superpooperxxx, line 18: I like the solution, but
1. `Animal.alive` is an array of animals and you save it to variable `animal`
2. `if (animal !== 0)` this is always true because as I said this is an array.

superpooperxxx (changes requested):
Good job🔥 Only one issue to fix

mpohorenyi (approved):
Great Job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
+
+  die() {
+    const animal = Animal.alive;
+    const index = animal.indexOf(this);
```
superpooperxxx, line 25: Just `index = Animal.alive.indexOf(this);`. Variable `animal` looks confusing, because seems like it should hold single value, but actually you save an array there.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
+    super(name, health, hidden);
```
superpooperxxx, line 15: Not sure you need to pass `hidden` here `super(name, health, hidden);`. `class Animal` doesn't have this property

superpooperxxx (approved):
Good job🔥

PR #2199, opened 2023-10-03
first solution

Clover247:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,51 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    this.type = 'herbivore';
```
polosanya, line 24: This property is actually redundant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,51 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.type = 'carnivore';
+    Animal.addAnimal(this);
+  }
+
+  bite(victim) {
+    if (!victim.hidden && victim.type === 'herbivore') {
```
polosanya, line 41: Better check the instance of the animal instead of the type
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Operators/instanceof

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,51 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.type = 'carnivore';
+    Animal.addAnimal(this);
```
polosanya, line 37: Move it to the constructor of the animal so you don't need to duplicate it for carnivores and herbivores

polosanya (changes requested):

superpooperxxx (approved):
🔥🔥🔥

PR #2200, opened 2023-10-04
Solution

hrysta:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static removeAnimalByName(name) {
+    Animal.alive = Animal.alive.filter(animal => animal.name !== name);
+  }
```
superpooperxxx, line 13: What if multiple animals have the same name ? Would you kill all of them ?

superpooperxxx (changes requested):
Good job🔥 Just one concerning moment

PR #2201, opened 2023-10-04
animals task solution

sglupa:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
-}
+  static alive = [];
```
polosanya, line 4: Move it out of the class to pass the tests
`Animal.alive = []`

PR #2203, opened 2023-10-05
first

Vladimir-Zadorozhnyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
anastasiia-tilikina, line 11:
```suggestion
    Animal.alive = Animal.alive.filter(animal => animal.health >= 0);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb.hidden === false) {
```
anastasiia-tilikina, line 29:
```suggestion
    if (!herb.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+      herb.health -= 50;
+    }
+
+    if (herb.health === 0) {
```
anastasiia-tilikina, line 33: you don't need this condition.
your method die filters animals according to health

anastasiia-tilikina (changes requested):
well done!
just a couple small improvements

src/herbivoresAndCarnivores.js
Vladimir-Zadorozhnyi: Herbivores should be removed from 'Animal.alive' array,                                              
  when their health is 0.

src/herbivoresAndCarnivores.js
Vladimir-Zadorozhnyi: if I try change condition but i got error / 'bite' method shouldn't decrease 'health' property of the carnivores/ I think one of the tests compares the value true and false but none of the others, for example undefined, so in the condition I use false.

src/herbivoresAndCarnivores.js
Vladimir-Zadorozhnyi: Thanks for tip. I changed code.

src/herbivoresAndCarnivores.js
maxim2310: you just need to add validation that your herb its Herbivore but not Carnivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
maxim2310, line 24:
```suggestion
  hide() {
    this.hidden = !this.hidden;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
```
maxim2310, line 20:
```suggestion
  constructor(name, health) {
    super(name, health);
    this.hidden = hidden;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
```
maxim2310, line 27: need to add constructor for Carnivores

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.name = name;
+  }
+
+  bite(herb) {
+    if (herb.hidden === false) {
```
sTorba24, line 34:
```suggestion
  bite(animal) {
    if (animal instanceof Herbivore && !animal.hidden) {
      animal.health -= 50;
    }
    
    if (animal.health <= 0) {
      Animal.die();
    }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      herb.health -= 50;
+      herb.die();
```
sTorba24, line 36: remove it, not needed to use .die() method if animal had damaged once

src/herbivoresAndCarnivores.js
sTorba24: name param `animal` because it's animal, then you check which instanceof it is

sTorba24 (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
sTorba24, line 6:
```suggestion
  constructor(name) {
    this.name = name;
    this.health = 100;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
```
sTorba24, line 19:
```suggestion
  constructor(name) {
    super(name);
    this.hidden = false;
```

src/herbivoresAndCarnivores.js
sTorba24: you can't create instance with health <100, it's not logically

src/herbivoresAndCarnivores.js
sTorba24: same here, it's not hidden by default

sTorba24 (changes requested):

VitaliyBondarenko1982 (approved):
GJ!

PR #2206, opened 2023-10-07
js herbivores and carnivores

IvanOrinchak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+      if (element.name === this.name) {
+        element.hidden = true;
+        this.hidden = true;
+
+        return element;
+      }
+
+      return element;
+    });
+  }
```
nazarmatsevych, line 27: This is an instance, so no need to use map here, you can simply change this.hidden to true
```suggestion
    this.hidden = true;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+    if (!(canBite instanceof Carnivore) && !canBite.hidden) {
+      Animal.alive = Animal.alive.map(element => {
+        if (element.name === canBite.name) {
+          element.health = element.health - 50;
+
+          return element;
+        }
+
+        return element;
+      });
```
nazarmatsevych, line 41: Same here, no need to use map

nazarmatsevych (changes requested):
Good work,  left comments to be fixed

VitaliyBondarenko1982 (approved):
Well done

PR #2207, opened 2023-10-07
problem with static property

Viktoriia-Vita:
why I can`t use "static alive = [ ]"???? linter give me an error " unexpected ="

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,51 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  // static alive = [];
+
+  constructor(health = 100, name) {
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
+  }
+
+  // static deleteDeadAnimal() {
+  //   Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
+  // }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(health, name);
+    this.hidden = false;
+  }
+
+  hide() {
+    if (this.hidden === false) {
+      this.hidden = true;
+    } else {
+      this.hidden = false;
+    }
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(health, name) {
+    super(name, health);
+  }
+
+  bite(victum) {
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,51 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(health, name) {
+    super(name, health);
+  }
+
+  bite(victum) {
+    if (victum.hidden === false
```
TarasHoliuk, line 40:
```suggestion
    if (!victum.hidden
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,51 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  // static alive = [];
+
+  constructor(health = 100, name) {
```
TarasHoliuk, line 6:
```suggestion
  constructor(name, health = 100) {
```
let's put obligate property in the first place

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,51 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(health, name) {
+    super(name, health);
+  }
+
+  bite(victum) {
```
TarasHoliuk, line 39:
```suggestion
  bite(victim) {
```
typo

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,51 @@
...
+    super(health, name);
+    this.hidden = false;
+  }
+
+  hide() {
+    if (this.hidden === false) {
+      this.hidden = true;
+    } else {
+      this.hidden = false;
+    }
```
TarasHoliuk, line 30: sounds logical to have the possibility set `hidden = false`, but it will not match with the method name (hide), so let's keep only the part where we set `hidden = true`

TarasHoliuk (changes requested):

TarasHoliuk:
> why I can`t use "static alive = [ ]"???? linter give me an error " unexpected ="

Unfortunately, in this task is used the old version of parser in the linter config, which doesn't understand such code (ECMA SCRIPT 6/7)

Esceype (approved):
Well done👍

PR #2208, opened 2023-10-08
solution

mifyy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor() {
+    this.name = '';
```
superpooperxxx, line 5: constructor should accept at least `name`. Because this way all of your animal will have `name = ''`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
+    this.hidden = false;
+  }
```
superpooperxxx, line 17: Again constructor should accept parameters name and health for Animal constructor which you pass via `super(name, health)`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden === false) {
+      target.health -= 50;
+    }
```
superpooperxxx, line 28: Instead of `target.hidden === false` we can write `!target.hidden`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+    if (target.health <= 0) {
+      Animal.alive = Animal.alive.filter((x) => x !== target);
+    }
```
superpooperxxx, line 32: 1. What is `x` ? Let's be more descriptive, you are looping though alive animals not Xs
2. You can put this logic in a separate static method in Animal class `Animal.alive = Animal.alive.filter((x) => x !== target);`

superpooperxxx (changes requested):
Good job🔥, but you have crucial mistakes

mifyy:
Please don't check the task if you see this message, I requested a review before applying changes by mistake

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
+
+  kill() {
+
+  }
```
ArtemTeslenko, line 25: no need to save an empty method

ArtemTeslenko (approved):
Good job!

PR #2209, opened 2023-10-08
Solution

RVDotsenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, bite) {
+    super(name);
+
+    Animal.addAlive(this);
+  }
```
etojeDenys, line 41: you can just call Animal.addAlive one time in the Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
 class Animal {
-  // write your code here
+  static addAlive(animal) {
+    this.alive.push(animal);
+  };
+
+  static removeFromAlive(animal) {
+    const indexOfAnimal = this.alive.indexOf(animal);
+
+    this.alive[indexOfAnimal] = 0;
```
etojeDenys, line 11: let's just use filter

etojeDenys (changes requested):

lerastarynets (approved):
You've got this, keep pushing forward!

PR #2213, opened 2023-10-11
Testing

Olesichka:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  hide() {
+    this.hidden === true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
```
Julia14012000, line 12: After the bite, you need to check whether the target no longer has health (health is 0). If so, it is removed from the alive list using the filter method.

Add this, please

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+      this.health -= damage;
+      if (this.health <= 0) {
+        Animal.removeFromAliveList(this);
+      }
+    }
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
```
Julia14012000, line 8: Add constructor(name, health).
Constructor constructor(name, health): The Herbivore constructor must accept two parameters - name (name of the animal) and health (health of the animal). This constructor first calls the constructor of its parent Animal class, passing it name and health. This is done using super(name, health);. Thus, the Herbivore class inherits the properties and methods of the Animal class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
 'use strict';
 
 class Animal {
```
Julia14012000, line 3: You can improve the code.

  The Animal class has a constructor that accepts two parameters: name (the name of the animal) and health (the level of the animal's health). Defaults to 100 if health is not provided.

Name and health properties: The constructor assigns the values of the name and health parameters to the corresponding properties this.name and this.health. Thus, each object of the Animal class has a name and a health level.

Animal.alive Array: The Animal class also has a static Animal.alive array that is used to store all created Animal objects. Whenever a new Animal object is created, it is automatically added to this array using Animal.alive.push(this);.

So, the Animal class allows you to create objects that represent different animals and store them in the Animal.alive array for later use.

Julia14012000 (changes requested):
Please, check comments

PR #2215, opened 2023-10-11
implementation

Daniilart01:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        const index = Animal.alive.indexOf(herbivore);
+
+        Animal.alive.splice(index, 1);
```
etojeDenys, line 32: let's just use the filter method

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #2220, opened 2023-10-13
Solution

kseniia-chepur:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+
+    if (obj instanceof Herbivore && !obj.hidden) {
+      obj.health -= 50;
+    }
+
+    const i = Animal.alive.findIndex(beast => !beast.health);
+
+    if (i !== -1) {
+      Animal.alive.splice(i, 1);
+    }
```
etojeDenys, line 43: you can just call the filter

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+  }
+
+  bite(obj) {
+    if (obj instanceof Carnivore) {
+      return;
+    }
+
+    if (obj instanceof Herbivore && !obj.hidden) {
+      obj.health -= 50;
+    }
```
etojeDenys, line 37:
```suggestion
    if (obj instanceof Herbivore && !obj.hidden) {
      obj.health -= 50;
    }
```
why do we need that check

etojeDenys (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(obj) {
+    if (obj instanceof Herbivore && !obj.hidden) {
+      obj.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter(beast => beast.health);
```
sTorba24, line 35:
```suggestion
    Animal.alive = Animal.alive.filter(beast => beast.health <= 0 );
```

sTorba24: -10 will give `true` in result(in case we add some actions with dmg different from 50)

sTorba24 (approved):

src/herbivoresAndCarnivores.js
kseniia-chepur: Clear, have fixed!

PR #2222, opened 2023-10-14
Solution

YevgeniySavchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      const index = Animal.alive.findIndex(beast => beast === animal);
+
+      Animal.alive.splice(index, 1);
```
etojeDenys, line 40: we can just use filter method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    Animal.alive.push(this);
```
etojeDenys, line 29: let's push to the alive array in the Animal constructor

etojeDenys (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
volodymyr-soltys97, line 28:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

volodymyr-soltys97 (approved):
🔥

PR #2228, opened 2023-10-17
Solution

InnaSh23:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(health, name) {
+    super(health, name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
```suggestion
    this.hidden = true;
```
This method should only hide an animal, not to toggle hidden value

nazarmatsevych (approved):
Awesome 🔥 
Left minor comment

PR #2233, opened 2023-10-19
develop

nikkias:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
```
etojeDenys, line 28: you can just push it one time in Animal constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    if (!(herbivore instanceof Herbivore) || herbivore.hidden) {
+      return;
+    }
+
+    herbivore.health -= 50;
+
+    if (herbivore.health === 0) {
+      Animal.alive = Animal.alive.filter(animal => (
+        animal.name !== herbivore.name
+      ));
```
etojeDenys, line 40: we can filter by health it will look simpler

etojeDenys (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  // eslint-disable-next-line no-useless-constructor
+  constructor(name) {
+    super(name);
+  }
```
etojeDenys, line 28: eslint gives you the tip that you can remove this constructor at all

etojeDenys (changes requested):

volodymyr-soltys97 (approved):

PR #2234, opened 2023-10-20
task solution

TenetaJ:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden === true) {
```
VitaliyBodnar, line 33:
```suggestion
    if (animal instanceof Carnivore || animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+    }
+
+    animal.health -= 50;
+
+    if (animal.health <= 0) {
+      const index = Animal.alive.indexOf(animal);
+
+      if (index > -1) {
+        Animal.alive.splice(index, 1);
+      }
```
VitaliyBodnar, line 44: it's better to use method filter here instead of indexOf and splice

VitaliyBodnar (changes requested):

src/herbivoresAndCarnivores.js
TenetaJ: done

src/herbivoresAndCarnivores.js
TenetaJ: done

maxim2310 (approved):
Good work!!!

PR #2235, opened 2023-10-21
solution task

Butovych:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    allAnimal.push(this);
+  }
+
+  bite(name) {
+    if (name.hidden === true || name instanceof Carnivore) {
```
VitaliyBondarenko1982, line 35:
```suggestion
    if (name.hidden || name instanceof Carnivore) {
```
 redundant to compare with boolean

VitaliyBondarenko1982 (approved):
Looks good.

PR #2236, opened 2023-10-23
solution

maksym-kostetskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbi) {
```
loralevitska, line 25: Don't cut variable naming
```suggestion
  bite(herbivore) {
```

OR

```suggestion
  bite(animal) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    if (herbi instanceof Herbivore && herbi.hidden !== true) {
+      herbi.health -= 50;
+    }
+
+    if (herbi.health === 0) {
+      const indexOfHerbi = Animal.alive.indexOf(herbi);
+
+      Animal.alive.splice(indexOfHerbi, 1);
```
loralevitska, line 33: Better to use filter array method for this

loralevitska (changes requested):
Good job! There are things that should be fixed:

Viktor-Kost (approved):

PR #2243, opened 2023-10-27
Solution

yuliiababych:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
nazarmatsevych, line 15: Try not to manipulate with array in this way, use filter method instead

nazarmatsevych (changes requested):
Good work, let's improve your solution a bit

nazarmatsevych (approved):
👍

PR #2245, opened 2023-10-28
the task was completed

Viktoria-Moroz:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
```
anastasiiavorobiova, line 29:
```suggestion

```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  bite(animal) {
+    if (animal && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter((beast) => beast.health > 0);
```
anastasiiavorobiova, line 35: It's better to move this method to the Animal class and call it here

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/herbivoresAndCarnivores.js
etojeDenys: you can remove the constructor here
image

etojeDenys (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #2246, opened 2023-10-28
js_herbivores_and_carnivores

Donovanoff:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
VitaliyBodnar, line 32:
```suggestion
    if (animal instanceof Herbivore && animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  die(animal) {
+    const index = Animal.alive.findIndex(beast => beast === animal);
+
+    Animal.alive.splice(index, 1);
```
VitaliyBodnar, line 13: it's better to use method filter here as it will make a code shorter

VitaliyBodnar (changes requested):

src/herbivoresAndCarnivores.js
Donovanoff: in your code animal.hidden - true, but we need false

etojeDenys (approved):
good job

PR #2250, opened 2023-11-01
solution

ArtemSydun:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.health = health || 100;
```
polosanya, line 5:
```suggestion
  constructor(name, health = 100) {
    this.health = health;
```
You can just set the default value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(health, name, hidden) {
+    super(health, name);
+    this.hidden = hidden || false;
```
polosanya, line 15: Same here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
+    if (obj instanceof Herbivore && !obj.hidden) {
```
polosanya, line 25: Let's use more meaningful names than just `obj`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+      obj.health -= 50;
+      Animal.alive = Animal.alive.filter(target => target.health > 0);
```
```suggestion
      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
Same here. It's not a target as you go through all alive animals

polosanya (changes requested):
Almost done!

roman-mirzoian (approved):
Well done 👍

PR #2258, opened 2023-11-03
solution

MariiaSuper:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Carnivore) {
+      return;
+    }
```
VitaliyBodnar, line 29: this if is redundant here 
```suggestion
```

VitaliyBodnar (changes requested):
Almost done!

anastasiiavorobiova (approved):
Awesome work!

PR #2269, opened 2023-11-09
Add Solution

Eater228:


src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,55 @@
...
+      this.die();
+    }
+  };
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
nazarmatsevych, line 25: Don't manipulate with array in this way, use filter method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,55 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
-}
+  constructor(name, health = 100) {
+    super(name, health);
+  };
+
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden !== true) {
```
nazarmatsevych, line 46:
```suggestion
    if (target instanceof Herbivore && !target.hidden) {
```

same but shorter

src/herbivoresAndCarnivores.js
```diff
@@ -1,16 +1,55 @@
 'use strict';
 
 class Animal {
-  // write your code here
-}
+  // static alive = [];
```
nazarmatsevych, line 4: Remove commented code please

nazarmatsevych (changes requested):
Almost done, please check comments :)

src/herbivoresAndCarnivores.js
Eater228: I know, but it didn't work when I first wrote it

src/herbivoresAndCarnivores.js
Eater228: Why?

src/herbivoresAndCarnivores.js
maxim2310: remove it plese

src/herbivoresAndCarnivores.js
maxim2310: must work if you did everything correct

src/herbivoresAndCarnivores.js
maxim2310: because this is a more obvious method

maxim2310 (changes requested):
please fix previous comment

Esceype (approved):
Well done!

PR #2274, opened 2023-11-10
add my animals

AsyaYeromina:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    if (!this.hidden) {
+      this.hidden = true;
+    } else {
+      this.hidden = false;
+    }
```
```suggestion
   this.hidden = true;
```
this method should not toggle hidden value and only set it to true

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(creature) {
+    if (creature instanceof Carnivore) {
+      return;
+    }
+
+    if (!creature.hidden) {
+      creature.health -= 50;
+    }
```
```suggestion
    if (creature instanceof Herbivore && !creature.hidden ) {
      creature.health -= 50;
    }
```

You can combine this conditions

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+
+    if (creature.health <= 0) {
+      const index = Animal.alive.indexOf(creature);
+
+      Animal.alive.splice(index, 1);
```
nazarmatsevych, line 41: Don't manipulate with array in this way, use filter method instead

nazarmatsevych (changes requested):
good work, let's make some improvements

src/herbivoresAndCarnivores.js
AsyaYeromina: why is it a bad idea?

SanyaBratashchuk (approved):

PR #2275, opened 2023-11-10
Complete task

Annbpiu:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    if (!this.hidden) {
+      this.hidden = true;
+    } else {
+      this.hidden = false;
+    }
```
```suggestion
   this.hidden = true;
```

This method should not toggle the hidden value and only change it to "true"

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(creature) {
+    if (creature instanceof Carnivore) {
+      return;
+    }
+
+    if (!creature.hidden) {
+      creature.health -= 50;
+    }
```
```suggestion
    if (creature instanceof Herbivore && !creature.hidden) {
      creature.health -= 50;
    }
```

Combine this conditions

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+
+    if (creature.health <= 0) {
+      const index = Animal.alive.indexOf(creature);
+
+      Animal.alive.splice(index, 1);
```
nazarmatsevych, line 41: Don't manipulate with array in this way, use filter method instead

nazarmatsevych (changes requested):
Almost done, please check comments below

anastasiiavorobiova (approved):
Awesome work!

PR #2276, opened 2023-11-10
Solution

errisi:


MrHunterBoi (approved):
LGTM

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
```
roman-mirzoian, line 26: Avoid using "magic values".

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
```
roman-mirzoian, line 24: You already have a class named `Animal`, so to avoid confusion, it is better to name the function argument something else: aim, victim, target etc.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
```
roman-mirzoian, line 28:
```suggestion
      if (!animal.health) {
```

roman-mirzoian (changes requested):

src/herbivoresAndCarnivores.js
errisi: > You already have a class named `Animal`, so to avoid confusion, it is better to name the function argument something else: aim, victim, target etc.

i can't use 'static' because of eslint, now i failed the test but i fixed it

src/herbivoresAndCarnivores.js
errisi: > Avoid using "magic values".

i can't use 'static' because of eslint, now i failed the test but i fixed it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+const DAMAGE_AMOUNT = 50;
+
 class Carnivore extends Animal {
-  // write your code here
+  bite(aim) {
+    if (aim instanceof Herbivore && !aim.hidden) {
+      aim.health -= DAMAGE_AMOUNT;
+
+      if (!aim.health) {
+        Animal.alive.splice(Animal.alive.indexOf(aim), 1);
```
ksyvvkaa, line 31: you could use filter here, like it's suggested in [checklist.md](https://github.com/mate-academy/js_herbivores_and_carnivores/blob/master/checklist.md)
image

ksyvvkaa (approved):
nice job👻

roman-mirzoian (approved):
Well done 👍

PR #2278, opened 2023-11-10
solution

MarekSax:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
nazarmatsevych, line 31: Don't manipulate with array in this way,  use filter method instead

nazarmatsevych (changes requested):
Almost done, please check my comment

nazarmatsevych (approved):
Awesome 🔥

PR #2280, opened 2023-11-11
solution

yevtok:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
Anton-Kuchmasov, line 6: health of every Animal by default = 100,
that's why it's better to use global constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.filter(creature => creature.health > 0);
```
Anton-Kuchmasov, line 31: nice)

Anton-Kuchmasov (approved):
nice job!

roman-mirzoian (changes requested):

Anton-Kuchmasov (changes requested):
use constant for value of bite (50 is magic number)


f.e. const POWER_OF_BITE = 50;

emilionila (changes requested):
There are still some issues which haven't been changed as Roman requested

Anton-Kuchmasov (approved):
😎

Anton-Kuchmasov (approved):
Nice job!
Mistakes and issues were fixed

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
+const DAMAGE_AMOUNT = 50;
+const HEALTH_AMOUNT = 100;
```
BudnikOleksii, line 4: It's `DEFAULT_HEALTH` I think, as if you path another to constructor the health amount will be different

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
```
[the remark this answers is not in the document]
BudnikOleksii: > You already have a class named `Animal`, so to avoid confusion, it is better to name the function argument something else: aim, victim, target etc.

Make sense

BudnikOleksii (approved):
Good job 👍

PR #2281, opened 2023-11-12
done task Herbivores and Carnivores

Anton-Kuchmasov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name = '', hidden = false) {
+    super(name);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
MrHunterBoi, line 21: Hide function shouldn't toggle the prop. Set it to `true` only

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    this.health = HEALTH_DEFAULT;
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name = '', hidden = false) {
```
MrHunterBoi, line 15: Not sure if hidden property should be passed as a parameter, but okay

MrHunterBoi (changes requested):
Nice one

src/herbivoresAndCarnivores.js
Anton-Kuchmasov: ● Animal › Herbivore should have 'hide' method which **inverts** 'hidden' property
 
 from tests

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
+const HEALTH_DEFAULT = 100;
+
 class Animal {
-  // write your code here
+  constructor(name = '') {
+    this.health = HEALTH_DEFAULT;
```
artemforsoff, line 7: you can declare a field immediately after declaring a class:
```
class Animal {
  health = HEALTH_DEFAULT;
}  
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+  constructor(name = '') {
+    super(name);
+  }
+
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+
+      if (victim.health <= 0) {
+        Animal.alive = Animal.alive.filter(
```
artemforsoff, line 35: will be prettier:
```
Animal.alive = Animal.alive.filter(() => {
  return animal => animal.name !== victim.name;
});
```

artemforsoff: also, it's better to create a static method for the class that will delete the animal:
```
if (victim.health <= 0) {
  Animal.kill(victim);
}
```

artemforsoff (approved):
Keep up the wonderful work! 🚀
pay attention to the comments

PR #2282, opened 2023-11-13
classes added

yefimov77:
- [x] Done!

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super();
+  }
+
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && animal.hidden !== true) {
```
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```
I think there is no need to revert the condition here. 
It's recommended to check for truly values for better readability.
Also, you can shorten the condition with `not` for boolean

polosanya (approved):
Well done! 👍

PR #2289, opened 2023-11-14
Solution

BoblykNazar:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  // eslint-disable-next-line no-useless-constructor
+  constructor(name) {
```
polosanya, line 27: Don't disable the eslint, better fix its error, it is valid

polosanya (changes requested):
Almost done!

BudnikOleksii (approved):
Good job 👍

PR #2294, opened 2023-11-15
Vladyslav Slobodian

VladSlob:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
```
vlasiuksasha, line 19: The hide() method in the Herbivore class only sets the hidden property to true, but it should be able to toggle between true and false. You can fix this by setting this.hidden = !this.hidden.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+      if (animal.health <= 0) {
```
vlasiuksasha, line 28: You're checking if the animal's health is less than or equal to zero after decreasing it by 50. It's better to check if the animal's health is exactly zero to ensure that the animal is dead before removing it from the alive array.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+        Animal.alive = Animal.alive.filter(a => a !== animal);
```
vlasiuksasha, line 29: You're reassigning the entire Animal.alive array when removing a dead animal. It's more efficient to find the index of the animal in the array and remove it using the splice method.

vlasiuksasha (changes requested):
check the comments

4ksandrey (changes requested):
Revise your code and make fixes, on the "Checks" tab in your PR npm test shouldn't has any issues

4ksandrey (approved):
Accepted, well done

PR #2295, opened 2023-11-15
solution_v

mrsvolodya:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+
+  bite(herb) {
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
+    };
+
+    for (let i = 0; i < Animal.alive.length; i++) {
+      if (Animal.alive[i].health === 0) {
+        Animal.alive.splice(i, 1);
+      }
```
nazarmatsevych, line 42: Use filter method instead

nazarmatsevych (changes requested):
Good work, please see my comment below

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+  bite(herb) {
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
+    };
+
+    Animal.alive.filter((el, i) => {
+      if (el.health === 0) {
+        Animal.alive.splice(i, 1);
+      }
+    });
```
VitaliyBodnar, line 43: - wrong filter usage
- el it's a bad name because it's a element of animals list, so it's better to rename it to animal 
```suggestion
    Animal.alive = Animal.alive.filter(animal => animal.health);
```

VitaliyBodnar (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+
+  bite(herb) {
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
+    };
+
+    Animal.alive.filter((animal, index) => {
+      if (animal.health === 0) {
+        Animal.alive.splice(index, 1);
+      }
```
```suggestion
    Animal.alive = Animal.alive.filter(animal => animal.health > 0) 
```

Please learn how to use [filter ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)method

You don't need splice method here

nazarmatsevych (changes requested):

VitaliyBodnar (approved):

PR #2296, opened 2023-11-15
Solution

OstinM:


yefimov77 (approved):
Good!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
ArtemTeslenko, line 4: Create CONSTANT for health

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health, hidden);
+    this.hidden = hidden;
+  }
+
+  hide(isHidden) {
+    this.hidden = isHidden || true;
+  }
```
ArtemTeslenko, line 22: [Checklist](https://github.com/mate-academy/js_herbivores_and_carnivores/blob/master/checklist.md) p. 3

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health, hidden);
+    this.hidden = hidden;
+  }
```
ArtemTeslenko, line 18:
```suggestion
  constructor(name, health) {
    super(name, health);
    this.hidden = false;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.hidden = isHidden || true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    const BITE_DMG = 50;
+
+    if (target instanceof Herbivore && target.hidden === false) {
```
ArtemTeslenko, line 29: do not compare with false `target.hidden === false` -> `!target.hidden`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+      target.health -= BITE_DMG;
+    }
+
+    if (target.health === 0) {
```
ArtemTeslenko, line 33: the same `target.health === 0` -> `!target.health`

ArtemTeslenko (changes requested):
Good job) There are some comments, please fix 😉

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide(isHidden = true) {
+    this.hidden = isHidden;
```
artemforsoff, line 23: if the function has a name hide this function should just hide but not toggle
in your case I can call like this `hide(false)`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    const BITE_DMG = 50;
```
artemforsoff, line 29: shortening to consonants is a bad practice, some will understand right away, but others will not
so better rename it to `BITE_DAMAGE`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= BITE_DMG;
+    }
+
+    if (!target.health) {
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
artemforsoff, line 36: it's better to create a static method for the class that will delete the animal:

```
if (target.health <= 0) {
  Animal.kill(victim);
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
+const ANIMAL_HELTH = 100;
```
artemforsoff, line 3: there is a mistake in the word _health_

artemforsoff (changes requested):
overall great job 🔥
fix the method and variable names

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
ArtemTeslenko, line 28: I agree with your logic, it is better to toggle hidden in such cases, but here in accordance with [checklist](https://github.com/mate-academy/js_herbivores_and_carnivores/blob/master/checklist.md) we should just hide, `this.hidden = true`, method name should reflect it's logic, `toggleHidden` for example can contain your logic, but here we should not change method's name

ArtemTeslenko (approved):
Good job) I am approving it, but see my comment 😉

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.health = ANIMAL_HEALTH;
+    this.name = name;
+
+    Animal.alive.push(this);
+  }
+
+  static killedAnimal(victim) {
```
artemforsoff, line 13: in your case, the name is a noun, the function should answer the question “what to do?”
so you should rename to `kill` or `killAnimal`

PR #2297, opened 2023-11-15
initial commit

Barosz30:
Animal.alive = [];

Wanted to make that 

static alive = [] ; but lint hated the idea


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,43 @@
...
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(enemy) {
+    if (enemy.hidden === false && enemy instanceof Herbivore) {
+      enemy.health -= 50;
+      if (enemy.health === 0) {
+        const index = Animal.alive.indexOf(enemy);
+        Animal.alive.splice(index, 1);
```
zaura333, line 38: Might be better to use filter there (as suggested in the checklist)

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,43 @@
...
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    if (this.hidden === false) {
+      this.hidden = true;
+    } else {
+      this.hidden = false;
```
zaura333, line 26: Just a thought: I used
`this.hidden = !this.hidden;`
there, saves us some ifs

zaura333 (approved):
Neat

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,39 @@
...
   // write your code here
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
```
mvjl000, line 13: This should be static but as You noticed, it's causing some linter errors :/ let's leave it like this then

mvjl000 (approved):
Nice!

PR #2298, opened 2023-11-15
Solution

zaura333:


Barosz30 (approved):
this.hidden = !this.hidden;

I'm gonna steal that above for bolean xD 
Great code!

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,38 @@
...
 class Animal {
   // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
```
mvjl000, line 12: This should be a static property

mvjl000 (changes requested):
Well done! I would change just one thing here

src/herbivoresAndCarnivores.js
mvjl000: If making it static causes some linter errors you don't have to change it ;)
Sorry, I've just realized ;)

zaura333: I managed to change linter options

zaura333: Nevermind, not really. Tests in VS code passed, not on github however, I've changed it back

mvjl000 (approved):

PR #2301, opened 2023-11-15
solution

MarcinLigmann:
Ignore the index.js file
It was only for testing purposes

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && beast.hidden !== true) {
+      beast.health -= 50;
+    }
+
+    if (beast.health <= 0) {
+      const index = Animal.alive.indexOf(beast);
+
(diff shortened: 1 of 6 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
```
jacekkosiba, line 11: Alive should be a static property of Animal class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && beast.hidden !== true) {
```
jacekkosiba, line 26: This is correct, but you could shorten
`beast.hidden !== true`
to
`!beast.hidden`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+      beast.health -= 50;
+    }
+
+    if (beast.health <= 0) {
+      const index = Animal.alive.indexOf(beast);
```
jacekkosiba, line 31: Instead of using `.indexOf()` and `.splice()`, you can just use the `.filter()` method

jacekkosiba (changes requested):

jacekkosiba (approved):

jacekkosiba (approved):

src/herbivoresAndCarnivores.js
mbruhler: and it is that way :)

src/herbivoresAndCarnivores.js
mbruhler: true

src/herbivoresAndCarnivores.js
mbruhler: yes, that's right

mbruhler (approved):
Approved!

PR #2302, opened 2023-11-15
Solution

Patrykmclaren14:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  // eslint-disable-next-line no-useless-constructor
+  constructor(name) {
+    super(name);
```
MarcinLigmann, line 28: I think super() is unnecessary here.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
MarcinLigmann, line 6: In my opinion, health should be introduced as a function parameter.
I'm talking about constructor(name, health = 100) {...

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    super(name);
+  }
+
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden) {
+      return;
+    }
+
+    if (!(target instanceof Carnivore)) {
+      target.health -= 50;
```
MarcinLigmann, line 37: The first two conditional statements in the bite() method can be turned into a single IF

MarcinLigmann (approved):
I think these are cosmetic comments, the code works and otherwise looks good. I approve :)

src/herbivoresAndCarnivores.js
mbruhler: I would argue, it's always 100 so both ways are correct imo.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = undefined;
```
mbruhler, line 7: why undefined?

mbruhler: and why you introduce hidden here? It's not a property of Animal in overall, just Herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  // eslint-disable-next-line no-useless-constructor
+  constructor(name) {
+    super(name);
+  }
+
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden) {
```
mbruhler, line 32: you can just check for that, if it's Carnivore then it's not Herbivore. Add the decreasing health logic here.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+      return;
+    }
+
+    if (!(target instanceof Carnivore)) {
+      target.health -= 50;
+
+      if (target.health <= 0) {
```
mbruhler, line 39: You can create a Animal method named `kill` and move this logic there.

mbruhler (changes requested):
Changes are needed, see the comments below

PR #2304, opened 2023-11-15
Finished

Fluebubble:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,47 @@
...
 class Carnivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (!animal.hidden && !(animal instanceof Carnivore)) {
+      animal.health -= 50;
```
yevtok, line 37: Avoid using "magic values".

yevtok (changes requested):
Good, but need changes

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,49 @@
 
 class Animal {
   // write your code here
+  static alive = [];
```
polosanya, line 5: You can try to create it outside of the class to avoid eslint error
`Animal.alive = []`

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,49 @@
...
+  }
 }
 
 class Herbivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+
+    this.hidden = false;
+    Animal.alive.push(this);
```
polosanya, line 20: DRY.
Move it to the constructor of the animal class, so all other animals will inherit it

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,49 @@
...
 
 class Carnivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (!animal.hidden && !(animal instanceof Carnivore)) {
```
polosanya, line 36:
```suggestion
  bite(animal) {
    if (!animal.hidden && animal instanceof Herbivore) {
```
There is no need to overcomplicate the condition

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,49 @@
...
+      const BITE_FORCE = 50;
```
polosanya, line 37: I think it's better to move it outside of the method scope

polosanya (changes requested):
Good job! Let's try to improve your code a bit)

src/herbivoresAndCarnivores.js
Fluebubble: The task: 
"Create a `Herbivore` class.
Herbivore has a method of `hide`, which changes the `hidden` property of the beast and helps to hide from carnivores."

ArtemTeslenko (approved):
Good job)

PR #2306, opened 2023-11-15
herbivoresAndCarnivores task solution

VeDono:
I don't understand how to create 'Animal.alive' throuh:
`class Animal {
  static alive = [];
  constructor(name) {
    this.name = name;
    this.health = HEALTH;
  }
}`
cause linter each time throw error when i trying write that.

Also i don't understant how to avoid error from linter when i write `static DAMAGE_PER_BITE = 500` in exactly class, so i do that not throuh `static` outside of classes.
I spend all evening for undestanding how resolve that throuh `static` and get fail, do you can tell me where my problem ?

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,50 @@
 'use strict';
 
+const HEALTH = 100;
+const DAMAGE_PER_BITE = 50;
+
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = HEALTH;
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(target) {
+    if (
+      target.health > 0
+      && target instanceof Herbivore
+      && !target.hidden
+    ) {
+      target.health -= DAMAGE_PER_BITE;
+
+      if (target.health === 0 || target.health < 0) {
+        Animal.alive = Animal.alive.filter((currentAnimal) => (
+          currentAnimal !== target
+        ));
+      }
+    }
+  }
 }
 
 module.exports = {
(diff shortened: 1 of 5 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
polosanya, line 19: DRY.
Move it to the Animal container, so all other classes have it too

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
polosanya, line 23: image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(target) {
+    if (
+      target.health > 0
+      && target instanceof Herbivore
+      && !target.hidden
```
polosanya, line 37: This condition looks overcomplicated. There is no need to check target's health as it will be deleted from alive array if the health is 0

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+    ) {
+      target.health -= DAMAGE_PER_BITE;
+
+      if (target.health === 0 || target.health < 0) {
```
polosanya, line 41:
```suggestion
      if (target.health <= 0) {
```
It can be simplified

polosanya (changes requested):
Nice! Let's try to improve your code a bit)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+const HEALTH = 100;
+const DAMAGE_PER_BITE = 50;
+
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = HEALTH;
+    Animal.alive.push(this);
+  }
```
roman-mirzoian, line 11: To answer your question from the comment: in this case, you don't need to set a default value via a constant, because that value itself will tell you what it means and it won't be considered a magic value:
```suggestion
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;
    Animal.alive.push(this);
  }
```
You also need to be able to pass a different value for the `health` field to the constructor, so it is important to use this option

roman-mirzoian (approved):
Well done!
To answer your question about static: tasks in mate has a strict linter that expects you to go without statics, so you did the right thing 🔥

PR #2307, opened 2023-11-15
solution

Timi2424:


adrian200002 (approved):
It seems to be correct for me. The only thing i would maybe change is to shorten if statement in line 29, but i'm not sure so i'm approving.

mbruhler (approved):
Well done!
Approved!

PR #2308, opened 2023-11-16
solution

adrian200002:


DanyloSemenov (changes requested):
1) The Animal constructor sets health to 100 by default, but it's overridden by the parameter value. It should prioritize the parameter value or default to 100.

2) The Herbivore class lacks the hide method mentioned in the task; it needs to be added.

3) Animal.alive should store objects with specified properties (name, health, hidden). Currently, only animal instances are pushed; create objects with these properties and push them.

mbruhler (approved):
Well done! Approved!

PR #2310, opened 2023-11-16
animals are good

Mantserov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden !== true) {
+      victim.health -= 50;
```
ilyapgitt, line 30: I understand now, that 50 is magic number, maybe, but I'm not sure. I made this mistake too

ilyapgitt (changes requested):
Awesome!!! But you shouldn`t have magic number)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden !== true) {
```
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
```
This condition can be simplified

polosanya (approved):
Good job 👍

PR #2311, opened 2023-11-16
solution

DanyloSemenov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
```
kate-prakofyeva, line 30: I would make a variable inside the method 
const DAMAGE = 50;

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
 'use strict';
-
 class Animal {
-  // write your code here
+  constructor(name, health = Animal.defaultHealth) {
```
kate-prakofyeva, line 3: health = 100

kate-prakofyeva (approved):
It's hard to review because I'm not sure I did the right thing myself.
I definitely don’t like the `Animal.alive = [];
Animal.defaultHealth = 100;`
, but I don’t know if it’s possible to do it in this way?

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+    this.name = name;
```
mbruhler, line 4: You don't need health as a constructor argument, as it's always 100. Move it as 
`this.health = 100;` to the constructor body

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+    Animal.alive.push(this);
+  }
+
+  static kill(victim) {
+    Animal.alive = Animal.alive.filter(animal => animal !== victim);
+  }
 }
 
+Animal.alive = [];
+Animal.defaultHealth = 100;
```
mbruhler, line 15: this is a proper way to mutate Animal class by adding static prop, but not actually needed. See the above.

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
```
mbruhler, line 19: If you get rid of the health argument in the Animal constructor, this would be just `super(name)`

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,44 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= Carnivore.damage;
+
+      if (target.health <= 0) {
+        Animal.kill(target);
```
mbruhler, line 34: Good job here on moving the killing logic to Animal class.

mbruhler (changes requested):
Hi,
Some changes are needed, see the comments below.

mbruhler (approved):
Good job, approved!

PR #2312, opened 2023-11-16
my solution

YehorHeneralov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
```
roman-mirzoian, line 27: Avoid using "magic values" (what is 50?)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    }
+
+    if (target.health === 0) {
```
roman-mirzoian, line 30: If the `health` value is less than zero, this condition will not be fulfilled and there will be extra elements with negative health values in the `alive` array

roman-mirzoian (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    const hitted = 50;
```
polosanya, line 26: The name is a bit confusing. Better name it `damage` or smth like that

polosanya (approved):
Good job!

PR #2313, opened 2023-11-16
Solution

YuriySobakar:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && !beast.hidden) {
+      beast.health -= 50;
```
roman-mirzoian, line 26: Avoid using "magic values" (what is 50?)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    }
+
+    if (beast.health === 0) {
```
roman-mirzoian, line 29: If `beast.health` is less than zero, this condition will not work and the array will contain an object with a negative health value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+      Animal.alive = Animal.alive.filter(animal => animal.health !== 0);
```
roman-mirzoian, line 30: Don't forget about type conversions
```suggestion
      Animal.alive = Animal.alive.filter(animal => animal.health);
```

roman-mirzoian (changes requested):

PonomarchukLana (approved):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
BudnikOleksii, line 6:
```suggestion
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;
```
Health is 100 by default.
But we can pass to the constructor another value if needed

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    const BITE = 50;
```
BudnikOleksii, line 25: This constant created every time when `bite` method is called, move this constant outside of class

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions to improve your solution

BudnikOleksii (approved):
🚀

PR #2316, opened 2023-11-16
Solution

ilyapgitt:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
-  // write your code here
+  constructor(name = '') {
+    super(name);
+  }
+  bite(production) {
+    if (production instanceof Herbivore && !production.hidden) {
+      production.health -= DAMAGE;
+    }
+
+    if (production.health === 0) {
```
roman-mirzoian, line 35: If the `health` value is less than zero, then the condition will not be fulfilled and there will be an object with a negative health value in the array of living animals

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      Animal.alive = Animal.alive.filter(person => person.health !== 0);
```
roman-mirzoian, line 36:
```suggestion
      Animal.alive = Animal.alive.filter(person => person.health);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name = '') {
+    super(name);
+  }
+  bite(production) {
```
roman-mirzoian, line 30: What does you mean by using `production`? 
You can choose more predictable name: aim, victim etc.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
+const FULL_HEALTH = 100;
+const DAMAGE = 50;
+
 class Animal {
-  // write your code here
+  constructor(name = '') {
```
roman-mirzoian, line 7: `name` should have no default value

roman-mirzoian (changes requested):

YehorHeneralov (approved):
good job)

PR #2317, opened 2023-11-17
solution

bertsu02:


Timi2424 (approved):
Great overall. If you want to resolve conflict with static variable and eslist, I suggest reading this https://github.com/mate-academy/js_herbivores_and_carnivores/issues/952

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
```
JoyCoffeeAddict, line 28: actually, `herb` is a wrong argument name
1) you can (but shouldn't) pass a carnivore to this function 
2) `herb` has another meaning as a word. it's like shortening carpet to a car.

JoyCoffeeAddict (changes requested):

src/herbivoresAndCarnivores.js
bertsu02: So just change argument name and its all good yeah ?

PR #2319, opened 2023-11-18
add task solution

heorhiig:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      const index = Animal.alive.indexOf(animal);
+
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
+      }
```
VitaliyBodnar, line 31: It's better to use method filter instead of indexOf + splice

VitaliyBodnar (changes requested):

etojeDenys (approved):
good job

PR #2320, opened 2023-11-18
Develop1

vinogradova8:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor (name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = this.hidden !== true;
```
VitaliyBodnar, line 20: can be less
```suggestion
    this.hidden = true;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health === 0) {
+      const index = Animal.alive.indexOf(herbivore);
+
+      Animal.alive.splice(index, 1);
```
VitaliyBodnar, line 33: It's better to use method filter instead of indexOf + splice

VitaliyBodnar (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
```
VitaliyBodnar, line 26:
```suggestion
    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```

VitaliyBodnar (approved):

PR #2322, opened 2023-11-19
Solution

mosowiecka:


bertsu02 (approved):
I think it looks fine good job!

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,44 @@
...
+    if (animal.hidden !== true && animal instanceof Herbivore) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        const index = Animal.alive.indexOf(animal);
+
+        if (index !== -1) {
+          Animal.alive.splice(index, 1);
+        }
+      }
```
JoyCoffeeAddict, line 40: it would be nice, if you would `console.error` when e.g. we want to bite hidden animal. 
to make it easier write your `if statements` with [guard clauses](https://learningactors.com/javascript-guard-clauses-how-you-can-refactor-conditional-logic/) 

so that you can check
```
if(!something) {
//log error and return
}
```

JoyCoffeeAddict (changes requested):

mbruhler (changes requested):
Please add changes from @JoyCoffeeAddict with guard clauses, your if statements are too deeply nested and it's hard to read.

Also resolve comments when fixing your code

mbruhler (approved):
Approved

PR #2323, opened 2023-11-20
Herbivore and Carnivores

PonomarchukLana:


YuriySobakar (approved):
Good job)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
+
 class Animal {
-  // write your code here
+  constructor(name, health = DEFAULT_HEALTH) {
+    this.name = name;
+    this.health = health;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
+    }
```
BudnikOleksii, line 13: You don't need to do it every time when an instance of an animal is created, only once. So you can move `Animal.alive = []` outside of class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
+    Animal.alive.push(this);
+  }
+
+  die() {
+    if (Animal.alive) {
+      const index = Animal.alive.indexOf(this);
+
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
+      }
```
BudnikOleksii, line 24: [CODE PERFOMANCE] - Try to avoid nesting loops (indexOf + splice), use filter instead

BudnikOleksii (approved):
Good work 👍 

PR #2324, opened 2023-11-20
Solution

szymanski-marcin:


web3-sorcerer (changes requested):
In your Carnivore class you don't have to create constructor and pass super(name) if you are not using it anyway.

mbruhler (approved):
Well done! Approved

PR #2326, opened 2023-11-20
task done

olegnovikov10:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
Anton-Kuchmasov, line 6: don't use `magic numbers` in your code 


it means that ALL literals (etc. numbers) should be initialized as global constants 

f.e.:
const HEALTH_DEFAULT = 100;

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
```
Anton-Kuchmasov, line 19: method hide() should invert hidden prop

f.e. if hidden was true, method hide() should turn in to false
if it was false, it should be true

Anton-Kuchmasov (changes requested):
nice job!
but need some fixes

src/herbivoresAndCarnivores.js
BudnikOleksii: > method hide() should invert hidden prop
> 
> f.e. if hidden was true, method hide() should turn in to false if it was false, it should be true

Actually, it shouldn't toggle as the function is called `hide`
[FUNCTION NAMING] - if the function has a name hide this function should just hide but not toggle

BudnikOleksii (approved):
Good work 👍

Anton-Kuchmasov (approved):
Perfect!

PR #2328, opened 2023-11-21
Solution

web3-sorcerer:


WiktoriaBielec (approved):
All good

mbruhler (approved):
Well done.

One note: it's hard to read this code
```
if (!Array.isArray(Animal.alive)) {
      Animal.alive = [];
}
```
In here it would be better to mutate the class with `Animal.alive` somwhere outside the Animal class.
In normal environment, declaring data member as `static alive = []` would be enough.

Approved.

PR #2330, opened 2023-11-21
Solution

WiktoriaBielec:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,45 @@
...
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      const index = Animal.alive.indexOf(herbivore);
+
+      Animal.alive.splice(index, 1);
+    }
```
MarcinLigmann, line 42: Instead of using .indexOf() and .splice(), you can just use the .filter() method

MarcinLigmann (approved):
The only thing I can fault is a bite() method described below.
Apart from that, the code works and looks good :)

src/herbivoresAndCarnivores.js
mbruhler: Depends...

mbruhler (approved):
Approved,

NOTE: for the future, please clear the code from comments

PR #2331, opened 2023-11-21
Solution

michal-reszka:


RadoslawBielski (approved):
I think you can do this better but is okey

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
mbruhler, line 14: hidden is not needed here, because when creating herbivore, he is not hidden by default

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(brute) {
```
mbruhler, line 25: Please use more generic and self explanatory argument names, 'animal' or 'herbivore' fits much better here.

mbruhler (changes requested):
Hi,
Your code needs minor changes and we will be good to go

mvjl000 (approved):
Alles gut!

PR #2332, opened 2023-11-21
Develop

RadoslawBielski:


adrian200002 (approved):
Looks fine for me.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(lion) {
```
mbruhler, line 25: why it's lion here :)

the argument name should name all the herbivores that are passed here.
It should be 'animal', 'herbivore' or something more self-explanatory

mbruhler (approved):
Code works well but your bite method argument is named in a bad way. Please use more generic and self-explanatory names for variables and arguments

PR #2334, opened 2023-11-22
add task soulution

Kozubowicz:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  };
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(name) {
```
anastasiiavorobiova, line 25:
```suggestion
  bite(animal) {
```
There is an animal object here, so it's better to rename it

anastasiiavorobiova (changes requested):
Great work! Added a few comments to fix

volodymyr-soltys97 (approved):
Well done 🔥

PR #2335, opened 2023-11-23
Herbivores and Сarnivores

roman-tarasov90:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super();
+  }
```
etojeDenys, line 27: constructor is redundant when you wrote there only the super

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  bite(target) {
+    if (target.hidden === false) {
```
etojeDenys, line 30: you need to add the check if we try to bite the Herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+      target.health = target.health - 50;
+    }
+
+    if (target.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(target), 1)
```
etojeDenys, line 35: you can just use one the filter method instead of these two

etojeDenys (changes requested):

DarkMistyRoom (approved):
Well done

PR #2340, opened 2023-11-24
Solution

ghost:


OstTomash:
I don`t know why all checks have failed

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
VitaliyBodnar, line 33: it's better to use method filter instead of splice with indexOf

VitaliyBodnar (changes requested):
Seems like there are some issues with github and tests. Don't worry about it.

src/herbivoresAndCarnivores.js
OstTomash: Thanks!
Just added this changes.

nazarmatsevych (approved):
👍

PR #2341, opened 2023-11-27
Write solution

GlibVK:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target.hasOwnProperty('hidden')) {
+      if (!target.hidden) {
+        target.health -= 50;
+      }
+    }
```
nazarmatsevych, line 32:
```suggestion
  bite(target) {
    if (targer instanceof Herbivore && !target.hidden) {
      target.health -= 50;
    }
```

Same but a bit shorter

nazarmatsevych (changes requested):
Great work, please check comment below

VitaliyBodnar (approved):
Good job!

PR #2351, opened 2023-12-02
first

ColumbusDeveloper:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
+  }
+  hide(val) {
+    if (val === undefined) {
+      this.hidden = true;
+    } else if (val === true || false) {
+      this.hidden = val;
+    }
```
nazarmatsevych, line 23: This method should not toggle hidden value and only change it to true

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+  }
+  bite(val) {
+    if (val instanceof Herbivore) {
+      if (val.health === 50 && !val.hidden === true) {
+        val.health = 0;
+
+        const herbDelIndex = Animal.alive.indexOf(val);
+
+        Animal.alive.splice(herbDelIndex, 1);
+      }
```
nazarmatsevych, line 40:
```suggestion
    if (val instanceof Herbivore && !val.hidden) {
        val.health -= 50;
}
        const herbDelIndex = Animal.alive.indexOf(val);

        Animal.alive.splice(herbDelIndex, 1);
      }
```

And use filter method instead of indexOf and splice

nazarmatsevych (changes requested):
Good work, please check comments below

PR #2355, opened 2023-12-03
add task solution

dimakrugly:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,58 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static removeFromAlive(animal) {
+    const index = Animal.alive.indexOf(animal);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
nazarmatsevych, line 15: Use filter method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,58 @@
...
+  bite(victim) {
+    if (victim instanceof Carnivore) {
+      return;
+    }
+
+    if (victim instanceof Herbivore && victim.hidden) {
+      return;
+    }
+
+    victim.health -= 50;
```
nazarmatsevych, line 50:
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
      victim.health -= 50;
    }
```

we can simplify this condition

nazarmatsevych (changes requested):
Good work, please check comments

nazarmatsevych (approved):
Great work!

PR #2357, opened 2023-12-04
Solution

wsadocha:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,49 @@
...
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === undefined || herbivore.hidden) {
+      return;
+    }
+
+    herbivore.health -= 50;
```
```suggestion
    if (herbivore instanceof Herbivore && !herbivore .hidden) {
     herbivore.health -= 50;
    }
```

We can simplify it a bit

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,49 @@
...
+
+    Animal.addAnimal(this);
+  }
+
+  static addAnimal(animal) {
+    this.alive.push(animal);
+  }
+
+  static removeAnimal(animal) {
+    delete this.alive[this.alive.indexOf(animal)];
```
nazarmatsevych, line 19: Don't manipulate with array in this way,  use filter method instead

nazarmatsevych (changes requested):

nazarmatsevych (approved):
Well done 👍

PR #2363, opened 2023-12-06
Animals - bite Zebra + atatic alive

baraban2003:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+    this.name = name;
+    this.health = health;
+
+    if (!Animal.alive.includes(this)) {
+      Animal.alive.push(this);
+    }
+  }
+
+  static removeFromAlive(animal) {
+    Animal.alive = Animal.alive.filter((a) => a !== animal);
```
l4st1m0za, line 14: Instead of reassigning the filtered array to Animal.alive, you can use Array.prototype.splice() to remove the animal from the original array. This will eliminate the need to create a new array each time an animal dies.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    if (!Animal.alive.includes(this)) {
+      Animal.alive.push(this);
```
l4st1m0za, line 9: There's no need to check if instance is already in the 'alive' array. Each animal is created only once and immediately added to the array.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (
+      animal instanceof Carnivore
```
l4st1m0za, line 38: The conditional check in the 'bite' method could be simplified by checking only if the animal is hidden. If 'hidden' is not a property of the animal (i.e., it is a Carnivore), it will be 'undefined' and thus falsy.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = true;
```
l4st1m0za, line 27: It would be better to not hard-code the 'hidden' property to always be true, but toggle it instead. This way, the 'hide' method can be used to both hide and unhide the herbivore.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+  static removeFromAlive(animal) {
+    Animal.alive = Animal.alive.filter((a) => a !== animal);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
l4st1m0za, line 21: This constructor can be simplified. As 'hidden' property is specific to Herbivore and has an initial value, we don't need to pass it as a parameter.

l4st1m0za (changes requested):
GJ! See the comments

src/herbivoresAndCarnivores.js
baraban2003: Agree, Tnx!

src/herbivoresAndCarnivores.js
baraban2003: I had some issues with Lint and it was one of the attempts

anastasiia-tilikina (approved):
Well done!

PR #2365, opened 2023-12-07
Solution

Miishca:


nazarmatsevych (changes requested):

PR #2371, opened 2023-12-09
Create Animals clases

VoloshynV:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
nazarmatsevych, line 22:
```suggestion
    this.hidden = true;
```

this method should only change hidden value to true and not toggle it

nazarmatsevych (approved):
Great work, left minor comment

PR #2375, opened 2023-12-10
Js_Herb_and_Carn try#1 linter does not work with static property

VolodymyrLiubyn:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,49 @@
...
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    if (this.hidden === true) {
+      this.hidden = false;
+    };
+    this.hidden = true;
+  }
```
```suggestion
    this.hidden = true;
  }
```

this method should not toggle hidden value and only change it to true

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,49 @@
...
+
+  bite(prey) {
+    if (prey.hidden === false) {
+      prey.health -= 50;
+
+      if (prey.health === 0) {
+        for (let i = 0; i < Animal.alive.length; i++) {
+          if (Animal.alive[i].name === prey.name) {
+            Animal.alive.splice(i, 1);
+          }
```
nazarmatsevych, line 43: Looks too complicated, use filter method instead

`Animal.alive = Animal.alive.filter(....`

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,49 @@
...
 }
 
 class Carnivore extends Animal {
   // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(prey) {
+    if (prey.hidden === false) {
```
```suggestion
    if (prey instanceof Herbivore && !prey.hidden) {
```

nazarmatsevych (changes requested):

src/herbivoresAndCarnivores.js
VolodymyrLiubyn: I got it, I'm fixing it

nazarmatsevych (approved):
👍

PR #2377, opened 2023-12-10
solution

MykhailoLoniak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+
+    return this;
+  }
+
+  checkIfDead() {
+    if (this.health <= 0) {
+      const index = Animal.alive.indexOf(this);
+
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
```
nazarmatsevych, line 44: Use filter method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
-class Carnivore extends Animal {
-  // write your code here
+class Herbivore extends Animal {
+  constructor(name, hidden = false) {
+    super(100, name);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
```suggestion
    this.hidden = true;
```

Hide method should not toggle hidden value and only change it to true

nazarmatsevych (changes requested):
Almost done, please check comments below

anastasiiavorobiova (approved):
Great work!

PR #2380, opened 2023-12-13
solution

iamgranovskaya:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+
+    Animal.alive.push(this);
```
VitaliyBodnar, line 17: Animal.alive.push(this); is duplicated in Herbivore and Carnivore, so make sense to move it to the class Animal 
(to constructor)

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #2384, opened 2023-12-14
new commit

Max-just:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,55 @@
...
 
 class Animal {
-  // write your code here
+  constructor(health = 100, name) {
+    this.health = health;
+    this.name = name;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
+    }
```
etojeDenys, line 10: you must declare the static array outside of the class
```
class Animal {...}

Animal.alive = []

...
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,55 @@
...
+    this.health -= amount;
+
+    if (this.health <= 0) {
+      this.removeDeadAnimal(this);
+    }
+  }
+
+  removeDeadAnimal(animal) {
+    Animal.alive = Animal.alive.filter(a => a !== animal);
+  }
```
etojeDenys, line 39:
```suggestion
    if (this.health <= 0) {
      Animal.removeDeadAnimal(this);
    }
  }
```
why do you need to create the another method if you have the static one
image

etojeDenys (changes requested):

Esceype (approved):
GJ!

PR #2394, opened 2023-12-21
add task colution

VazilX:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+
+    this.hidden = false;
+  }
+
+  hide() {
+    if (this.hidden) {
+      this.hidden = false;
+    } else {
+      this.hidden = true;
+    }
```
```suggestion
  this.hidden = true;
```

this method should not toggle hidden value and only change it to true

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hasOwnProperty('hidden') && animal.hidden === false) {
```
nazarmatsevych, line 37:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static dying() {
+    const dyIndex = this.alive.findIndex(animal => animal.health <= 0);
+
+    Animal.alive.splice(dyIndex, 1);
```
nazarmatsevych, line 7: Use filter method instead

nazarmatsevych (changes requested):
Good work, please check the comments below

nazarmatsevych (approved):
👍

PR #2397, opened 2023-12-25
solution

pivkopa:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
```
nabokok, line 9: All alive animals should be in the static Animal.alive array.

nabokok (changes requested):
Good job! Please, look at the comment

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(herbivore) {
+    if (herbivore.hidden === false && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
```
DanilWeda, line 33: Magic numbers, use constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    }
+
+    if (herbivore.health === 0) {
```
DanilWeda, line 36: Magic numbers, use constant

DanilWeda (changes requested):
All good! 
Check some comments, and need to figure it out with **static** in your case.

olya-shyian (changes requested):
Great solution!✨

DanilWeda (approved):
Great. 
Thanks for the update deeps, now static is working correctly! 
Approved with comments ( numbers need to be in constants).

PR #2401, opened 2023-12-27
Solution

isaeev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
nazarmatsevych, line 4: Linter does not like static properties. You can create this property like below:

`Animal.alive = []`

this property must now be moved outside of the Animal class

nazarmatsevych (changes requested):
Great work, please check the comment below to fix tests

lerastarynets (approved):
lgtm 🚀

PR #2402, opened 2023-12-27
Solution

deonara:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+
+      if (target.health <= 0) {
+        const index = Animal.alive.indexOf(target);
+
+        if (index !== -1) {
+          Animal.alive.splice(index, 1);
+        }
```
etojeDenys, line 33: you can just use the filter here, instead of the indexOf and splice

etojeDenys (changes requested):

nazarmatsevych (approved):
🔥

PR #2407, opened 2023-12-27
Solution

Yur1yBubl1k:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
```
ValeriaMyronets, line 10: fields should be moved to top in class Animal

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(victim) {
+    if (victim instanceof Herbivore
+      && victim.hidden === false) {
```
ValeriaMyronets, line 30: field hidden type of Boolean, use this without ' === false'

```javascript
if (victim instanceof Herbivore && !victim.hidden)
```

ValeriaMyronets (changes requested):
Good job, but needs a few changes✨🥰

PR #2408, opened 2023-12-28
add solution

ValeriaMyronets:


yaneznayou (approved):
Good Job But it seems to me that the tests failed

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive.splice(Animal.alive.indexOf(animal), 1);
+      }
```
olya-shyian, line 36: you can create a basic method for this like `killAnimal`

olya-shyian (changes requested):
Good job!✨

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        this.killAnimal(animal);
+      }
+    }
+  }
+
+  killAnimal(animal) {
+    Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
DanilWeda, line 41: It is necessary to make sure that the element is found.

Checklist:
When you are using indexOf and splice combo - you should check if element was found before deleting element.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
```
DanilWeda, line 32: all numbers can be put in constant (for best readability)

DanilWeda (changes requested):
All good! 
Please, fix minor issues. 
Many thanks!

PR #2411, opened 2023-12-28
add solution

kostrubitskii:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
vSainiuk, line 4: I think there is no need to add a parameter as health because it is a default property.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
```
vSainiuk, line 13: There is no need to duplicate 'constructor' because when we add some arguments they automatically add in constructor in an inheritable class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore)
+    && (!animal.hidden || animal.hidden === undefined)
```
vSainiuk, line 26: The exclanation mark before condition already means that you expect a boolean value of not true, so second condition no longer makes sense)

vSainiuk (approved):
Well done!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
```
DanilWeda, line 10: Alive needs to be static in Animal.
If the linter drops an error try updating deeps.

src/herbivoresAndCarnivores.js
DanilWeda: Correct

src/herbivoresAndCarnivores.js
DanilWeda: Correct

src/herbivoresAndCarnivores.js
DanilWeda: optional

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore)
+    && (!animal.hidden || animal.hidden === undefined)
+    ) {
+      animal.health -= 50;
```
DanilWeda, line 28: all numbers can be put in constant (for best readability)

DanilWeda (changes requested):
Great! 
Check some improvements. 
Many thanks for your request!

DanilWeda (changes requested):
Please correct the comments I left.
( Maybe a platform bug)

src/herbivoresAndCarnivores.js
kostrubitskii: I have problem and i dont know hot to fix it... 
What i need to do ?
Дште

olya-shyian (changes requested):
Good!) Please fix the previous comments

src/herbivoresAndCarnivores.js
DanilWeda: or you can add https://www.npmjs.com/package/@babel/eslint-parser in your project, and add key parser: nameOfLibrary in eslinter config.

DanilWeda (approved):
Great! 
Approved.

PR #2414, opened 2023-12-28
solution

vSainiuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  static defaultDamage = 50;
+
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= Carnivore.defaultDamage;
+
+      if (victim.health === 0) {
+        const indexVictim = Animal.alive.indexOf(victim);
+
+        Animal.alive.splice(indexVictim, 1);
```
vpashko, line 33: It works now, but you can follow first rule from checklist : 
When you are using indexOf and splice combo - you should check if element was found before deleting element.

So you can fix it before the mentor review.

vpashko (approved):
Great job! But small fixes can be added.

src/herbivoresAndCarnivores.js
DanilWeda: Correct

DanilWeda (changes requested):
All good! Great! 
Please, add a condition for finding the element.
And check test suites!

PR #2416, opened 2023-12-28
First solution

Nikita211112:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+      this.die();
+    }
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
nazarmatsevych, line 23: Use filter method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden === false) {
```
```suggestion
    if (target instanceof Herbivore && !target.hidden) {
```

nazarmatsevych (changes requested):
Almost done, please check the comments below :)

nazarmatsevych (approved):
Awesome 🔥

PR #2417, opened 2023-12-28
add solution

ssaviak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    super(name);
+    Animal.alive.push(this);
+  }
+
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```
EdenParadise, line 35: Consider creating methods for both, damaging herbivore and removing herbivore from array, so code can be reusable for different kid of animals

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
```
EdenParadise, line 27: There is no need to use ` Animal.alive.push(this);` since you already have that in constructor of class that you extend

EdenParadise (changes requested):
Great job overall <3 , need to fix some details

PR #2418, opened 2023-12-28
task solved

vpashko:
- [x] When you are using indexOf and splice combo - you should check if element was found before deleting element.
- [x] Try to avoid nesting loops (indexOf + splice), use filter instead
- [x]  if the function has a name hide this function should just hide but not toggle

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
   extends: '@mate-academy/eslint-config',
-  parser: "babel-eslint",
+  "parser": "@babel/eslint-parser",
   env: {
     jest: true
   },
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -30,7 +30,9 @@ class Carnivore extends Animal {
     if (victim.health === 0) {
       const index = Animal.alive.indexOf(victim);
 
-      Animal.alive.splice(index, 1);
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
+      }
     }
   }
 }
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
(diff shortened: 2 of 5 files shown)


.eslintrc.js
artem8746, line 1: do not add extra files, add just `src`

artem8746, line 1: do not add extra files, add just `src`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= 50;
+    }
+
+    if (victim.health === 0) {
+      const index = Animal.alive.indexOf(victim);
```
artem8746, line 31: [checklist] Try to avoid nesting loops (indexOf + splice), use filter instead

artem8746 (approved):
Good job!!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
+      }
```
olya-shyian, line 35: create a static method (in class Animal) for updating `alive` and use here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= 50;
```
olya-shyian, line 27: create a variable for number 50 (in order to understand what this number means)

olya-shyian (changes requested):
Good job!👍

VasiliyYarosh (approved):
Good job

YegorKorenkov (approved):
Good job!

PR #2419, opened 2023-12-28
Done! (try1)

EdenParadise:
Removing animals from array was kind of a challange because i wasnt sure what to choose, delete them in method bite() or delete them in other methods

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,52 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
+Animal.removeFromAlive = function(target) {
+  const index = this.alive.indexOf(target);
+
+  if (index !== -1) {
+    this.alive.splice(index, 1);
+  }
+};
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
+
+  decreaseHealth(amount) {
+    this.health -= amount;
+
+    if (this.health <= 0) {
+      Animal.removeFromAlive(this);
+    }
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(target) {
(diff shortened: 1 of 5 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+      Animal.removeFromAlive(this);
+    }
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
OleksandrRezanov, line 43: You don`t need to write constructor here, you can remove it and your code will work the same

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.decreaseHealth(50);
```
OleksandrRezanov, line 47: Is it necessary to create decreaseHealth function? You use it only once. 
Also you can make const biteValue = 50 and use it insted of premitives

OleksandrRezanov (approved):
Good work! You can simplify your code.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
```
DanilWeda, line 11: Alive needs to be static in Animal.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
+
+  decreaseHealth(amount) {
+    this.health -= amount;
+
+    if (this.health <= 0) {
```
DanilWeda, line 34: For 0, 50, and 100, the best way add in constant. ( for best readability)

DanilWeda (changes requested):
Great! Check some comments! 
Thanks!

src/herbivoresAndCarnivores.js
EdenParadise: linter doesnt allow it

olya-shyian: try to write ```static alive = [];``` instead

DanilWeda (approved):
Good job. 
Need add eslint-parser to fix test suite crash)

Approved. 
Thanks.

PR #2422, opened 2023-12-29
Solution

avakiel:


AntonDuchenko (approved):
Good job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
```
DanilWeda, line 11: alive needs to be static ( if the linter drops warning, add eslint-parser in deeps.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
DanilWeda, line 4: all numbers ( 0, 50, 100) move in constants.

DanilWeda (changes requested):
Thanks for the request! 
Check small comments! 
Good job!

src/herbivoresAndCarnivores.js
avakiel: Discussed it on Q&A. Mentor said don't touch the linter, and my solution is OK.

DanilWeda (approved):
Great! 
Approved.
Thanks.

PR #2427, opened 2024-01-01
add task solution

dyarechuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  health = 100;
```
vSainiuk, line 5: The tests failed. Try to avoid test failures by using a different way to write static and property values for objects in the code

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
YegorKorenkov, line 4: For fix test errors please try to do this points:
1. `npm i @babel/eslint-parser`
2. in .eslintrc.js change
`parser: "@babel/eslint-parser"`
3. In package.json check the package version
`"@babel/eslint-parser": "^7.22.15",`

src/herbivoresAndCarnivores.js
YegorKorenkov, line 5: Should be in constructor

YegorKorenkov (changes requested):
Good job!
Please check minor mistakes and if you want try to fix tests

src/herbivoresAndCarnivores.js
dyarechuk: Tried to instal @babel/eslint-parser, add this parser in .eslintrc.js file and change package version in package.json
But tests still fail, how can I resolve it?😅

YegorKorenkov (approved):
Well done! Don't worry about tests

PR #2428, opened 2024-01-02
add task solution1

MaxKalalashnyk02:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
```
dyarechuk, line 9: The "hidden" property is not necessary for carnivores, as stated in the example:

const deer = new Herbivore('Bembi');
const panther = new Carnivore('Bagira');
const lion = new Carnivore('King');
const rabbit = new Herbivore('Max');

Animal.alive === [
  {name: 'Bembi', health: 100, hidden: false},
  {name: 'Bagira', health: 100},
  {name: 'King', health: 100},
  {name: 'Max', health: 100, hidden: false},
];

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+  }
+
+  hide() {
+    this.hidden = true;
+    Animal.removeDeadAnimals();
```
dyarechuk, line 25: You can remove this line, because you remove dead animal from alive array when the carnivore bites herbivore

dyarechuk (approved):
Good Job! 🔥
Check the comments)

YegorKorenkov (changes requested):
You only have the errors that Dmytro indicated in, please look through his comments and correct them

PR #2430, opened 2024-01-02
Solution

yepolotn1ak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      const indexDies = Animal.alive.findIndex(
+        objAnimal => objAnimal.health === 0
+      );
+
+      Animal.alive.splice(indexDies);
+    }
```
Esceype, line 36: Use filter method instead and move it to method of Animal class

Esceype (changes requested):
GJ!
Read the following comment:

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  checkHealth() {
+    if (this.health === 0) {
+      Animal.alive = Animal.alive.filter(animal => animal.health > 0);
+    }
+  }
```
anastasiia-tilikina, line 14: it is better to make this method to be a static method of Animal class since we're filtering static property.

make it like this:

```javascript
Animal.checkHealth = function() { //...
``` 

also, we omit additional check in this case
```suggestion
  checkHealth() {
      Animal.alive = Animal.alive.filter(animal => animal.health >= 0);
  }
```

anastasiia-tilikina (changes requested):
Almost done!
let's improve your code a bit more)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
+Animal.alive = [];
+
+Animal.checkHealth = function() {
+  Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```
anastasiia-tilikina, line 14:
```suggestion
  Animal.alive = Animal.alive.filter(animal => animal.health >= 0);
```

anastasiia-tilikina (approved):
Well done!

But pay attention on minor issue here, in my comment

PR #2439, opened 2024-01-06
add task solution

Victoriia-Melnyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(object) {
+    if ((object instanceof Herbivore) && (object.hidden === false)) {
```
```suggestion
    if (object instanceof Herbivore && !object.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+      object.health -= 50;
+    }
+
+    const indexToRemove = Animal.alive.findIndex(animal => animal.health <= 0);
+
+    if (indexToRemove !== -1) {
+      Animal.alive.splice(indexToRemove, 1);
+    }
```
nazarmatsevych, line 38: Use filter method instead

nazarmatsevych (changes requested):

nazarmatsevych (approved):
🔥

PR #2441, opened 2024-01-07
Solution

Cy90sure:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,44 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
l4st1m0za, line 31: The bite method is not checking if the animal to bite is an instance of Carnivore. It should not decrease the health of the animal if it is a Carnivore.

l4st1m0za (changes requested):
Good job, but some changes needed

l4st1m0za (approved):
Good job!

PR #2442, opened 2024-01-08
Solution

o-cherpak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+  bite(obj) {
+    if (obj.hidden === false) {
+      obj.health -= 50;
+    }
+
+    if (obj.health === 0) {
+      const index = Animal.alive.indexOf(obj);
+
+      Animal.alive.splice(index, 1);
+    }
```
etojeDenys, line 44: you can simplify it using the filter method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
```
etojeDenys, line 28: we don't have name property in task description

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super();
+    this.name = name;
+
+    this.hidden = false;
+
+    Animal.alive.push(this);
```
etojeDenys, line 19: we can push it only one time in the Animal constructor

etojeDenys (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #2447, opened 2024-01-10
Solution

tania-kuzmenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.health = 100;
+    this.hidden = false;
+    Animal.alive.push(this);
```
etojeDenys, line 20: you can push only one time just move to the Animal conctructor

etojeDenys (approved):
lgtm

PR #2450, opened 2024-01-11
Animal, Herbivores and Сarnivores have been created

Dmytro-Bashlai:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    if (this instanceof Animal) {
```
etojeDenys, line 8: redundant check, it's always will be true

etojeDenys (approved):
lgtm

PR #2452, opened 2024-01-12
solution

yurovych:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb.hidden === false) {
+      herb.health -= 50;
+    }
+
+    if (herb.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(herb), 1);
```
nazarmatsevych, line 30: use filter method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herb) {
+    if (herb.hidden === false) {
```
nazarmatsevych, line 25:
```suggestion
    if (!herb.hidden) {
```
and consider check the instance of herb

nazarmatsevych (changes requested):

yurovych:
I didn't get what to do with this string
if (!herb.hidden) {

etojeDenys (approved):
lgtm

PR #2457, opened 2024-01-15
add task solution "js_herbivores_and_carnivores"

Andriy-Kutyshenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,62 @@
...
+
+  // checkHealth method - checks the health of the herbivore
+  // and removes it from the alive array if health <= 0
+  checkHealth() {
+    if (this.health <= 0) {
+      const index = Animal.alive.indexOf(this);
+
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
+      }
```
anastasiia-tilikina, line 42: the method of deleting items from static property - is better to make static as well.
and better to use filter method here.

anastasiia-tilikina (changes requested):
Good job!
jus a small suggestion.
Also, it is good that you intend to document your code via some explanation comment. But here is my advice: The best way to document your code - is writing self-descriptive code) It is great if you've written such code - wich doesn't need comments.

Use comments in case if the logic is complicated and non-obvious)

etojeDenys (approved):
awesome

PR #2458, opened 2024-01-17
solution

fessenko-tr:


maxim2310 (approved):
Good work, just when you see method that is called isStillAlive, you are expected that this method will return boolean value, but not filter some array

PR #2467, opened 2024-01-21
Add solution

VladyslavKukharchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,58 @@
...
+    this._health = 100;
+    Animal.alive.push(this);
+  }
+
+  static die(animal) {
+    const index = Animal.alive.indexOf(animal);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
etojeDenys, line 15: you can just use filter here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,58 @@
...
+  get health() {
+    return this._health;
+  }
+
+  set health(value) {
+    this._health = value;
+
+    if (this._health <= 0) {
+      Animal.die(this);
+    }
```
etojeDenys, line 27: looks redundant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,58 @@
...
+  bite(target) {
+    if (target instanceof Carnivore) {
+      return;
+    }
+
+    if (target.hidden) {
+      return;
+    }
+
+    target.health -= 50;
```
etojeDenys, line 54: better to call die here

etojeDenys (approved):
good job

PR #2470, opened 2024-01-22
task solution added

kupriiev:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,48 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
etojeDenys, line 34: looks redundant, 'cause you call only super here so you can remove it

etojeDenys (approved):
good job

PR #2479, opened 2024-01-25
Solution

Krzychu92:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+  static removeFromAlive(animal) {
+    const index = Animal.alive.indexOf(animal);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
nazarmatsevych, line 14: Don't manipulate with an array in this way, use filter method instead

nazarmatsevych (changes requested):
Almost done, please check the comment below

nazarmatsevych (approved):
🔥

PR #2483, opened 2024-01-25
Herbivores and Carnivores function | JS

julia-dronova:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
```
vadiimvooo, line 6: Your approach is good but it's better to define default value on func argument

```suggestion
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
+    super(name, health);
+    this.hidden = hidden || false;
+  }
```
vadiimvooo, line 25: The same situation as at previous comment.

```suggestion
  constructor(name, health, hidden = false) {
    super(name, health);
    this.hidden = hidden;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+
+  hide() {
+    this.hidden = true;
```
vadiimvooo, line 28: It's better to change the hidden value to oposite one but not necessary

```suggestion
    this.hidden = !this.hidden;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+  constructor(name, health) {
+    super(name, health);
+    this.demage = 50;
+  }
+
+  bite(herbivore) {
+    if (
+      herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= this.demage;
+      Animal.alive = Animal.removeDead();
```
vadiimvooo, line 42: It's not good practise to call this method every time bite one is called. Just add check
```suggestion
      if (herbivore.health <= 0) {
        Animal.alive = Animal.removeDead();
      }
```

vadiimvooo (changes requested):

src/herbivoresAndCarnivores.js
julia-dronova: Thank you for feedback!) All recommendations were applied.

IvanFesenko (approved):
well done

PR #2484, opened 2024-01-25
Solution

OksanaZderko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
anastasiia-tilikina, line 16: move it to Animal Class to avoid code duplication

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
anastasiia-tilikina, line 31:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.filter(an => an.health > 0);
```
anastasiia-tilikina, line 36: move this filter to a static method of Animal class and call it here.
something like this

```javascript
Animal.removeDead = () => { //...
```

anastasiia-tilikina (changes requested):
just a couple fixes

anastasiia-tilikina (approved):
Well done! 👍🏼

PR #2485, opened 2024-01-25
Solution

reznik-denis:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      const index = Animal.alive.indexOf(animal);
+
+      Animal.alive.splice(index, 1);
```
etojeDenys, line 33: you can just use filter method

etojeDenys (approved):
lgtm

src/herbivoresAndCarnivores.js
reznik-denis: +

PR #2492, opened 2024-01-30
task solution

killfox93:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
nazarmatsevych, line 17: Use filter method instead

nazarmatsevych (changes requested):
Almost done, please check the comment below

lerastarynets (approved):
You're on the right track!

PR #2495, opened 2024-01-31
Solution

MaksymKolesnyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore) {
+      const an = Animal.alive.findIndex(obj => obj === animal);
+
+      if (an && animal.hidden === false) {
+        Animal.alive[an].health -= 50;
+      }
```
etojeDenys, line 32:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
        animal.health -= 50;
      }
```
we can simplify it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
+      if (Animal.alive[an].health === 0) {
+        Animal.alive.splice(an, 1);
+      }
```
etojeDenys, line 36: just filter that array

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #2498, opened 2024-01-31
add task solution

YevhenProtasov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
```suggestion
    this.hidden = true
```
This method should only change hidden value to true

nazarmatsevych (approved):
Great work!

PR #2503, opened 2024-02-03
solution

kapesha:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      const toDelete = Animal.alive.indexOf(animal);
+
+      Animal.alive.splice(toDelete, 1);
+    }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,40 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

Same but shorter

nazarmatsevych (changes requested):
Almost done, let's make some imrpovements

nazarmatsevych (approved):
👍

PR #2506, opened 2024-02-04
Solution

monikaprochniewicz:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(health, name) {
+    super(100, name);
+  }
+
+  bite(herbAnimal) {
+    if (!herbAnimal.hidden && herbAnimal instanceof Herbivore) {
```
BogdanMaliuta, line 30:
```suggestion
    if (
      !herbAnimal.hidden 
      && herbAnimal instanceof Herbivore
    ) {
```

BogdanMaliuta (approved):
WD, but split lines with more than 80 chars on rows

PR #2508, opened 2024-02-06
add animal herbivore carnivore classes

self51:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    const canBite = herbivore.hidden === false
```
VitaliyBondarenko1982, line 26:
```suggestion
    const canBite = !herbivore.hidden
```
it will be the same

VitaliyBondarenko1982 (approved):
GJ!

PR #2514, opened 2024-02-07
The solution for the assignment was implemented

alionazherdetska:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,39 @@
...
+  }
 }
+Animal.alive = [];
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
etojeDenys, line 16: you can push and write it only one time in the Animal constructor
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,39 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
```
etojeDenys, line 27:
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,39 @@
...
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+      Animal.alive = Animal.alive.filter((creature) => creature.health > 0);
+
+      return target.health;
+    } else {
+      return -1;
+    }
```
etojeDenys, line 36:
```suggestion
    }
    
    
    return -1;
```

etojeDenys (approved):
good job

PR #2528, opened 2024-02-12
Solution

Zava96:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    super(100, name);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    if (this.hidden === false) {
+      this.hidden = true;
+    } else {
+      this.hidden = false;
+    }
```
etojeDenys, line 24:
```suggestion
    this.hidden = !this.hidden;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    super(100, name);
+  }
+
+  bite(herbAnimal) {
+    if ((!herbAnimal.hidden) && (herbAnimal instanceof Herbivore)) {
+      herbAnimal.health -= 50;
+    }
+
+    if (herbAnimal.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(herbAnimal), 1);
```
etojeDenys, line 39: we can just use the filter method here instead of splice and indexOf

etojeDenys (approved):
good job

PR #2529, opened 2024-02-12
Develop

BoGoBooM:
I think I've started to understand JS
At least better than before, but still difficult to master


etojeDenys (approved):
awesome

PR #2539, opened 2024-02-15
solution

Prokopovych16:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,43 @@
...
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        const index = Animal.alive.indexOf(animal);
+
+        if (index !== -1) {
+          Animal.alive.splice(index, 1);
+        }
+      }
```
etojeDenys, line 39: you can just use the filter method instead of the indexOf and splice

etojeDenys (approved):
good job

PR #2540, opened 2024-02-15
Develop

zozulia-developer:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
etojeDenys, line 16: you can move it in the Animal constructor and write it only once

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
```
etojeDenys, line 28:
```suggestion

```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    if (
+      animal instanceof Herbivore
+        && !animal.hidden
+        && animal.health > 0
+    ) {
+      animal.health -= 50;
+    }
+
+    if (animal.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
etojeDenys, line 40: better to use the filter method instead of the splice and indexOf

etojeDenys (changes requested):

PR #2544, opened 2024-02-17
add Solution

Taras-Hanyshyn:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(anim) {
+    if(anim.hidden === false) {
```
nazarmatsevych, line 29:
```suggestion
    if(!anim.hidden) {
```

same but a bit shorter and consider check the instance of `anim`

nazarmatsevych (changes requested):
Almost done, please fix the comment below

Andriy-Shymkiv (approved):
🔥

PR #2550, opened 2024-02-19
Solution

sergii-stognii:


sergii-stognii:
I feel that its a bot issue cause there shouldnt be a problem

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
Viktor-Kost, line 4: move this initialization outside the class (like this: `Animal.alive = [];`)
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    this.health -= damage;
+
+    if (this.health <= 0) {
+      this.die();
+    }
+  }
+
+  die() {
+    Animal.alive = Animal.alive.filter(animal => animal !== this);
+  }
```
Viktor-Kost, line 22: each Animal shouldn't have these methods. You should just implement this logic to `Carnivore `
```suggestion
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.takeDamage(50);
```
Viktor-Kost, line 39: so here you'll implement logic to take a damage. 
Plus add a condition to remove target from an array if target is dead
```suggestion
      target.health -= 50;
```

Viktor-Kost (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #2553, opened 2024-02-20
Add solution

ViktoriiaRepo:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  }
+
+  bite(animal) {
+    if (animal instanceof Carnivore
+      || (animal instanceof Herbivore && animal.hidden)) {
+      return;
+    }
+
+    animal.health -= 50;
+    Animal.removeDeadAnimal();
```
polosanya, line 40: Would be great to check if animal's health is zero before calling it to avoid extra calls

polosanya (changes requested):
Comments above

BudnikOleksii (approved):
Good work 👍

PR #2554, opened 2024-02-20
solution

ChrisHryts:


HolyRaveOS (approved):
Good joob!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
```
Moroz-Dmytro, line 6:
```suggestion
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false && herbivore instanceof Herbivore) {
```
Moroz-Dmytro, line 26:
```suggestion
    if (!herbivore.hidden && herbivore instanceof Herbivore) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+      herbivore.health -= 50;
```
Moroz-Dmytro, line 27: Avoid using "Magic values"
Create constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
Moroz-Dmytro, line 25: `herbivore` is not really good name here, because it can be also a `carnivore`

src/herbivoresAndCarnivores.js
Moroz-Dmytro: But also it's magic value, create constant

Moroz-Dmytro (changes requested):
Good work 👍 
But we can improve your solution

stassokolov1 (approved):
Looks good to me🔥

PR #2555, opened 2024-02-21
solution

StageUa:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,58 @@
 'use strict';
 
+// Define the Animal class
```
OleziO, line 3: Remove all comments from the code.

OleziO (changes requested):

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,57 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Carnivore) {
+    } else if (target.hidden) {
+    } else {
+      target.decreaseHealth(50);
+    }
+  }
```
anastasiiavorobiova, line 50:
```suggestion
    if (target instanceof Herbivore && !target.hidden) {
      target.decreaseHealth(50);
    }
  }
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

etojeDenys (approved):
good job

PR #2563, opened 2024-02-24
Add solution

SokolAlena:


src/herbivoresAndCarnivores.js
```diff
@@ -2,16 +2,42 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && beast.hidden === false) {
```
```suggestion
    if (beast instanceof Herbivore && !beast.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,16 +2,42 @@
...
+      beast.health -= 50;
+    }
+
+    if (beast.health === 0) {
```
SanyaBratashchuk, line 31:
```suggestion
    if (beast.health <= 0) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,16 +2,42 @@
...
+      const indexForDelete = Animal.alive.indexOf(beast);
+
+      Animal.alive.splice(indexForDelete, 1);
```
SanyaBratashchuk, line 34: u can use filter here instead of using 2 array methods

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #2565, opened 2024-02-25
Solution

flocky674:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+    }
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
etojeDenys, line 19: you can just use the filter method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
+
+  die() {
+    super.die();
+  }
```
etojeDenys, line 35: redundant, it works like this by default

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
+    }
```
etojeDenys, line 10: FYI we can just create it out of the class 
```
class Animal {...}

Animal.alive = [];

```

etojeDenys (changes requested):

flocky674:
solution*

etojeDenys (approved):
awesome

PR #2566, opened 2024-02-26
Solution

Salmaralex:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (!(target instanceof Herbivore)) {
+    } else if (target.hidden) {
+    } else {
+      target.takeDamage(50);
+
+      if (target.health <= 0) {
```
l4st1m0za, line 47: Checking if target.health <= 0 and calling target.die() method is redundant. The 'takeDamage' method already does this.

l4st1m0za (approved):
Good job!

PR #2569, opened 2024-02-26
done

anastasia1383:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
ViktoriiaRepo, line 4: You can create it as follow to avoid test error:
Animal.alive = []

ViktoriiaRepo (approved):
Good job!

BudnikOleksii (approved):
Good work 👍

PR #2570, opened 2024-02-26
add task solution

Marina-Semipyatnaya:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
etojeDenys, line 14: we can just use the filter method here

etojeDenys (approved):
good job

PR #2572, opened 2024-02-27
add task solution

DenLysiak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= 50;
+    }
+
+    if (victim.health === 0) {
```
l4st1m0za, line 31: The condition 'victim.health === 0' only checks if the victim's health is exactly 0. It should be 'victim.health <= 0' because the health can drop below 0.

PR #2576, opened 2024-02-28
add task solution

kshepetska:


Vasya564 (approved):
Good job🚀

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
BudnikOleksii, line 4: Avoid magic values, create constants

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.name = name;
+    this.health = health;
+
+    if (!Animal.alive) {
+      Animal.alive = [];
```
BudnikOleksii, line 9: Do it only once outside the class so you don't need this check

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    }
+
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
```
BudnikOleksii, line 18:
```suggestion
  constructor(name, health) {
    super(name, health);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive = Animal.alive.filter((a) => a !== animal);
```
BudnikOleksii, line 33: Avoid one-letter variables/arguments

BudnikOleksii (changes requested):
Good job 👍 
Let's improve your solution slightly

BudnikOleksii (approved):
Looks better 👍

PR #2578, opened 2024-02-28
Solution

kovaltar:
Unfortunately it looks like Linter doesn't support static properties in classes, so to pass Linter I commented line
static alive = [];
in class Animal, and added this property after class

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,47 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  // static alive = [];
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      Animal.alive.forEach((item, index, arr) => {
+        if (item.name === animal.name) {
+          if (animal.health <= 0) {
+            arr.splice(index, 1);
+          } else {
+            item.health = animal.health;
+          }
```
anastasiiavorobiova, line 40:
```suggestion
          }
```

anastasiiavorobiova (approved):
Great work!

PR #2579, opened 2024-02-28
solution

k-shestakov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(creature) {
+    if (creature instanceof Herbivore && !creature.hidden) {
+      creature.health -= 50;
+
+      if (creature.health <= 0) {
+        const deadAnimal = Animal.alive.indexOf(creature);
+
+        Animal.alive.splice(deadAnimal, 1);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
nazarmatsevych, line 20:
```suggestion
    this.hidden = true;
```

this method should only change hidden value to true

nazarmatsevych (changes requested):
Almost done, let's make some improvements

etojeDenys (approved):
lgtm

PR #2582, opened 2024-02-29
add solution

Vlobdam:


IRONH4MMER:
To fix `npm test` issues try this
image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (Animal.alive.includes(target) && target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+
+      if (target.health <= 0) {
```
IRONH4MMER, line 29: The creature is already dead when health is on 0%, change it to 50

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
IRONH4MMER, line 4: You must define it like `Animal.alive` just like in readme file

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
IRONH4MMER, line 20: You can't unhide using `hide()` method

IRONH4MMER (changes requested):
Good work, but a few improves needed!

src/herbivoresAndCarnivores.js
Vlobdam: you still reference to it like Animal.alive

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  static alive = [];
+
+  static updateAliveList = () => {
+    Animal.alive = Animal.alive.filter(animal => animal.health > 0)
+  }
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
```
BudnikOleksii, line 13: - "Health is 100 by default." So it means that you still can pass a health parameter to the constructor if needed
- 100 is a magic value, create a constant for it
- Put constructor before methods

BudnikOleksii (approved):
Good work 👍 
Just a few changes are required, check the comments

PR #2584, opened 2024-02-29
Solution

winniepuhh:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
-  // write your code here
+  bite(target) {
+    if (!(target instanceof Herbivore)) {
+      return;
+    }
+
+    if (target.hidden) {
+      return `${target.name}`;
+    }
+    target.decreaseHealth(50);
```
dima8528, line 45: same staff with a trial line

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+
+    if (this.health <= 0) {
+      Animal.alive = Animal.alive.filter(animal => animal !== this);
+
+      return `${this.name}`;
+    }
+  }
 }
 
+Animal.alive = [];
```
dima8528, line 21: Here I think we need to use the key word 'static' but if your solution works, I don't even know... Let mentors judge then

dima8528 (changes requested):

src/herbivoresAndCarnivores.js
winniepuhh: when i use just return i do not need use trial line

src/herbivoresAndCarnivores.js
winniepuhh: i can use static only in class

BudnikOleksii (approved):
Looks good 👍

PR #2586, opened 2024-02-29
WildLife

DenNice-cloud:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(carnivoreBite) {
+    if (carnivoreBite.constructor.name !== 'Carnivore'
+      && carnivoreBite.hidden !== true) {
+      carnivoreBite.health -= 50;
```
olduma, line 32:
```
if (!(carnivoreBite instanceof Carnivore) && !carnivoreBite.hidden) {
  carnivoreBite.health -= 50;
```

olduma (approved):
Overall good job! You can make a little edit for reliability!

BudnikOleksii (changes requested):
Good work 👍 
Just a few suggestions, check them

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
+const currentHealth = 100;
```
BudnikOleksii, line 3:
```suggestion
const DEFAULT_HEALTH = 100;
```
It's default I think, and it should be in upper case

BudnikOleksii (approved):
👍

PR #2588, opened 2024-02-29
add task solutions

dima8528:


illya-onyshchuk (approved):
Well done 👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
BudnikOleksii, line 8: - Avoid magic values such as 100, etc. Create constants
- Health is 100 by default, but I think it should be possible to path another health value if needed

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution a bit

src/herbivoresAndCarnivores.js
dima8528: Somewhy in vs code it's said that the commit was sent but here I can't see it. Odd

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
BudnikOleksii, line 6: - Avoid magic values such as 100, etc. Create constants

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.health = health;
+
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
```
BudnikOleksii, line 16:
```suggestion
  constructor(name, health) {
    super(name, health);
```

BudnikOleksii (approved):
Approved, but check the comments

PR #2589, opened 2024-02-29
add solution

olduma:


Graf6612:
Hello, please check your code because some tests failed)

Graf6612 (changes requested):
Hello,good job but some test some tests failed,plese fix the error))

olduma:
I have no idea! Locally all test passed without any trouble.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
```
BudnikOleksii, line 5: - Avoid magic values such as 100, etc. Create constants

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
+
+  decreaseHealth(amount = 50) {
```
BudnikOleksii, line 22: It's a strange method that is used only in other Class, so it doesn't make much sense

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    this.health -= amount;
+
+    if (this.health <= 0) {
+      Animal.alive = Animal.alive.filter(a => a !== this);
```
BudnikOleksii, line 26: Avoid one-letter variables and parameters

BudnikOleksii (changes requested):
Good work 👍 
Just a few moments should be fixed, check the comments

Moroz-Dmytro (approved):
Looks good 🥇

PR #2594, opened 2024-03-01
Solution

ivan-gritsenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore) {
+      const indexOfHerbivore = Animal.alive.findIndex((animal) => {
```
yaGodya, line 27: You can use an arrow function instead of an anonymous function in the findIndex method for convenience.

yaGodya (approved):
Great job, the code looks clear and readable.

My experience does not allow me to find an opportunity to somehow improve the code.

However, a mentor can give you some advice.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  name = '';
+  health = 100;
```
BudnikOleksii, line 6: - You don't need to set name here, just set it inside a constructor
- Avoid magic values such as 100, etc. Create constants

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore) {
+      const indexOfHerbivore = Animal.alive.findIndex((animal) => {
+        return animal === herbivore && animal.hidden === false;
+      });
```
BudnikOleksii, line 29: It doesn't make sense, as `bite` method accept herbivore itself

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+      if (indexOfHerbivore !== -1) {
+        Animal.alive[indexOfHerbivore].health -= 50;
```
BudnikOleksii, line 32:
```suggestion
      herbivore.health -= 50;
```
And 50 is a magic value

BudnikOleksii (changes requested):
Almost done, let's get rid of redundant things and improve your solution slightly

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
+const HEALTH = 100;
+const DAMAGE = 50;
```
BudnikOleksii, line 4:
```suggestion
const DEFAULT_HEALTH = 100;
const HERBIVORE_DAMAGE = 50;
```
Describe your variables better, so everyone can understand to what they are related

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
BudnikOleksii, line 28:
```suggestion
    if (!herbivore.hidden) {
```
You don't need to compare with `false`

BudnikOleksii (approved):
Approved, but I left a few comments for the future, check them

PR #2597, opened 2024-03-01
Solution

Math8585:


Oleksandr2108 (approved):
Good job!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
BudnikOleksii, line 6: - Avoid magic values such as 100, etc. Create constants

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  static alive = [];
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+
+    Animal.alive.push(this);
```
BudnikOleksii, line 17: You should do it in the Animal class, as if any Animal is created it should be stored in this array

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    Animal.alive.push(this);
+  }
+
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
BudnikOleksii, line 33:
```suggestion
    if (!herbivore.hidden) {
```
You don't need to compare with false

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+      herbivore.health -= 50;
+
+      if (herbivore.health === 0) {
```
BudnikOleksii, line 36:
```suggestion
      if (!herbivore.health) {
```
With 0 as well
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

BudnikOleksii (approved):
Good work 👍 
Left a few suggestions, check them and fix your code accordingly

PR #2598, opened 2024-03-01
add solution

yaGodya:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal.constructor === Carnivore || animal.hidden) {
+      return;
+    }
+
+    animal.health -= ANIMAL_DAMAGE;
+
+    if (animal.health <= MINIMUM_HP) {
+      Animal.alive = Animal.alive.filter((a) => a !== animal);
```
mrkovaliv, line 33: Perhaps it would be better to move this logic to a separate method `die` (for the Animal class) for better reusability and readability.
  
```
  die() {
    Animal.alive = Animal.alive.filter((a) => a !== this);
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
+const ANIMAL_DAMAGE = 50;
+const ANIMAL_BASE_HP = 100;
+const MINIMUM_HP = 0;
```
mrkovaliv, line 5: Change "MINIMUM_HP" to "ANIMAL_MINIMUM_HP" to make the constant names consistent.

mrkovaliv (changes requested):
Nice work, but I have some small improvements

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+const ANIMAL_DAMAGE = 50;
+const ANIMAL_BASE_HP = 100;
+const ANIMAL_MINIMUM_HP = 0;
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = ANIMAL_BASE_HP;
+    this.hidden = false;
```
Moroz-Dmytro, line 11: Only `Herbivore` should have this field `hidden`, not every animal

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    Animal.alive.push(this);
+  }
+
+  static alive = [];
+
+  die() {
+    Animal.alive = Animal.alive.filter((a) => a !== this);
```
Moroz-Dmytro, line 18: `a` is a bad name

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.constructor === Carnivore || animal.hidden) {
```
Moroz-Dmytro, line 30: Better to use [instanceof](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof)

Moroz-Dmytro (changes requested):
Good work 👍 
But we can improve your solution

BudnikOleksii (approved):
👍

PR #2601, opened 2024-03-01
task done

Oleksandr2108:


Moroz-Dmytro (changes requested):
Good progress 👍 
Need to fix some comments below to make it better

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+    this.name = name;
+    this.health = Animal.health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
   // write your code here
+  constructor() {
+    super();
```
BudnikOleksii, line 18: The class should accept `name` and pass it to the parent constructor via super

BudnikOleksii (approved):
👍

PR #2602, opened 2024-03-02
add task solution

vlesia:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
```
IRONH4MMER, line 30: Looks like a magic number

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
IRONH4MMER, line 4: You should define it outside of the class.

IRONH4MMER (changes requested):
Great job👍
There are a few minor fixes left to make

Moroz-Dmytro (changes requested):
Good work 👍 
But need to fix some comments below

Moroz-Dmytro (changes requested):
Please, fix comments above
Need to reject again, because of some problems with platform

BudnikOleksii (approved):
Good work 👍

PR #2604, opened 2024-03-03
Develop

ogerenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+      this.die();
+    }
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
etojeDenys, line 25: we can just use the filter method

etojeDenys (approved):
good job

PR #2605, opened 2024-03-03
animals

OMedvid:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
etojeDenys, line 17: we can just use the filter method

etojeDenys (approved):
good job

PR #2608, opened 2024-03-03
solution

DmitryKot96:


vlesia (approved):
Good job!
Perhaps you should make the decreaseHealth method static.

PR #2615, opened 2024-03-05
MySolution

DenLys21:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === undefined || herbivore.hidden === true) {
+      return;
+    }
+
+    herbivore.health -= 50;
+
+    if (herbivore.health === 0) {
+      Animal.alive.splice(Animal.alive.indexOf(herbivore), 1);
```
nazarmatsevych, line 30: Use filter method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === undefined || herbivore.hidden === true) {
+      return;
+    }
+
+    herbivore.health -= 50;
```
nazarmatsevych, line 27:
```suggestion
    if (!herbivore.hidden) {
          herbivore.health -= 50;
    }
```

You can simplify this check. And also you need to check the instance of herbivore object, it should be the instance of Herbivore  class

nazarmatsevych (changes requested):
Good start, let's make some improvements

etojeDenys (approved):
lgtm

PR #2616, opened 2024-03-05
Solution

getchard:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
etojeDenys, line 33: it's called by default, so we can remove it

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #2620, opened 2024-03-06
tests were passed

Greshch:


src/herbivoresAndCarnivores.js
```diff
@@ -2,6 +2,18 @@
...
+  hidden = false;
+
+  constructor(name) {
+    this.name = name;
+    Animal.alive.push(this);
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
SanyaBratashchuk, line 16: only instances of Herbivore should have hide method

src/herbivoresAndCarnivores.js
```diff
@@ -10,6 +22,25 @@ class Herbivore extends Animal {
...
+      return;
+    }
+
+    if (obj.hidden === true) {
+      return;
+    }
+
+    if (obj.health <= 0) {
+      return;
+    }
```
SanyaBratashchuk, line 36:
```suggestion
    if (obj instanceof Carnivore || obj.hidden) {
      return;
    }
```

src/herbivoresAndCarnivores.js
```diff
@@ -10,6 +22,25 @@ class Herbivore extends Animal {
 
 class Carnivore extends Animal {
   // write your code here
+  bite(obj) {
```
SanyaBratashchuk, line 25:
```suggestion
  bite(victim) {
```

SanyaBratashchuk (changes requested):

src/herbivoresAndCarnivores.js
Greshch: Ok.

Greshch:
I think it's much better.

VitaliyBondarenko1982 (approved):
Looks good.

PR #2625, opened 2024-03-07
add solution

Dmitrel:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+    this.hidden = hidden;
+    Animal.alive.push(this);
```
okarnaukh, line 17: You don`t need this line, because you already do this in parent constructor.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
+      herbivore.health -= 50;
```
okarnaukh, line 28: You can use a constant for this value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
okarnaukh, line 6: You can use a constant for health value

okarnaukh (commented):
Good job. It could be improved a little

okarnaukh (changes requested):
Good job. It could be improved a little

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false) {
```
BudnikOleksii, line 29:
```suggestion
    if (!herbivore.hidden) {
```

BudnikOleksii (approved):
👍

PR #2629, opened 2024-03-08
Solution

Hy-tapa-kot:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static removeFromAlive(animal) {
+    const index = Animal.alive.indexOf(animal);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
nazarmatsevych, line 17: Use filter method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
```
nazarmatsevych, line 38:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

nazarmatsevych (changes requested):
Good work, let's make some imrpovements

nazarmatsevych (approved):
Well done 👍

PR #2632, opened 2024-03-09
solution

Dev1NA:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+
+  bite(animal) {
+    if (
+      animal instanceof Carnivore ||
+      (animal.hidden && animal instanceof Herbivore)
+    ) {
+      return;
+    }
+
+    animal.health -= 50;
```
```suggestion
    if (!animal.hidden && animal instanceof Herbivore) {
      animal.health -= 50;
    }
```

We can simplify it a bit

nazarmatsevych (changes requested):
Almost done, let's make some improvements

nazarmatsevych (approved):
Good work!

PR #2636, opened 2024-03-10
add task solution

khromakeiqt:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
okarnaukh, line 8: You can use constant for health value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
```
okarnaukh, line 37: The same, you can use constant

okarnaukh (approved):
Nice!

src/herbivoresAndCarnivores.js
BudnikOleksii: > You can use constant for health value

Good point

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  checkHealth() {
```
BudnikOleksii, line 12: It's not check as you also remove animal from the `Alive` array

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    if (this.health <= 0) {
+      const index = Animal.alive.indexOf(this);
+
+      if (index > -1) {
+        Animal.alive.splice(index, 1);
+      }
```
BudnikOleksii, line 18: [CODE PERFOMANCE] - Try to avoid nesting loops (indexOf + splice), use filter instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
BudnikOleksii, line 31: Method `hide` should only hide, not toggle

src/herbivoresAndCarnivores.js
BudnikOleksii: > The same, you can use constant

Good point

BudnikOleksii (changes requested):
Almost done. Left a few comments, check them

BudnikOleksii (approved):
👍

PR #2640, opened 2024-03-11
task solution

Kacper7001:


YarynaPuhach:
The code effectively implements the concept of Herbivores and Carnivores as extensions of the Animal class.

PR #2641, opened 2024-03-11
solution

weswojciech:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,38 @@
...
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
   // write your code here
+  constructor() {
+    super();
+    this.hidden = false;
+  }
```
marinatea, line 19: the Herbivore is missing a parameter 'name' that is expected by the Animal constructor and in the Herbivore class constructor, you must apply the parameter name to the constructor of the parent Animal class using the super(name) methods :)

src/herbivoresAndCarnivores.js
marinatea, line 1: "If the health of the animal reaches 0, the beast dies and it should be removed from the static array `Animals.alive`." - it seems that you did not check this condition and did not remove the dead animals from the live Array :)

marinatea (changes requested):

PR #2642, opened 2024-03-11
Task solved

bartKarb:
https://github.com/bartKarb/js_herbivores_and_carnivores/tree/develop

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,16 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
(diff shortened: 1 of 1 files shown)


konradwasilewski444 (changes requested):
Very good, i think that the hole idea with filter at the end is very intresting and so easier (than for example my idea). But you dont fave to declarate constructor in every class. For example in Carnivarore it is not necessary u know. Because of 'extands' it is already there so you dont have to  write it aain in my opinion. 
Dont be mad at me pls haha

bartKarb:
Yup, I just tested it without constructor at the Carnivore class and the initial functionality remains untouched.

PR #2645, opened 2024-03-11
Develop

konradwasilewski444:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(someHerbivore) {
+    if (
+      typeof someHerbivore === 'object' &&
```
TideTwo, line 26: Maybe you don't need to check if it's an object, if you already check if it's an instance of the herbivore class, though I'm not sure.

TideTwo (approved):

src/herbivoresAndCarnivores.js
konradwasilewski444: you 100% rigth, i should delete it, instance is object

nataliaklonowska (approved):

nataliaklonowska (approved):
Great work 🚀

PR #2648, opened 2024-03-12
Solution

pawelkloc:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Carnivore || animal.hidden) {
+      return;
+    } else {
+      animal.health -= 50;
+    }
```
choeqq, line 38: when if is true else will never be hit, so `else` is redundant in this case

choeqq (changes requested):

choeqq (approved):

PR #2659, opened 2024-03-18
Develop

Beliyandr:


maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
+  constructor(name, health) {
+    super(name, health);
+    this.name = name;
+  }
```
etojeDenys, line 25: you specified the name in Animal constructor so that we can remove all this
```suggestion
 
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.name = name;
+  }
```
etojeDenys, line 36: same here

etojeDenys (changes requested):

maxim2310 (approved):
Good work!!!

PR #2667, opened 2024-03-22
add solution

rwlp:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,48 @@
...
+    Animal.alive.push(this);
+  }
+
+  bite(herbivore) {
+    if (!herbivore.hidden && !(herbivore instanceof Carnivore)) {
+      herbivore.health -= 50;
+
+      if (herbivore.health < 1) {
+        Animal.alive.splice(
+          Animal.alive.indexOf((e) => e.name === herbivore.name),
```
EdPirro, line 41: `indexOf` takes a value to be found and not a function, perhaps you meant to use `findIndex` instead;
The way this function is defined your `.indexOf` will always return `-1` which will cause your `splice` to always remove the last animal in `Animal.alive` array;

Try this test case:
```js
test(`Herbivores should be removed from 'Animal.alive' array, when their health is 0 (but Zebra is not the last defined animal)`, () => {
    const zebra = new Herbivore('Zebra');
    const tiger = new Carnivore('Tiger');

    tiger.bite(zebra);
    tiger.bite(zebra);

    expect(Animal.alive.includes(zebra))
        .toBe(false);

    expect(Animal.alive.includes(tiger))
        .toBe(true);
});
```

EdPirro, line 41: Also, using `findIndex` by the animal's name may cause some issues when two animals with the same name exist

Try this test case:
```js
test('Two same herbivore only one dies', () => {
    const olderZebra = new Herbivore('Zebra');
    const yongerZebra =  new Herbivore('Zebra');
    const lion = new Carnivore("Lion");

    lion.bite(yongerZebra);
    lion.bite(yongerZebra);

    expect(Animal.alive.includes(olderZebra))
      .toBe(true);

    expect(Animal.alive.includes(lion))
      .toBe(true);
});
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,48 @@
...
+    super(name);
+
+    Animal.alive.push(this);
+  }
+
+  bite(herbivore) {
+    if (!herbivore.hidden && !(herbivore instanceof Carnivore)) {
+      herbivore.health -= 50;
+
+      if (herbivore.health < 1) {
```
EdPirro, line 39: Since this is a herbivore's behavior it would make sense to allow the `Herbivore` class handle it and not move into the `Carnivore` class, could you come up with a way to achieve this?

Tip: remember `getters` and `setters`

**This is a personal suggestion/challenge, doing it this way is not required**

EdPirro (changes requested):
Great job, @rwlp! Very clean code structure👏. I spotted just a few small issues/suggestions that I'd like you to go through, please let me know if you have any questions!

rwlp:
Hello @EdPirro, I make the changes.

src/herbivoresAndCarnivores.js
```diff
@@ -2,16 +2,72 @@
...
+    return this.innerHealthVar;
+  }
 }
 
+const tiger = new Carnivore('Tiger');
+const zebra = new Herbivore('Zebra');
+
+tiger.bite(zebra);
+tiger.bite(zebra);
+
```
EdPirro, line 70: I don't think this is meant to be here

src/herbivoresAndCarnivores.js
```diff
@@ -2,16 +2,72 @@
...
+  bite(herbivore) {
+    herbivore.health -= 50;
+  }
+
+  set health(totalLifeDecreased) {}
+
+  get health() {
+    return this.innerHealthVar;
+  }
```
EdPirro, line 62: Default `health` behavior could be moved to the base class `Animal`

EdPirro (changes requested):
Awesome job, @rwlp 👏! I think you just left some debug code in lines 65 to 70, please remove them and you code should be good to go!

src/herbivoresAndCarnivores.js
rwlp: I create a default getter class in parente class. But we have some rules applied in prettier. I make some improvisation to assure that getter health method is inherited from parent class using 'super' keyword.

EdPirro (approved):
That's it, @rwlp! Great job, keep it up! 🚀

PR #2668, opened 2024-03-22
Add solution

olafchuszno:


BogdanMaliuta (approved):
GJ!
But for my opinion better solution would be to add health and hide to constructor, it`s allows to create different animals

PR #2670, opened 2024-03-22
Solution

valost:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden !== true) {
```
nazarmatsevych, line 35:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

same but a bit shorter

nazarmatsevych (approved):
Good work!

PR #2674, opened 2024-03-23
add task solution

ogchy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
etojeDenys, line 32:
```suggestion
```
looks redundant

etojeDenys (approved):
good job

PR #2676, opened 2024-03-25
add task solution

romanyshyns:


src/herbivoresAndCarnivores.js
```diff
@@ -1,19 +1,92 @@
...
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden === false) {
```
nazarmatsevych, line 73:
```suggestion
    if (target instanceof Herbivore && !target.hidden) {
```

Same but a bit shorter

nazarmatsevych (approved):
Good work!
please check the comment below

PR #2679, opened 2024-03-26
Develop

KrisPetrienko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && !beast.hidden) {
+      beast.health -= 50;
+    }
+
+    if (beast.health === 0) {
```
l4st1m0za, line 28: The animal is removed from the alive array only when its health is exactly zero. It should be removed when its health is zero or less.

l4st1m0za (approved):
Good job!

PR #2681, opened 2024-03-27
Solution task

IvanPovich:
```
  constructor(name) {
    super(name);
    this.haveTarget = false;
  }
  ```
  
  without the string 'this.haveTarget = false;' the test shows a syntax error


l4st1m0za (approved):
Good job!

PR #2682, opened 2024-03-27
Solution

nataliiailchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(target) {
+    if (!(target instanceof Herbivore) || target.hidden) {
+      return;
+    }
+    target.health -= 50;
```
```suggestion
    if (target instanceof Herbivore && !target.hidden) {
      target.health -= 50;
    }
```

we can simplify it a bit

nazarmatsevych (approved):
Good work! Please check the comment below

PR #2687, opened 2024-03-29
Task

georgiyevv:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 
 class Herbivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
VitaliyBondarenko1982, line 27:
```suggestion
    this.hidden = true;
```
this method should only hide animal.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+  }
```
VitaliyBondarenko1982, line 35: looks like redundant constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
   // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
```
VitaliyBondarenko1982, line 24: you can set this.hidden = false in Animal constuctor and remove constructor here.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments.

VitaliyBondarenko1982 (approved):
Good job.

PR #2689, opened 2024-03-29
js_herbivores_and_carnivores

misha-vynnyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  }
+
+  // Method to check and handle the animals death
+  checkHealth() {
+    if (this.health <= 0) {
+      const index = Animal.alive.indexOf(this);
+
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
+      }
```
nazarmatsevych, line 19: Use filter method instead:

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
```suggestion
    this.hidden = true;
```

This method should only change `hidden `value to `true`

nazarmatsevych (changes requested):
Almost done, please check the comments below :)

anastasiiavorobiova (approved):
Great work!

PR #2692, opened 2024-03-30
add task solution

timossha:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
+      animal.health -= 50;
+
+      if (animal.health === 0) {
+        const index = Animal.alive.indexOf(animal);
+
+        Animal.alive.splice(index, 1);
+      }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden === false) {
```
nazarmatsevych, line 25:
```suggestion
    if (!animal.hidden) {
```

and check the instance of animal

nazarmatsevych (changes requested):

nazarmatsevych (approved):
Well done 👍

PR #2697, opened 2024-04-01
Develop

Allanpt:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,59 @@
...
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    if (this.constructor.name !== 'Herbivore') {
+      Animal.alive.push(this);
+    }
```
EdPirro, line 11: Why avoid `push`ing Herbivores to alive in the `Animal` `constructor` and have it re-implemented in the `Herbivore` `constructor`?

Avoid implementing the same logic in multiple places!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,59 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+
+    if (this.constructor.name !== 'Herbivore') {
```
EdPirro, line 9: Not the best way to check if an object is `instanceof` a class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,59 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+
+    Animal.alive = Animal.alive.map((objAnimal) => {
+      if (objAnimal.name === this.name) {
+        return { ...objAnimal, hidden: !objAnimal.hidden };
+      } else {
+        return objAnimal;
+      }
+    });
```
EdPirro, line 32: `Animal.alive` holds a reference to the animal, every change done to the object will already be accessible through `Animal.alive`; there is no need for this map!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,59 @@
...
+        .map((objAnimal) => {
+          if (objAnimal === animal) {
+            if (objAnimal.health <= 50) {
+              return 'died';
+            }
+            objAnimal.health -= 50;
+          }
+
+          return objAnimal;
+        })
```
EdPirro, line 53: Again, `Animal.alive` has a reference to each `Animal`, changing the instance is enough you don't need this map

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,59 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(animal) {
+    if (animal.constructor.name !== 'Carnivore' && animal.hidden !== true) {
```
EdPirro, line 42: Not the best way to check if an object is `instanceof` a class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,59 @@
...
+      Animal.alive = Animal.alive
```
EdPirro, line 43: Removing the animal from `Animal.alive` is a herbivore's behavior thus, it would make sense to allow the `Herbivore` class handle it and not move into the `Carnivore` class, could you come up with a way to achieve this?

Tip: remember getters and setters

**This is a personal suggestion/challenge, doing it this way is not required**

EdPirro (changes requested):
Great effort @Allanpt! You're almost there! 👏
I pointed out some improvements; please review them thoroughly and, if you have any questions, just let me know!

src/herbivoresAndCarnivores.js
Allanpt: cause i need to add Herbivore hidden key

src/herbivoresAndCarnivores.js
Allanpt: can i just remove this whole map?

EdPirro (approved):
Awesome job, @Allanpt! Keep it up 🚀

PR #2700, opened 2024-04-02
Solution

Andrew77788:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey.hasOwnProperty('hidden') && !prey.hidden) {
+      prey.health -= 50;
+    }
+
+    if (prey.health === 0) {
```
anastasiia-tilikina, line 28: health potentially could be less than 0 let's consider this case

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+      Animal.alive.splice(Animal.alive.indexOf(prey), 1);
```
anastasiia-tilikina, line 29: let's make this logic a static method of Animal class. and call it here
it could be something like

```javascript
Animal.removeDead = (...
```

anastasiia-tilikina (changes requested):
Well done!

let's just make a couple of improvements

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey.hasOwnProperty('hidden') && !prey.hidden) {
```
anastasiiavorobiova, line 24: Consider using instanceOf instead of `hasOwnProperty`

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden) {
+      prey.health -= 50;
+    }
+
+    if (prey.health <= 0) {
+      Animal.removeDead = Animal.alive.splice(Animal.alive.indexOf(prey), 1);
```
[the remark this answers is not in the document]
Andrew77788: де його оголосити, щоб у був доступ до 'prey' ?) не розумію, дякую

PR #2702, opened 2024-04-03
solutionV1

kgGugala:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
```
yevhenii-pyl, line 22: Can be simplified just to a `this.hidden = true;` since the method is called `hide` :)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+      this.hidden = true;
+    } else {
+      this.hidden = false;
+    }
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
yevhenii-pyl, line 32: Same here, we can simplify to a simple check: 

```
      target.health -= 50;
```
and then if the health is below 50 - remove the animal from alive array.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    if (!Animal.alive) {
```
yevhenii-pyl, line 8: Can't we just create the static alive array? :)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+      Animal.alive = [];
+    }
+
```
yevhenii-pyl, line 11: To avoid splicing the alive array when the animal is dead, let's introduce a static method `removeFromAlive` or something like that, this should incapsulate this logic within the Animal class.

yevhenii-pyl (changes requested):
Great job! Let's refactor a bit and we are good to go!

yevhenii-pyl (approved):
You got it! 👍

PR #2711, opened 2024-04-05
js_herbivores_and_carnivores

ivanKo95:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+    }
+
+    if (herbivore.health <= 0) {
+      const index = Animal.alive.indexOf(herbivore);
+
+      if (index > -1) {
+        Animal.alive.splice(index, 1);
+      }
```
loralevitska, line 34: Use filter method here

loralevitska (changes requested):
Almost done) Let's improve

src/herbivoresAndCarnivores.js
ivanKo95: good catch! done)

Viktor-Kost (approved):

PR #2715, opened 2024-04-08
task solution

MatheusStefan:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+    if (prey instanceof Herbivore) {
+      if (!prey.hidden) {
+        prey.health -= 50;
+      }
+
+      if (prey.health <= 0) {
+        const index = Animal.alive.indexOf(prey);
+
+        if (index !== -1) {
+          Animal.alive.splice(index, 1);
```
jv-aquino, line 41: This verification and removal of the Animal (prey) should be added to the animal class as a better practice: think about it, if you needed to make another animal die / check it, you would need to reuse this whole code

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,50 @@
...
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, hidden = false) {
+    super(100, name, hidden);
+  }
+
+  hide() {
```
jv-aquino, line 19: a nice thing to add would be an unhide function

jv-aquino (changes requested):
Nice job Matheus, clean code and I liked the way you set the default params! I left some improvements that could be made, hope you like it!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,60 @@
...
+  static alive = [];
+
+  constructor(health = 100, name, hidden = false) {
+    this.health = health;
+    this.name = name;
+    this.hidden = hidden;
+    Animal.alive.push(this);
+  }
+
+  die() {
```
jv-aquino, line 13: i think checkDead() would be a better naming, like, die() implies that it's making the animal die

you may want to read these [short naming conventions](https://www.syncfusion.com/blogs/post/top-javascript-naming-convention)

jv-aquino (approved):
Great job! Your naming conventions are kinda nice, I just suggested some small changes to make them better, keep up your good work!

PR #2721, opened 2024-04-10
solution

myemets:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden !== true && herbivore instanceof Herbivore) {
```
anastasiia-tilikina, line 31:
```suggestion
    if (!herbivore.hidden && herbivore instanceof Herbivore) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
anastasiia-tilikina, line 30:
```suggestion
  bite(beast) {
```
I would rather choose another name for this param since we don't know exactly if it is herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+
+    Animal.alive.push(this);
+  }
+
+  static checkDead() {
```
anastasiia-tilikina, line 13:
```suggestion
  static removeDead() {
```
since we are not just checking but removing

VitaliyBondarenko1982 (approved):
Good job.

PR #2732, opened 2024-04-12
Solution

odunchyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
alexander-ignatow, line 6: it would be great to use constants for such numbers

```suggestion
  static DEFAULT_HEALTH = 100;

  constructor(name, health = DEFAULT_HEALTH) {
```

alexander-ignatow (approved):
Good! 👍

PR #2733, opened 2024-04-12
solution

ArtemZakhar:
Hi)
Have a nice day!)

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,54 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+  }
+
+  getAnimalHealth(name) {
+    return Animal.alive.find((item) => item.name === name).health;
+  }
+
+  getBited() {
+    Animal.alive = Animal.alive.filter((item) => item.health !== 0);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+
+    Animal.alive.push(this);
+  }
+
+  bitten() {
+    this.health -= 50;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    Animal.alive.push(this);
+  }
+
+  bite(victim) {
+    if (victim.hidden || victim instanceof Carnivore) {
+      return undefined;
+    }
+
+    victim.bitten();
+    this.getBited();
+  }
 }
 
 module.exports = {


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+
+    Animal.alive.push(this);
```
alexander-ignatow, line 25: This should happen in the constructor of an `Animal`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    Animal.alive.push(this);
```
alexander-ignatow, line 41: This should happen in the constructor of an Animal - or you will have to duplicate it for each `Animal` subclass

alexander-ignatow (changes requested):
Good solution, but try not to over-complicate or over-engineer your code - keep it simple

In addition - mind the naming of the functions

src/herbivoresAndCarnivores.js
ArtemZakhar: This is a problem for me.
Because I have no idea how to send hidden property from Herbivore to Animal.
Please advise)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
+  }
+
+  bite(victim) {
+    if (victim.hidden || victim instanceof Carnivore) {
+      return undefined;
```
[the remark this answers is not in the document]
ArtemZakhar: Linter won't let me return nothing.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+  }
+
+  getAnimalHealth(name) {
```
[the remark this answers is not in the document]
ArtemZakhar: Have no idea) This task was very hard for me) 
I changed my code a bunch of times. 
Even now, when I open after your comment I can't remember the purpose of this method)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
+    return Animal.alive.find((item) => item.name === name).health;
+  }
+
+  getBited() {
```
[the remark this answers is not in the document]
ArtemZakhar: Good point. Thank you)
Firstly I used reduce to control bites and death. After refactoring name was left the same but it has to be changed!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  bitten() {
```
undeletable, line 21: Better naming for this method would be `bite`.
Method name should usually be an answer to 'What does it do?' question. Here it answers 'What's the result after its execution?'

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    this.health -= 50;
```
undeletable, line 22: It's better to define 50 as constant as well

undeletable (approved):
Well done! There are just a couple of small improvements left, so I'm approving.

src/herbivoresAndCarnivores.js
ArtemZakhar: I have bite method in Carnivore class. 
In this case, the animal is bitten, so I suppose the name should be healthReduce.

Thank you!

undeletable: Suggesting `bite` was my mistake (herbivore receives bites, but doesn't bite), so I've edited my comment with more appropriate names. Anyway, I see from your reply that you've got the point.

PR #2734, opened 2024-04-12
add task solution

PrytulaAlexandr:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+
+    this.health = Animal.FULL_HEALTH;
+
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  static HIDDEN = false;
```
mrkovaliv, line 17: you don't need make `HIDDEN`  like a static field

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+
+    this.health = Carnivore.FULL_HEALTH;
```
mrkovaliv, line 35: you don't need `this.health = Carnivore.FULL_HEALTH;` because you do this actions in parent class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  static HIDDEN = false;
+
+  constructor(name) {
+    super(name);
+
+    this.health = Herbivore.FULL_HEALTH;
```
mrkovaliv, line 22: also don't need

mrkovaliv (changes requested):
Nice work 👍 👍 👍 

But we can improve this code)))

src/herbivoresAndCarnivores.js
PrytulaAlexandr: Hi Volodymyr! Deleted static.

src/herbivoresAndCarnivores.js
PrytulaAlexandr: Delete this.health = Carnivore.FULL_HEALTH;

src/herbivoresAndCarnivores.js
PrytulaAlexandr: Delete constructor.

PrytulaAlexandr:
Thanks for check, wish you good weekends))

mrkovaliv (approved):
Great work 🚀 🚀 

Approved 🏁 

Have a nice weekend ☺️

PR #2735, opened 2024-04-12
tests Ok.

sevenlays:
Help me make it better, please:)

Can i make it without lines 14, 15, 16, 17??

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health) {
+    super(name, health);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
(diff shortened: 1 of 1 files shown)


undeletable:
Lines 14-17 are OK, because `hidden` property only makes sense for `Herbivore` class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
undeletable, line 6: It makes sense to define 100 as constant.
Same note for 50 inside `Carnivore.bite()`.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
```
undeletable, line 27: I'd advise to define logic related to `herbivore.health` inside `Herbivore` class. You can add method like `receiveBite` or `reduceHealth` which reduced `health` by 50. Also, you can use getter and setter for `health` property - in setter you can modify `Animal.alive` value.

undeletable (approved):
Good solution, just a couple of improvements would make sense

PR #2740, opened 2024-04-13
Solution

Voloshyn-Oleg:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  static animals() {
+    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super();
+    this.name = name;
```
alexander-ignatow, line 17: Name should be assigned in the base class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
-
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super();
+    this.name = name;
+  }
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
```
alexander-ignatow, line 31: Makes sense to make `50` a constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(health = 100) {
```
alexander-ignatow, line 5: Probably `100` should be a constant

alexander-ignatow (changes requested):
Good solution, but some changes are required

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(health) {
```
undeletable, line 5: Constructor method of `Animal` class should take `name` parameter as well, and set the corresponding property

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    const FULL_HEALTH = 100;
+
+    this.health = health;
+
+    this.health = FULL_HEALTH;
```
undeletable, line 10: The logic related to health is not correct: first you set `this.health` to value from argument, and then rewrite it with default value. Instead of that, you should just define `FULL_HEALTH` as default value of function parameter.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    Animal.alive.push(this);
+  }
+  static animals() {
```
undeletable, line 13: This method is not needed, because you already handle this in `bite()` method of `Carnivore` class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  }
 }
-
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super();
+  }
+  bite(target) {
+    const BITE_CARNIVORE = 50;
```
undeletable, line 32: `BITE_CARNIVORE` is not the best naming for this constant. I'd suggest to name it something like `HEALTH_REDUCTION_ON_BITE`, so that it clearly describes the meaning of constant.

undeletable (changes requested):
The solution is fine, just please take care of the commented places.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.hidden = false;
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
-
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
```
mrkovaliv, line 24: you don't need this constructor here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    super();
+  }
+  bite(target) {
+    const HEALTH_REDUCTION_ON_BITE = 50;
```
mrkovaliv, line 28: better move on outside the class

mrkovaliv (approved):
Nice work 👍 👍 👍 

Approved 🏁

PR #2741, opened 2024-04-13
add task solution

JohnyJDX:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.name = name;
```
undeletable, line 17: There's no need to initialize `this.name` in `Herbivore`'s constructor, because it's already done in constructor of parent class.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
undeletable, line 6: It makes sense to define default health value as constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.name = name;
+  }
+
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
```
undeletable, line 33: I'd advise to handle damage from bite inside `Herbivore` class. Health reduction by 50 is the feature of herbivores, so it's better to define such feature inside their class. Here's how it can be implemented:
```
  ...
  receiveBite() {
    if (!this.hidden) {
      this.health -= 50;
      if (this.health >= 0) {
        Animal.alive = ...
      }
    }
  }
}
```
This would make classes design clearer and easier to maintain

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+      victim.health -= 50;
+
+      Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```
undeletable, line 36: There's no need to change `Animal.alive` value on each bite. We can do it only when herbivore's health reduces to 0. In the comment above I've provided an example for that.

undeletable (changes requested):
The soluton looks fine, I'd just ask for some improvements. See the comments to the code.

alexander-ignatow (approved):
Very good!

PR #2742, opened 2024-04-14
Solution

MarchAlbion:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
 
 class Animal {
   // write your code here
+  static alive = [];
+
+  constructor(name, health) {
+    this.name = name;
+    this.health = health || 100;
+    Animal.alive.push(this);
+  }
```
alexander-ignatow, line 11: It is better to use default value for the health

```suggestion
  constructor(name, health = 100) {
    this.name = name;
    this.health = health;
    Animal.alive.push(this);
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+  static deadAnimal() {
```
alexander-ignatow, line 12: Please fix the naming
```suggestion
  static filterOutDeadAnimals() {
```

```suggestion
  static removeEatenAnimals() {
```

etc

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
+  }
 }
 
 class Herbivore extends Animal {
   // write your code here
+  constructor(health, name, hidden) {
+    super(health, name);
+
+    this.hidden = hidden || false;
```
alexander-ignatow, line 22: As above - prefer default value (it gives the solution with the better readability)

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+  }
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(herb) {
```
alexander-ignatow, line 31: the parameter name is misleading

a) you are not guaranteed to get `Herbivore`
b) `herb` is a plant, not an animal :)

```suggestion
  bite(animal) {
```

```suggestion
  bite(other) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
+    }
+
+    if (herb.health < 50) {
```
alexander-ignatow, line 36: This looks weird

why not
```suggestion
    if (herb.health <= 0) {
```
?

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(herb) {
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
```
alexander-ignatow, line 33: Makes sense to make `50` constant

alexander-ignatow (changes requested):
Good!

The solution works, but there are some issues to fix and to make the solution better

mrkovaliv (approved):

src/herbivoresAndCarnivores.js
mrkovaliv: missed comment

PR #2743, opened 2024-04-14
Solution

GrigoryyKach:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static deadAnimal(animal) {
```
alexander-ignatow, line 12: Please, fix the method name

```suggestion
  static removeEaten(animal) {
```

```suggestion
  static removeDeadAnimal(animal) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,52 @@
...
-  // write your code here
+  bite(target) {
+    if (
+      target instanceof Carnivore ||
+      (target instanceof Herbivore && target.hidden)
+    ) {
+      return;
+    }
+
+    target.health -= 50;
```
alexander-ignatow, line 44: It makes sense to make 50 the constant

alexander-ignatow (changes requested):
Good, but some minor changes are requierd

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
 'use strict';
 
+const damage = 50;
```
alexander-ignatow, line 3: Good, but in this case it is better to make the constant a part of the `Carnivore` class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,54 @@
...
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static removeDeadAnimal(animal) {
+    // Variant with filter:
+    // Animal.alive = Animal.alive.filter(a => a !== animal);
+
```
alexander-ignatow, line 17: It is best not to do so

If you don't need the code - remove it

alexander-ignatow (approved):
Good! Lets go with it, but please consider my comments for the future

PR #2744, opened 2024-04-14
Solution

TetianaVeremchuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
undeletable, line 19: If `this.hidden` is `false`, `hide()` method would make it `true`. This is not what we expect from the function called `hide()`: hide action means that as result animal would become hidden.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
undeletable, line 6: It makes sense to define default health value as constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
+      herbivore.health -= 50;
```
undeletable, line 26: It makes sense to define default health reduction value as constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
undeletable, line 24: It's better to rename parameter from `herbivore` to something general (for example, `animal`). Generally speaking, the passed value is not always a herbivore.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    if (herbivore instanceof Herbivore && !herbivore.hidden) {
```
undeletable, line 25: I'd advise to handle damage from bite inside `Herbivore` class. Health reduction by 50 is the feature of herbivores, so it's better to define such feature inside their class. Here's how it can be implemented:
```
  ...
  receiveBite() {
    if (!this.hidden) {
      this.health -= 50;
      if (this.health <= 0) {
        Animal.alive = ...
      }
    }
  }
}

  ...
  bite(animal) {
    if (animal instanceof Herbivore) {
      animal.receiveBite();
    }
  }
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 }
+Animal.alive = [];
```
undeletable, line 10: It's better to define `alive` inside `Animal` class as static property. This would help you to be sure that `Animal.alive` is always initialized before it's used.

undeletable (changes requested):
The soultion looks fine, but it needs some improvements. See my comments to the code.

mrkovaliv (approved):

PR #2745, opened 2024-04-14
Solution

OTsiolkovskiy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (!victim.hidden && victim instanceof Herbivore) {
+      victim.health -= 50;
```
alexander-ignatow, line 25: It is better to make `50` a constant

alexander-ignatow (approved):
Good 👍 

Please, see the comments

PR #2747, opened 2024-04-15
solution

Esther-Guimaraes:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+      Animal.alive = Animal.alive.filter((el) => el.health > 0);
+    }
```
jv-aquino, line 29: This verification here should be done elsewhere, don't you think? If we want to reuse this check to see if the animal has died for, for example, poison, we would have to copy the entire code again - which is not a best practice

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
```
jv-aquino, line 19: it's also cool to have an unhide function (or else the animal will be forever hidden)

jv-aquino (changes requested):
Good job, there are some small changes that will improve it!

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+      Animal.updateAlives();
+    }
```
EdPirro, line 37: The bite handling is a Herbivore behavior and therefore, it would be considered a best practice to implement it inside the `Herbivore` class.
As an optional task, could you think of a way to move this logic into the Herbivore class?

EdPirro (approved):
Great job, @Esther-Guimaraes! Keep it up! 🚀

I pointed out a personal (optional) suggestion that I think would improve your solution. If you have any questions about it, just let me know!

PR #2749, opened 2024-04-15
solution

denyssheremeta:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    const MAX_HEALTH = 100;
+
```
alexander-ignatow, line 6: please declare MAX_HEALTH outside of the constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
+  }
```
alexander-ignatow, line 22: If the animal is hidden, `hide` method unhides it.

You need either to fix the method name (ex: `toggleHidden`, etc) or change the method logic (per task, you don't need to unhide animal)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
```
alexander-ignatow, line 28: Why not make `50` a constant?

alexander-ignatow (changes requested):
Good, but some minor changes are needed

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    const DAMAGE_AMOUNT = 50;
```
alexander-ignatow, line 27: Good, but `DAMAGE_AMOUNT` should be the property of `Carnivore` (not local variable inside the method)

alexander-ignatow (approved):
Looks better now, but please consider my comment for the future

PR #2750, opened 2024-04-15
Solution

AYupko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
 'use strict';
-
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
undeletable, line 5: It makes sense to define default `health` value as constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden === false) {
```
undeletable, line 25: I'd advise to handle damage from bite inside `Herbivore` class. Health reduction by 50 is the feature of herbivores, so it's better to define such feature inside their class. Here's how it can be implemented:
```
  ...
  receiveBite() {
    if (!this.hidden) {
      this.health -= 50;
      if (this.health >= 0) {
        Animal.alive = ...
      }
    }
  }
}

  ...
  bite(target) {
    if (target instanceof Herbivore) {
      target.receiveBite();
    }
  }
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+      target.health -= 50;
+    }
+
+    Animal.alive = Animal.alive.filter((creature) => creature.health > 0);
```
undeletable, line 29: There's no need to change `Animal.alive` value on each bite. We can do it only when herbivore's health reduces to 0. In the comment above I've provided an example for that.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden === false) {
+      target.health -= 50;
```
undeletable, line 26: It makes sense to define health reduction value as constant

undeletable (changes requested):
The solution is fine. Just please add some improvements.

mrkovaliv (approved):

PR #2751, opened 2024-04-15
solution

Olek-91:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
```
alexander-ignatow, line 27: It makes sense to make `50` a constant too

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.health = 100;
```
alexander-ignatow, line 7: It makes sense to make `100` a constant

alexander-ignatow (approved):
Good, but see my comments below

PR #2752, opened 2024-04-15
Solution

lmmaarriiaa:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+  }
+
+  static alive = [];
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
```
alexander-ignatow, line 15: You should do it in the constructor of the base class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
```
alexander-ignatow, line 32: You should do it in the constructor of the base class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+
+  hidden = false;
+
+  hide() {
+    if (this.hidden === true) {
+      this.hidden = false;
+    }
+
+    this.hidden = true;
+  }
```
alexander-ignatow, line 26: This method is misleading, as it can also unhide hidden animals (that is opposite to `hide`)
You need to either rename it or to change method logic (so it will only hide the animal)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
```
alexander-ignatow, line 37: It will be nice to make `50` the constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+
+      if (animal.health === 0) {
```
alexander-ignatow, line 39: What if `animal.health` somehow becomes *less* than 0?

alexander-ignatow (approved):
Good, but lets improve your solution!

src/herbivoresAndCarnivores.js
lmmaarriiaa: Thank you! I wasn`t sure about this one

PR #2756, opened 2024-04-15
Solution

dyakym:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100, hidden = false) {
+    this.name = name;
+    this.health = health;
+    this.hidden = hidden;
```
alexander-ignatow, line 8: Not sure, if you need this here, as according to the task `hiding` is something specific to `Herbivores`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    const animalIndex = Animal.alive.indexOf(target);
```
alexander-ignatow, line 25: You don't need `animalIndex` if the animal cannot be bitten

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+    if (target.hidden !== true && target instanceof Herbivore) {
+      target.health -= 50;
```
alexander-ignatow, line 28: It makes sense to make `50` a constant

alexander-ignatow (changes requested):
Good, but you need to make some changes to your code

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
```
mrkovaliv, line 5: move on 100 to `const FULL_HEALTH = 100;`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.hidden = hidden;
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  damage = 50;
```
mrkovaliv, line 23: also move to to constants

mrkovaliv (approved):
Nice work 👍 👍 👍

PR #2757, opened 2024-04-15
Develop

ordila:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
```
alexander-ignatow, line 30: makes sense to make `50` a constant

alexander-ignatow (approved):
Good, but please check my comment for the future

PR #2759, opened 2024-04-15
Solution

ropanefarias:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (
+      animal instanceof Carnivore ||
+      (animal.hidden && animal instanceof Herbivore)
```
EdPirro, line 29: Misleading conditional; this validation would allow any Animal that is not a `Carnivore`  nor a `Herbivore` through in this case, where an Animal can be only either a `Carnivore` or a `Herbivore` it works (which makes the `animal instanceof Herbivore` check redundant) but it would fail in a more generic scenario.

Bite should work if the `target is a Herbivore and is not hidden`; there is a better way to write this

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    ) {
+      return;
+    }
+    animal.health -= 50;
```
EdPirro, line 33: Additionally, the bite handling is a Herbivore behavior and therefore, it would be considered a best practice to implement it inside the `Herbivore` class.
As an **optional** task, could you think of a way to move this logic into the Herbivore class?

EdPirro (changes requested):
Great job, @ropanefarias! 👏

I'll just ask you to review the `bite` method conditional, it is slightly more complicated than it must be.

Also, I pointed out an additional (and optional) task, if you feel like taking it on and have any questions, just let me know!

src/herbivoresAndCarnivores.js
ropanefarias: I don't think I understand, Edu! From what I understand from the exercise, the bite is a carnivore method: "Carnivore has a `bite` method, which takes a herbivore object and decreases the object's health by `50`. The method does not work if it is another сarnivore, or the herbivore is currently hiding.". That's why I put it in the class of carnivores, not herbivores. The bite is from carnivore, not herbivore.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Herbivore) || animal.hidden) {
+      return;
+    }
+    animal.health -= 50;
+
+    if (animal.health <= 0) {
```
jv-aquino, line 32: As a better practice this part here of removing health or/and checking if the animal is dead should probably be done inside the Animal class or herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(health, name, hidden) {
+    super(health, name);
+    this.hidden = false;
+  }
+  hide() {
```
jv-aquino, line 20: optional: an unhide function would also be cool to not make it hidden permanently

jv-aquino (changes requested):
Good job Rodrigo, I will just suggest two more things to make your code great - try also to reflect on them and think to yourself why this is a better practice (e.g. what if we wanted to create new instances of animals that also need to die or take damage?)

jv-aquino (changes requested):
(platform issues)

PR #2762, opened 2024-04-16
add task solution

tarassamchenko:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  static DEFAULT_HEALTH_POINTS = 100;
```
mrkovaliv, line 5: you don't need to make these constans like static fields

just 

```
const DEFAULT_HEALTH_POINTS = 100; 

class Animal {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  static BITE_DAMAGE = 50;
+  static DEFAULT_STEALTH = false;
```
mrkovaliv, line 7: rename this variable

mrkovaliv (changes requested):
Nice work 👍 

Added some improvements)))

src/herbivoresAndCarnivores.js
tarassamchenko: I did that, but "npm run lint" didn't pass, I'll try again

tarassamchenko: I try it now and it works. But why it didn`t work before?(

mrkovaliv: may be you use this variant:


```
class Animal {

const DEFAULT_HEALTH_POINTS = 100; 
```
this one isn't ok))

mrkovaliv (approved):

mrkovaliv:
🚀 🚀 🚀

src/herbivoresAndCarnivores.js
tarassamchenko: Ні , коли ти відхилив завдання , я просто його відкотив до версії яка не проходила тести та перейменував їх

PR #2763, opened 2024-04-16
Animal solution

ilyasolo881:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
+      animal.health -= 50;
+      Animal.removeAnimal();
```
undeletable, line 32: There's no need to launch this method after every bite. It's only needed when an animal's health is not greater than 0.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
undeletable, line 6: It makes sense to define default `health` value as constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
+      animal.health -= 50;
```
undeletable, line 31: It makes sense to define value of health reduction as constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.constructor.alive.push(this);
+  }
+
+  static removeAnimal() {
```
undeletable, line 12: More appropriate naming for this method would be `removeDeadAnimals` or smth like that. `removeAnimal` name implies that you remove a single animal - which is not exactly correct.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
```
undeletable, line 30: I'd advise to handle damage from bite inside `Herbivore` class. Health reduction by 50 is the feature of herbivores, so it's better to define such feature inside their class. Here's how it can be implemented:
```
  ...
  receiveBite() {
    if (!this.hidden) {
      this.health -= 50;
      if (this.health <= 0) {
        Animal.alive = ...
      }
    }
  }
}

  ...
  bite(animal) {
    if (animal instanceof Herbivore) {
      animal.receiveBite();
    }
  }
}
```

undeletable, line 30: I'd suggest to check if animal is a herbivore - instead of checking if it's not a carnivore. According to task conditions, this is the feature of herbivores - not everyone except carnivores.
This is the same when you only have two classes, but the difference appears when potentially third class is added.

undeletable (changes requested):
The soultion is fine, I'd just like to suggest several improvements

src/herbivoresAndCarnivores.js
```diff
@@ -1,35 +1,43 @@
...
   }
 
   hide() {
     this.hidden = true;
   }
+
+  receiveBite() {
+    if (!this.hidden) {
+      this.health -= Herbivore.DAMAGE_FROM_BITE;
+      Animal.removeDeadAnimals();
```
undeletable, line 32: We can check if `this.health <= 0` and call `Animal.removeDeadAnimals()` only if that's true - to avoid redundant calls of that method

undeletable (approved):
The solution overall is fine, so I'm approving. Just please check a small comment and take that into account in similar cases.

PR #2766, opened 2024-04-17
add ecosystem

vinasxz:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
+
+      if (target.health <= 0) {
+        target.die();
+      }
```
EdPirro, line 39: The bite handling is a Herbivore behavior and therefore, it would be considered a best practice to implement it inside the `Herbivore` class.
As an **optional** task, could you think of a way to move this logic into the Herbivore class?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
+
+  hide() {
+    this.hidden = true;
+  }
```
EdPirro, line 29: Would be nice to have a `unhide` function to allow a `Herbivore` to become visible again

EdPirro (approved):
Great job, @vinasxz! Keep it up! 🚀

I pointed out two personal (optional) suggestion that I think would improve your solution. If you have any questions about it, just let me know!

PR #2768, opened 2024-04-20
Herbivores and Сarnivores

Yaroslav65:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  static addElement(carnivore) {
+    Animal.alive.push(carnivore);
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore) {
+      if (!animal.hidden) {
```
nazarmatsevych, line 35:
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

We can combine these conditions

nazarmatsevych (approved):
Good work, please check the comment below

src/herbivoresAndCarnivores.js
Yaroslav65: Done!))

PR #2770, opened 2024-04-20
added solution

SlavaRevit:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    Animal.alive.push(this);
+  }
+  bite(obj) {
+    if (!(obj instanceof Carnivore || obj.hidden)) {
+      obj.health -= 50;
+
+      if (obj.health <= 0) {
+        const indexOfObjectElement = Animal.alive.indexOf(obj);
+
+        Animal.alive.splice(indexOfObjectElement, 1);
```
nazarmatsevych, line 34: Use filter method method instead

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
+  bite(obj) {
+    if (!(obj instanceof Carnivore || obj.hidden)) {
```
nazarmatsevych, line 28:
```suggestion
    if (obj instanceof Herbivore && !obj.hidden) {
```

nazarmatsevych (changes requested):
Almost done, please fix the comments below

l4st1m0za (approved):
Good job!

PR #2772, opened 2024-04-22
add solution

stepanyukvova200:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  alive() {
```
undeletable, line 12: Looks like this method is not needed

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
undeletable, line 8: It makes sense to define default health value as constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && !beast.hidden) {
+      beast.health -= 50;
```
undeletable, line 41: It makes sense to define health reduction value as constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(beast) {
+    if (beast instanceof Herbivore && !beast.hidden) {
```
undeletable, line 40: I'd advise to handle damage from bite inside `Herbivore` class. Health reduction by 50 is the feature of herbivores, so it's better to define such feature inside their class. Here's how it can be implemented:
```
  ...
  receiveBite() {
    if (!this.hidden) {
      this.health -= 50;
      if (this.health <= 0) {
        Animal.alive = ...
      }
    }
  }
}

  ...
  bite(beast) {
    if (beast instanceof Herbivore) {
      beast.receiveBite();
    }
  }
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+      beast.health -= 50;
+      Carnivore.removeDeadAnimals();
```
undeletable, line 42: There's no need to launch this method after every bite. It's only needed when an animal's health is not greater than 0.

undeletable (changes requested):
The solutionooks good, just a couple of improvements make sense. Please take care of comments.

alexander-ignatow (changes requested):
Looks much better!

But you forgot one minor thing

polosanya (approved):
Well done!

PR #2787, opened 2024-04-25
add task solution

danzaris:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (!(target instanceof Herbivore) || target.hidden) {
+      return;
+    }
+    target.health -= 50;
+    Animal.removeDead();
```
```suggestion
    if (target instanceof Herbivore && !target.hidden) {
      target.health -= 50;
      Animal.removeDead();
    }
```

We can simplify this condition

nazarmatsevych (approved):
Good work!
Please check the comment below

PR #2789, opened 2024-04-26
added Animal, Herbivore and Carnivore classes

yuriipodria:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,33 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        delete Animal.alive[Animal.alive.indexOf(animal)];
```
l4st1m0za, line 27: Using `delete` on an array leaves holes in the array. This can lead to incorrect indexing. It's more efficient to use the `splice` method to remove elements from an array.

l4st1m0za (approved):
Good job!

PR #2799, opened 2024-05-01
add task solution

pakhomovalex:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,43 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor() {
+    super();
+    Animal.alive.push(this);
+  }
+
+  bite(animal) {
+    if (!(animal.hidden || animal instanceof Carnivore)) {
```
```suggestion
    if (!animal.hidden && animal instanceof Herbivore) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,43 @@
...
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
nazarmatsevych, line 38: Use filter method instead:

nazarmatsevych (changes requested):
Almost dome, please check the comments below

l4st1m0za (approved):
Good job!

PR #2808, opened 2024-05-08
add task solution

ArtemDemeniev:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
 'use strict';
-
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
undeletable, line 5: It's better to define default health value as constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,48 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(victim) {
+    if (victim instanceof Carnivore || victim.hidden) {
+      return;
+    }
+    victim.decreaseHealth(50);
```
undeletable, line 44: I'd advise to move details of reaction to bite inside of `Herbivore` class. So that if you want to change herbivores' reaction to bite, you go and make this change inside `Herbivore` class, not `Carnivore` one. Health reduction by 50 is herbivores feature, so let's keep it in their class. It could be something like this:
```
  ...
  receiveBite() {
    if (!this.hidden) {
      this.decreaseHealth(HERBIVORE_HEALTH_DECREASE); // also define 50 as constant
    }
  }
}
...
  ...
  bite(victim) {
    if (victim instanceof Carnivore) {
      return;
    }
    victim.receiveBite();
  }
}
```

undeletable (changes requested):
The solution looks fine. I just suggest a couple of improvements for your code.

PR #2814, opened 2024-05-10
added solution

syavaYki:


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,76 @@
...
+panther.bite(deer);
+rabbit.hide();
+panther.bite(rabbit);
+
+// Animal.alive ===
+//   [
+//     { name: 'Bagira', health: 100 },
+//     { name: 'King', health: 100 },
+//     { name: 'Max', health: 100, hidden: true },
+//   ];
```
vadiimvooo, line 72:
```suggestion

```

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,76 @@
...
-  // write your code here
+  static alive = [];
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  bitten() {
+    if (this.health - 50 > 0) {
```
vadiimvooo, line 12:
```suggestion
    if (this.health - 50) {
```

vadiimvooo: Also it would be nice to store the hp value in some const and use it whenever you need it.

vadiimvooo (changes requested):
Almost done. Just a few comments to fix.

vadiimvooo (approved):
Well done!)

PR #2827, opened 2024-05-17
Solution

SylwiaNG:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
DmytroOleksiiuk, line 6: It's not needed to pass health to constructor, it's 100 by default when a new Animal is created

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (!(target instanceof Herbivore) || target.hidden) {
+      return;
+    }
+    target.health -= 50;
+    Animal.removeDeadAnimals();
```
DmytroOleksiiuk, line 34: if (target.health <= 0) {
      Animal.removeDeadAnimals();
    }

DmytroOleksiiuk (changes requested):
Good start, please check the comments

EwaKobiela (approved):

PR #2828, opened 2024-05-18
add solution

Nazarin565:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden !== true) {
```
```suggestion
    if (animal instanceof Herbivore && !animal.hidden) {
```

Same but a bit shorter

nazarmatsevych (approved):
Great work!

PR #2835, opened 2024-05-21
Solution

BagietaCyganeta:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,57 @@
...
+
+  die() {
+    this.alive = false;
+
+    const index = Animal.alive.indexOf(this);
+
+    if (index > -1) {
+      Animal.alive.splice(index, 1);
+    }
+  }
```
BogdanMaliuta, line 21: You dont need this method, just add this logic to your checkHealth method

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,57 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.alive = true;
```
BogdanMaliuta, line 9: also you don`t need this because this redundant, you just setting it false when animal dies

BogdanMaliuta (changes requested):
Not bad solution, but here is some redundant things

PR #2841, opened 2024-05-27
Solution

yuliaPel:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    super();
+  }
+
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.alive = Animal.alive.filter(
+          (aliveAnimal) => aliveAnimal !== animal,
```
dinaminahi, line 35: We can't compare aliveAnimal and animal as they are both objects, so we are comparing their references 
Let's filter alive by their health

dinaminahi (changes requested):
Good job 👍 
Let's do some fixes

polosanya (approved):
Well done!

PR #2851, opened 2024-06-01
herbivores and carnivores model

kozlovoleksii:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,30 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
+    if (obj instanceof Herbivore && !obj.hidden) {
+      obj.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```
SanyaBratashchuk, line 26:
```suggestion
    if (victim,health <= 0) {
      Animal.alive = Animal.alive.filter(animal => animal.name !== victim.name)
    }
```

SanyaBratashchuk (changes requested):

src/herbivoresAndCarnivores.js
kozlovoleksii: if I make such edits, I fail the test.... 

image

SanyaBratashchuk: Oh, my bad didn't take into account that name isn't unique

SanyaBratashchuk (approved):
LGTM

PR #2857, opened 2024-06-06
js_herbivores_and_carnivores

LitviVit:


roman-mirzoian (approved):
Good job. Also recommend using constants for health and bite values.

PR #2860, opened 2024-06-06
Solution for the Herbivores and Carnivores task

lesdyman:
Hi there! Here is my solution for the Herbivores and Carnivores task. Hope I've done everything as it is meant to be done. But please, do not hesitate to let me know if there is anything that could be done better.

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  health = 100;
+
+  constructor(name) {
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  health = 100;
```
roman-mirzoian, line 5: Example of using a constant:
```suggestion
  health = MAX_HEALTH;
```

roman-mirzoian (approved):

PR #2861, opened 2024-06-06
Solution

demkoveronika:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
```
Example: 
```
const MAX_HEALTH = 100;
const BITE_DAMAGE = 50;
const DEAD_VALUE = 0;
```

roman-mirzoian (approved):

PR #2862, opened 2024-06-06
Solution

OSpyrydonov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
```
roman-mirzoian, line 2: Using constants for health and bite values ​​will help avoid magic numbers and improve code readability.

Example:
```suggestion
const MAX_HEALTH = 100;
const BITE_DAMAGE = 50;
const DEAD_VALUE = 0;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static animalDead() {
```
roman-mirzoian, line 12: It is better to change the name of the method `animalDead` to `removeDeadAnimals` for better clarity.

roman-mirzoian (approved):
Well done and please pay attention to the comments.

PR #2864, opened 2024-06-06
solution

yaroslav1177:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
```
roman-mirzoian, line 2: Using constants for health and bite values ​​will help avoid magic numbers and improve code readability.
```
const MAX_HEALTH = 100;
const BITE_DAMAGE = 50;
const DEAD_VALUE = 0;
```

roman-mirzoian (approved):
Well done!

PR #2865, opened 2024-06-06
add solution

RuslanHnatyshyn:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    Animal.alive = Animal.alive.filter((animal) => animal !== this);
+  }
```
roman-mirzoian, line 14: The `die` method in the Animal class filters the list of live animals by the health of each instance. 
This works, but depending on this method in the Herbivore class can complicate the code. 

Better to have this method at the Animal class level to avoid such problems.
Example: 
```suggestion
  static removeDeadAnimals() {
    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
```
Example:
```
const MAX_HEALTH = 100;
const BITE_DAMAGE = 50;
const DEAD_VALUE = 0;
```

roman-mirzoian (approved):
Well done and please pay attention to the comments.

src/herbivoresAndCarnivores.js
RuslanHnatyshyn: if i added constants the test will be false.

RuslanHnatyshyn:
Thank you for your comments.

PR #2868, opened 2024-06-07
add task solution

Stasandrushchak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,32 @@
 'use strict';
```
roman-mirzoian, line 2: For improvement, you can use constants to avoid magic numbers.
Example: 
```
const MAX_HEALTH = 100;
const BITE_DAMAGE = 50;
const DEAD_VALUE = 0;

```

roman-mirzoian (approved):

PR #2869, opened 2024-06-07
task solution

seriuksergii:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
 
+  static dellete() {
+    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
+  }
+}
```
roman-mirzoian, line 15: Here is better to use a more descriptive name for this function.
Example:
```suggestion
  static deleteDeadAnimals() {
    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
  }
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
```
Example:

```
const MAX_HEALTH = 100;
const BITE_DAMAGE = 50;
const DEAD_VALUE = 0;
```

roman-mirzoian (changes requested):
Good work, here are the comments that will improve your code

seriuksergii:
Hi, I fixed my mistakes.

roman-mirzoian (approved):
Well done!

PR #2871, opened 2024-06-07
Solution

pushyshyn:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
```
roman-mirzoian, line 11: Here is an example of a separate method for removing dead animals, which improves readability and logical structure of the code.

```suggestion
    Animal.alive.push(this);
  }
  static removeDeadAnimals() {
    Animal.alive = Animal.alive.filter(animal => animal.health > 0);
  }
}
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
```

Example:
```
const MAX_HEALTH = 100;
const BITE_DAMAGE = 50;
const DEAD_VALUE = 0;
```

roman-mirzoian (changes requested):
Good work, here are the comments that will improve your code.

roman-mirzoian (approved):
Well done!

PR #2876, opened 2024-06-07
add carnivores and herbivores solution

FaiHamid:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= 50;
+
+      if (victim.health <= 0) {
+        const index = Animal.alive.indexOf(victim);
+
+        Animal.alive.splice(index, 1);
+      }
```
roman-mirzoian, line 37: I recommend improving the method of removing dead animals from the massif.
let's put this in a separate static method, changing this logic to remove dead animals from the Animal.alive array using filtering:

```
static removeDeadAnimals() {
}
```
This method is called inside the `bite` method:
```suggestion
      Animal.removeDeadAnimals();
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
```
roman-mirzoian, line 2: To avoid magic numbers in your code, you can add constants for health and bite values.

```
const MAX_HEALTH = 100;
const BITE_DAMAGE = 50;
const DEAD_VALUE = 0;
```

roman-mirzoian (changes requested):
Good job and please pay attention to the comments.

roman-mirzoian (approved):
Well done!

PR #2878, opened 2024-06-07
Add animal classes

GoVolodya:
Have added three animal related classes.

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  health = 100;
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    Animal.alive.push(this);
+  }
+
+  static get alive() {
+    return this.alive;
+  }
+  static set alive(aliveAnimals) {
+    this.alive = aliveAnimals;
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target.hidden === false) {
```
```suggestion
    if (!target.hidden) {
```
and the bite method should check if the target is an instance of Herbivore, fix it please

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.name = name;
+    Animal.alive.push(this);
+  }
+
+  static get alive() {
+    return this.alive;
+  }
+  static set alive(aliveAnimals) {
+    this.alive = aliveAnimals;
+  }
```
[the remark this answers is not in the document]
nazarmatsevych: Remove it please

nazarmatsevych (changes requested):
Good start, please fix comments below

src/herbivoresAndCarnivores.js
GoVolodya: Sure, it's not always better to have less code, better to make it more readable and predictable instead.

nazarmatsevych (approved):
👍

PR #2885, opened 2024-06-12
Solution

DSB-18:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (
+      animal instanceof Herbivore &&
+      !(animal instanceof Carnivore) &&
+      !animal.hidden
+    ) {
+      animal.health -= BITE_DAMAGE;
+      Animal.alive = Animal.alive.filter((item) => item.health > DEAD_VALUE);
```
roman-mirzoian, line 35: This action can be transferred to a separate method

```
static removeDeadAnimals() {
    Animal.alive = Animal.alive.filter((animal) => animal.health > DEAD_VALUE);
}
```

```
 Animal.removeDeadAnimals();
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (
+      animal instanceof Herbivore &&
+      !(animal instanceof Carnivore) &&
```
roman-mirzoian, line 31: Checking that the victim is not a predator is unnecessary here.
Since the condition animal instanceof Herbivore already guarantees that the victim is not a predator.
```suggestion
```

roman-mirzoian (changes requested):
Good job here are the comments to improve your code.

src/herbivoresAndCarnivores.js
DSB-18: It was an idea from AI 😄
image

src/herbivoresAndCarnivores.js
DSB-18: Done.
Do I have to fix AI comments?

roman-mirzoian (approved):
Well done!

PR #2888, opened 2024-06-13
Solution to task

Karanelus:
All instances of the `Animal` class must have properties `health` and a `name`.
Health is `100` by default.

Create a `Herbivore` class.
Herbivore has a method of `hide`, which sets the `hidden` property of the beast to the value of `true`, and helps to hide from carnivores.

Create a `Сarnivore` class.
Carnivore has a `bite` method, which takes a herbivore object and decreases the object's health by `50`. The method does not work if it is another сarnivore, or the herbivore is currently hiding.

`Herbivore` and `Сarnivore` should extend an `Animal` class.

All alive animals should be in the static `Animal.alive` array.
If the health of the animal becomes less than or equal to `0`, the beast dies and it should be removed from the static array `Animals.alive`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -130,9 +130,9 @@
       }
     },
     "@babel/eslint-parser": {
-      "version": "7.23.10",
-      "resolved": "https://registry.npmjs.org/@babel/eslint-parser/-/eslint-parser-7.23.10.tgz",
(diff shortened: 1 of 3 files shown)


natalia-klonowska (approved):
Great work 🚀

PR #2899, opened 2024-06-18
Solution

ievgen-volkov:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden = false) {
+    super(name, health);
+
+    this.hidden = hidden;
+  }
+
+  hide() {
+    this.hidden = !this.hidden;
```
```suggestion
    this.hidden = true;
```

This method should not toggle hidden value and only change it to true

nazarmatsevych (approved):
Good work 👍

PR #2901, opened 2024-06-19
Solution

alshust:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
vadiimvooo, line 6: use constants to store values such 100, 50, 0 hp

vadiimvooo (changes requested):
Good job, only one suggestion

nazarmatsevych (approved):
Well done 👍

PR #2902, opened 2024-06-19
add solution

AnastasiiaCagle:


anastasiia-tilikina (changes requested):
You shouldn't have remove calling `removeAnimals` method but move it outside if statement to avoid nested if's
now your tests are failed.

pls fix it

etojeDenys (approved):
lgtm

PR #2916, opened 2024-07-01
Solution

IryB-k:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
```
maxim2310, line 8: you could try to use short form of this 
https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      if (!(animal instanceof Carnivore)) {
```
maxim2310, line 28: its unnecessary because if animal is Herbivore, then animal is not Сarnivore for sure

maxim2310 (changes requested):
Good work, but lets make some improvement

TarasHoliuk (approved):

PR #2919, opened 2024-07-02
Solution

MariaKolisnyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Herbivore && !victim.hidden) {
+      victim.health -= 50;
+    }
+
+    if (victim.health <= 0) {
+      Animal.alive.splice(Animal.alive.indexOf(victim), 1);
```
etojeDenys, line 30: you can just use the filter method

etojeDenys (changes requested):

src/herbivoresAndCarnivores.js
MariaKolisnyk: ok

anastasiiavorobiova (approved):
Great work!

PR #2921, opened 2024-07-04
solution

kshvetsova:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  setBite() {
+    if (this.hidden === false) {
```
anastasiiavorobiova, line 26:
```suggestion
    if (!this.hidden) {
```

anastasiiavorobiova: Consider implementing a check of the instance of the animal

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+      this.health -= 50;
+    }
+
+    if (this.health === 0) {
```
anastasiiavorobiova, line 30:
```suggestion
    if (this.health <= 0) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
```
anastasiiavorobiova, line 4:
```suggestion
  static alive = [];
  
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/herbivoresAndCarnivores.js
kshvetsova: It doesn't work, because hidden might be undefined

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  hidden = false;
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  setBite() {
```
TarasHoliuk, line 26: BTW, why do you need `setBite` wrapper?
Just do everything you need inside `bite` method
That might be a reason you need to check it like this: `'hidden' in this`

This particular point isn't critical, but you might check it like this:
animal instanceof Herbivore && !animal.hidden

hidden always defined in Harbivere instance

TarasHoliuk (changes requested):

src/herbivoresAndCarnivores.js
kshvetsova: You are right )

etojeDenys (approved):
lgtm

PR #2958, opened 2024-07-19
Solution

iyehorova:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  static died() {
```
BudnikOleksii, line 5: Method name should start with verb

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+    Animal.alive = Animal.alive.filter(({ health }) => health > 0);
+  }
+
+  health = 100;
+
+  constructor(name) {
```
BudnikOleksii, line 11: You should allow the creation of animals with different  "health", 100 just by default if not passed. And create a constant for this value as it's a magic value in this case

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(bittenAnimal) {
+    if (bittenAnimal instanceof Carnivore || bittenAnimal.hidden === true) {
```
BudnikOleksii, line 27:
```suggestion
    if (bittenAnimal instanceof Carnivore || bittenAnimal.hidden) {
```
You don't need to compare with true as it's already boolean

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+      return;
+    }
+
+    bittenAnimal.health -= 50;
```
BudnikOleksii, line 31: 50 is a magic value

BudnikOleksii (changes requested):
Good work 👍 

Moroz-Dmytro (approved):

PR #2966, opened 2024-07-21
HerbivoresAndCarnivores

victoriasokolvak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
```
BudnikOleksii, line 15: What about `health`?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
BudnikOleksii, line 6: 100 - is a magic number, create a constant for this value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (
+      animal instanceof Herbivore &&
+      !animal.hidden &&
+      !(animal instanceof Carnivore)
```
BudnikOleksii, line 29:
```suggestion
```
You have already checked that it's instance of herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+    ) {
+      animal.health -= 50;
```
BudnikOleksii, line 31: Vagic value as well

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+
+      if (animal.health <= 0) {
+        const index = Animal.alive.indexOf(animal);
+
+        if (index > -1) {
+          Animal.alive.splice(index, 1);
```
BudnikOleksii, line 37: [CODE PERFOMANCE] - Try to avoid nesting loops (indexOf + splice), use filter instead

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

ArturErmolenko (approved):
well done 🔥

PR #2969, opened 2024-07-22
Solution

Kateryna99:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name) {
+    this.health = 100;
```
BudnikOleksii, line 6:
```suggestion
  constructor(name, health = DEFAULT_HEALTH) {
    this.health = health;
```
- You should allow to set different health as well, 100 is only by default in case it's not passed
- And 100 is a magic valu, so create a constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+    if (dead.health <= 0) {
+      Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
+    }
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
```
BudnikOleksii, line 21: The same about health

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+  }
+
+  canBeBitten() {
+    return this.hidden;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
```
BudnikOleksii, line 35: Useless constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+    super(name);
+  }
+
+  bite(animal) {
+    if (!(animal instanceof Carnivore) && !animal.canBeBitten()) {
+      animal.health -= 50;
```
BudnikOleksii, line 41: 50 is a magic value as well

BudnikOleksii (changes requested):
Good job 👍 
Let's fix a few moments

BudnikOleksii (approved):
Looks great 🚀

PR #2976, opened 2024-07-23
add solution

RimmaBuzuluk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
BudnikOleksii, line 4: 100 is magic value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    Animal.alive = Animal.alive.filter((animalItem) => animalItem.health > 0);
+  }
 }
 
+Animal.alive = [];
+
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
```
BudnikOleksii, line 18: What about `health`?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    this.hidden = false;
+  }
+  hide() {
```
BudnikOleksii, line 21: Add blank line between code blocks/functions/methods

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
+      animal.health -= 50;
```
BudnikOleksii, line 29: - You don't need to compare with boolean value such as "false"
- 50 is a magic value as well

BudnikOleksii (changes requested):
Good job 👍 
Let's fix a few moments

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
+const defoultHealth = 100;
+const biteCount = 50;
```
BudnikOleksii, line 4: Constants should be in UPPER_CASE

BudnikOleksii (approved):
👍

PR #2980, opened 2024-07-23
solution

vadimurbanovich:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
 
 class Animal {
   // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
```
BudnikOleksii, line 6: 100 is a magic value, create a constant

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
   // write your code here
+  constructor(name, hidden = false) {
+    super(name);
```
BudnikOleksii, line 16: What about health?

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,41 @@
...
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  bite(animal) {
+    if (
+      !(animal instanceof Carnivore) &&
+      animal instanceof Herbivore &&
+      animal.hidden === false
```
BudnikOleksii, line 31: You don't need to compare with boolean values

BudnikOleksii (changes requested):
Good job 👍

ArturErmolenko (approved):

PR #2988, opened 2024-07-24
add solution

Mazurok001:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
```
BudnikOleksii, line 5: 100 is a magic number, create a constant

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+  }
+
+  removeDead() {
+    Animal.alive = Animal.alive.filter((el) => el.health > 0);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
BudnikOleksii, line 17: You set default health in the parent class, so it's redundant here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
BudnikOleksii, line 30: The same, and constructor is redundant here as well as you don't do any other operations

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+
+  bite(prey) {
+    const isHerbivore = prey instanceof Herbivore;
+    const isHidden = prey.hidden;
+    const canBite = isHerbivore && !isHidden;
+
+    if (canBite) {
+      prey.health -= 50;
```
BudnikOleksii, line 38: 50 is a magic value as well

BudnikOleksii (changes requested):
Good job 👍 
Only a few minor fixes are required

mrkovaliv (approved):
👍 👍 👍

PR #2991, opened 2024-07-25
Solution

SashaPas05:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  die() {
+    if (this.health <= 0) {
+      Animal.alive = Animal.alive.filter((animal) => animal !== this);
+    }
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = FULL_HEALTH) {
```
BudnikOleksii, line 25: You set a default value in the parent class, so you don't need to do it here

BudnikOleksii (approved):
👍

PR #3006, opened 2024-08-05
solution

AlenaMalysheva:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,47 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
etojeDenys, line 37: if you call only super inside the constructor then it's redundant 
```suggestion

```

etojeDenys (approved):

PR #3012, opened 2024-08-07
solved

YevaDrach:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
etojeDenys, line 31: it's redundant to write constructor only with super

etojeDenys (approved):

PR #3018, opened 2024-08-12
classes task

sinner1993:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.health = 100;
+    this.name = name;
+    this.hidden = false;
```
etojeDenys, line 9: only Herbivore should have hidden field

etojeDenys (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #3029, opened 2024-08-21
s_herbivores_and_carnivores

Valentin-1988:


src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,45 @@
...
 }
 
 class Carnivore extends Animal {
   // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
+
+  bite(prey) {
+    if (prey instanceof Herbivore && !prey.hidden && prey.health > 0) {
```
anastasiia-tilikina, line 39:
```suggestion
    if (prey instanceof Herbivore && !prey.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,45 @@
...
   // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  checkHealth() {
```
anastasiia-tilikina, line 13: this method should belong to class not to its instances. Make it static

anastasiia-tilikina (changes requested):
Just a couple of comments

src/herbivoresAndCarnivores.js
```diff
@@ -2,14 +2,45 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
   // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
etojeDenys, line 36: it's redundant to write the constructor only with super, you can remove it
```suggestion

```

src/herbivoresAndCarnivores.js
etojeDenys: image
still not fixed

etojeDenys (changes requested):

anastasiiavorobiova (changes requested):
Consider fixing previous comments

PR #3030, opened 2024-08-22
animal classes solution

MaksymMohyla:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
etojeDenys, line 30: it's redundant to add constructor only with super, you can delete it
```suggestion

```

etojeDenys (approved):

PR #3041, opened 2024-08-30
Solution

joaoguilherme000:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+    Animal.alive.push(this);
+  }
+
+  bite(herbivoreObj) {
+    if (herbivoreObj instanceof Herbivore && herbivoreObj.hidden === false) {
+      herbivoreObj.health -= 50;
+      Animal.alive = Animal.alive.filter((el) => el.health > 0);
```
pedro-ruas, line 33: Instead of filtering the whole array every time a bite occurs, you could check if the herbivore has died, and if so, filter the array.
That way you would remove many unnecessary filters (at least one filter for every herbivore and all filters when a carnivore attacks another)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
pedro-ruas, line 6: The question says all animals start with 100 health, so there's no need to receive health in the constructor, you could simply fix it to 100, just like you did for the `hidden` property for the Herbivore.
But this is ok, as it gives the code more flexibility if future implementations ask for it

pedro-ruas (approved):
Your code looks great and I see you have understood the assignment. There are two points that I commented which you can take a look for future references, but all good!

PR #3042, opened 2024-08-30
Solution

WGoliasz:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
natalia-klonowska, line 9: only herbivores need hidden property so move it to Herbivore class making use of super in constructor

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore.hidden === false && herbivore instanceof Herbivore) {
```
natalia-klonowska, line 28:
```suggestion
    if (!herbivore.hidden && herbivore instanceof Herbivore) {
```

natalia-klonowska (changes requested):
Great work 😃

PR #3048, opened 2024-09-01
solution

Svidruk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
+    this.name = name;
+    this.health = health;
+    this.hidden = false;
```
BudnikOleksii, line 9: Animal class doesn't have this property, only `Herbivore` has
image

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive = Animal.alive.filter((animal) => animal !== this);
+    }
```
BudnikOleksii, line 19:
```suggestion
    Animal.alive = Animal.alive.filter((animal) => animal !== this);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
BudnikOleksii, line 6: 100 and 50 are magic values, create constants for them

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
BudnikOleksii, line 36: You  can remove `constructor` if you don't add any additional properties

BudnikOleksii (changes requested):
Good work 👍 
Just a few fixes are required

PR #3060, opened 2024-09-04
add task solution

Syogunok:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
```
etojeDenys, line 15: we can simplify it just using the filter method

etojeDenys (approved):
good job

PR #3068, opened 2024-09-06
solved task

PikalovAndrey:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
BudnikOleksii, line 6: 100 and 50 are magic values, create constants for them

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,44 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
BudnikOleksii, line 31:
```suggestion
```
If you don't add any new properties you don't need a constructor

BudnikOleksii (approved):
Good work 👍 
Approved, but check the comments

PR #3074, opened 2024-09-09
add task solution

Dorosh90:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
+
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
```
etojeDenys: you can push it only one time in Animal constructor

etojeDenys (changes requested):

etojeDenys (approved):

PR #3078, opened 2024-09-12
Add task solution

WatarJoy:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal.hidden !== true && animal instanceof Herbivore) {
```
igorkruz: create a variable for this condition

```suggestion
    const isHerbivore = animal instanceof Herbivore;
    if (!animal.hidden && isHerbivore) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,39 @@
...
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        Animal.removeDeadAnimals();
+      }
```
igorkruz: you can directly call Animal.removeDeadAnimals() after the health is reduced, regardless of the animal's exact health. This ensures the removal is always checked after any bite

```suggestion
        Animal.removeDeadAnimals();
```

igorkruz (changes requested):
Great 🔥
Let's improve your solution

BudnikOleksii (approved):
👍

PR #3079, opened 2024-09-12
Solution

hma-3:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+  constructor(name) {
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
```
igorkruz: The hidden property is well-handled, but consider explicitly setting it in the constructor to avoid confusion about which animals have it
```suggestion
   constructor(name) {
    super(name);
    this.hidden = false;
  }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (herbivore instanceof Herbivore && herbivore.hidden === false) {
```
igorkruz: you can move this conditions to a variable
```suggestion
    const isHerbivore = animal instanceof Herbivore;
    if (isHerbivore  && !herbivore.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+      herbivore.health -= 50;
+
+      if (herbivore.health <= 0) {
+        Animal.alive = Animal.alive.filter((animal) => animal !== herbivore);
+      }
```
igorkruz:
```suggestion
        Animal.alive = Animal.alive.filter((animal) => animal.health !== 0);

```

igorkruz (changes requested):
Great 🔥
Let's improve your solution

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animalToBite) {
+    const isHerbivore = animalToBite instanceof Herbivore;
+
+    if (isHerbivore && animalToBite.hidden === false) {
```
igorkruz: you can use `!animalToBite.hidden` instead of `animalToBite.hidden === false`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT

```suggestion
    if (isHerbivore && !herbivore.hidden) {
```

igorkruz (changes requested):
GJ 👍 
please refer to the attached link and improve your solution

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT

igorkruz (approved):
Good Job 🔥

PR #3082, opened 2024-09-13
Solution

Hradivnyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 
 class Animal {
-  // write your code here
+  static alive = [];
+  health = 100;
+
+  constructor(name) {
+    this.name = name;
+    Animal.alive.push(this);
+  }
```
```suggestion
  constructor(name, health = DEFAULT_HEALTH) {
    this.name = name;
    this.health = health;
    
    Animal.alive.push(this);
  }
```
- You should accept health in case it will be different, just by default it should be 100

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(obj) {
```
BudnikOleksii: `obj/arr/object/array/res/result/item/etc` are bad names

BudnikOleksii (changes requested):
Good work 👍

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    Animal.alive.push(this);
+  }
+
+  demage(amout = DEFAULT_DEMAGE) {
+    this.health -= amout;
+
+    if (this.health <= 0) {
+      Animal.alive = Animal.alive.filter((item) => item !== this);
+    }
+  }
```
VolodymyrKirichenko: 1) Move this method to `Carnivore` class, cuz u `Herbivore` should not have access to this method
2) `item !== this`? U need to remove all values ​​that are equal to 0 or less. Change the condition in ur filter
3) After the move, u won't need to create a (damage) method to perform the action)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(victim) {
+    if (victim instanceof Carnivore || victim.hidden) {
+      return;
+    }
+
+    victim.demage(DEFAULT_DEMAGE);
```
VolodymyrKirichenko: do the action only when 2 conditions are met.
u can shorten ur code
```suggestion
    if (victim instanceof Herbivore && !victim.hidden) {
       victim.demage(DEFAULT_DEMAGE);
    }
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = DEFAULT_HEALTH) {
+    this.name = name;
+    this.health = DEFAULT_HEALTH;
+    Animal.alive.push(this);
+  }
+
+  demage(amout = DEFAULT_DEMAGE) {
```
VolodymyrKirichenko: syntax error
```suggestion
  damage(amount = DEFAULT_DEMAGE) {
```

VolodymyrKirichenko (changes requested):
Nice work 👍 
but let's improve ur code

BudnikOleksii (approved):
Good job 👍

PR #3083, opened 2024-09-13
add task solution

D-Ptashevskyi:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
```
```suggestion
  constructor(name, health = DEFAULT_HEALTH) {
    this.name = name;
    this.health = health;
    
    Animal.alive.push(this);
  }
```
- You should accept health in case it will be different, just by default it should be 100
- Only Herbivore has a `hidden` property
image

BudnikOleksii (changes requested):
Good work 👍

BudnikOleksii (approved):

PR #3084, opened 2024-09-13
Solution

KrivorotkoYurii:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
```
VolodymyrKirichenko, line 5: use CONSTANTS instead magic values(in all the cases)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
VolodymyrKirichenko, line 24: do u need constructor in this place? why?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    super(name, health);
+  }
+
+  bite(beast) {
+    if (beast instanceof Herbivore && beast.hidden === false) {
```
VolodymyrKirichenko, line 29: will better
```suggestion
    if (beast instanceof Herbivore && !beast.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hidden = false;
+  constructor(name, health = 100) {
```
VolodymyrKirichenko, line 14: I don't think u need it since u have a constructor in the parent class.
At least - health is not needed

VolodymyrKirichenko (approved):
Good job, but check my comments and fix them 🥹 
Try to pick up the constructors one by one and see the result

PR #3085, opened 2024-09-13
add solution

YuliiaShatkovska:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
+    if (!herbivore.hidden && herbivore instanceof Herbivore) {
+      herbivore.health -= 50;
+
+      Animal.alive = Animal.alive.filter((animal) => animal.health !== 0);
```
VolodymyrKirichenko: Instead of filtering out animals with exactly 0 health, you should filter out animals with health less than or equal to 0
```suggestion
      Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+  }
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(herbivore) {
```
VolodymyrKirichenko: you're taking the wrong argument.
It can be not only a herbivore

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
```
VolodymyrKirichenko: what the magic numbers? Use CONSTANTS for all such cases

VolodymyrKirichenko (changes requested):
Good job 👍 
But review my comments and fix them

VolodymyrKirichenko (approved):
Well done! 🔥

PR #3087, opened 2024-09-13
solution

Ostkreuzzz:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    this.health = health;
+    this.name = name;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && animal.hidden === false) {
```
BudnikOleksii: You don't need to compare with false as it's already boolean

BudnikOleksii (changes requested):
Good work 👍

BudnikOleksii (approved):

PR #3088, opened 2024-09-13
task solution

Ivanavi808:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+  }
+
+  die() {
+    Animal.alive = Animal.alive.filter((animal) => animal !== this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```

BudnikOleksii (changes requested):
Good work 👍

BudnikOleksii (approved):
👍

PR #3090, opened 2024-09-14
add solution

stasLiutyww:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
BudnikOleksii, line 8:
```suggestion
  constructor(name, health = DEFAULT_HEALTH) {
    this.name = name;
    this.health = health;
```
- You should accept the health as well, by default it should be 100

BudnikOleksii (changes requested):
Good work 👍

PR #3092, opened 2024-09-14
Solution

v1RnT:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
```suggestion
  constructor(name, health = DEFAULT_HEALTH) {
    this.name = name;
    this.health = health;
```
- You should accept health as well, only by default it's 100

BudnikOleksii (changes requested):
Good work 👍

BudnikOleksii (approved):
👍

PR #3095, opened 2024-09-15
task has been solved

coldmelnyk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+
+    Animal.alive.push(this);
+  }
+
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= 50;
+
+      if (victim.health === 0) {
+        Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```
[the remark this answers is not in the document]
coldmelnyk: I want to do filtering only in cases when somebody was killed for saving performance

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    this.name = name;
+    this.health = health;
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    super(name, health);
+    this.hidden = hidden;
+
+    Animal.alive.push(this);
```
BudnikOleksii: You can push in the animal class so you don't have to do it in all other classes that extend it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,45 @@
...
+    super(name, health);
+
+    Animal.alive.push(this);
+  }
+
+  bite(victim) {
+    if (victim instanceof Herbivore && victim.hidden === false) {
+      victim.health -= 50;
+
+      if (victim.health === 0) {
```
```suggestion
      if (victim.health <= 0) {
```
In case `victim` health is 30 for example and then get damage 50 the health become -20 but it will be still life with your condition

BudnikOleksii (changes requested):
Good work 👍

BudnikOleksii (approved):

PR #3096, opened 2024-09-15
solving task

EustasTheMonk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  health = 100;
+  constructor(name) {
+    this.name = name;
```
```suggestion
  constructor(name, health = DEFAULT_HEALTH) {
    this.name = name;
    this.health = health;
```
- You should accept health as well, only by default it's 100

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,31 @@
...
+  hidden = false;
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && target.hidden !== true) {
```
BudnikOleksii: You don't need to compare with true, it's already boolean

BudnikOleksii (changes requested):
Good work 👍

BudnikOleksii (approved):
👍

PR #3097, opened 2024-09-16
solution

misulkin:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    Animal.alive = Animal.alive.filter((animal) => animal !== this);
+  }
```
etojeDenys, line 14: this method manipulates the static array so you can make it static too

etojeDenys (approved):

PR #3098, opened 2024-09-16
solution

mbulchak:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
VolodymyrKirichenko: use enum for all CONSTANTS

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,36 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static alive = [];
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
```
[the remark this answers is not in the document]
VolodymyrKirichenko: The constructor of the `Herbivore` class should not include a hidden parameter, as the hide method should be responsible for setting the hidden property to true. Please remove the hidden parameter from the constructor.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
...
+    super(name, health);
+    this.hidden = false;
```
VolodymyrKirichenko, line 20:
```suggestion
```

VolodymyrKirichenko (changes requested):
Good job 🔥 
But review my comments and fix them🥹

VolodymyrKirichenko (approved):
Well done 🔥

PR #3099, opened 2024-09-16
add solution

Tulenty:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+  }
+
+  hide(){
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name){
```
TarasHoliuk: image

Fix it (you can completely remove constructor here - it should work properly anyway)

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,49 @@
...
+    super(name);
+  }
+
+  bite(herbivore){
+    if(!(herbivore instanceof Herbivore)){
+      return;
+    }
+    if(herbivore.hidden){
+      return;
+    }
```
TarasHoliuk:
```suggestion
  bite(herbivore) {
    if (!(herbivore instanceof Herbivore)) {
      return;
    }

    if (herbivore.hidden) {
      return;
    }
```
Such formatting also might lead to linter errors - fix formatting

TarasHoliuk (changes requested):
GJ! But you should fix linter errors

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    Animal.alive.push(this);
+  }
+
+  die() {
+    Animal.alive = Animal.alive.filter((animal) => animal !== this);
+  }
```
etojeDenys, line 14: better to make this method static, because it manipulates the static Animal field

etojeDenys (approved):

PR #3144, opened 2024-10-13
solving problem

kami2693:
but i really have problems with it

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,16 +1,69 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
+    Animal.alive.push(this);
+  }
+
+  isAlive() {
+    return this.health > 0;
+  }
+
+  checkHealth() {
+    if (this.health <= 0) {
+      this.die();
+    }
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
+    }
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  hide() {
+    this.hidden = true;
+  }
+
+  checkHealth() {
+    super.checkHealth();
+
+    if (!this.isAlive()) {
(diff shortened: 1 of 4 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,70 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
BudnikOleksii, line 9: - Only `Herbivore` class should have `hidden` property
- 100, 50 are magic values, create constants

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,70 @@
...
+    Animal.alive.push(this);
+  }
+
+  isAlive() {
+    return this.health > 0;
+  }
+
+  checkHealth() {
+    if (this.health <= 0) {
```
BudnikOleksii, line 18:
```suggestion
    if (!this.isAlive()) {
```
You have a method to check if animal is alive

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,70 @@
...
+      this.die();
+    }
+  }
+
+  die() {
+    const index = Animal.alive.indexOf(this);
+
+    if (index !== -1) {
+      Animal.alive.splice(index, 1);
```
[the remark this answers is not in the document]
BudnikOleksii: > [CODE PERFORMANCE] You are using `indexOf` and `splice` to remove the dead animal from the `alive` array. This can be inefficient because `indexOf` scans through the array to find the animal, and `splice` also scans through the array to remove it. It would be more efficient to use the `filter` method, which only scans through the array once.

Good point

src/herbivoresAndCarnivores.js
```diff
@@ -1,17 +1,70 @@
...
+    this.hidden = true;
+  }
+
+  checkHealth() {
+    super.checkHealth();
+
+    if (!this.isAlive()) {
+      this.hidden = false; // Reset hidden state when it dies
+    }
+  }
```
BudnikOleksii, line 43: You don't need update the method for `Herbivore`

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments to improve your solution

PR #3171, opened 2024-10-24
Task solution

Dominik-Dudek-code:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
-  // write your code here
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+    }
+
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.filter(
+        (currentAnimal) => currentAnimal !== animal,
+      );
```
zibi-dev: What would be better is to have this as method inside Animal class

Zibi95 (changes requested):
Hey there. One small change and we are fine.

Dominik-Dudek-code:
Is that what u meant?

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.health = health;
+    Animal.alive.push(this);
+  }
+
+  static isAnimalDead(animal) {
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.filter(
+        (currentAnimal) => currentAnimal !== animal,
+      );
+    }
```
zibi-dev: It don't have to be static. We can call this method on object instance.` this.isAnimalDead`

Zibi95 (approved):
Yea. Its fine. I have left one comment.

PR #3198, opened 2024-11-12
solution

k0tlik02:
Herbivores and Carnivores solution

### DIFF
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -22,7 +22,7 @@ class Herbivore extends Animal {
     this.hidden = hidden;
   }
 
-  hide(hidden) {
+  hide() {
     this.hidden = true;
   }
 }
--- a/src/herbivoresAndCarnivores.js
+++ b/src/herbivoresAndCarnivores.js
@@ -1,9 +1,12 @@
 'use strict';
 
+const FULL_HEALTH = 100;
+const DAMAGE_FROM_BITE = 50;
+
 class Animal {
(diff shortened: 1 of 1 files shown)


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  checkHealth() {
+    if (this.health <= 0) {
+      Animal.alive = Animal.alive.filter((animal) => animal !== this);
+    }
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
BudnikOleksii: You set a default health in the `Animal` class

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
BudnikOleksii: You don't need `constructor` if you don't set any other properties

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions to improve your solution

PR #3199, opened 2024-11-12
solution

Fridrif27:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
+    this.hidden = false;
```
BudnikOleksii, line 7: - Only `Herbivore` class should have `hidden` property
- 100, 50 are magic values, create constants

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
+
+    Animal.alive.push(this);
+  }
+
+  static removeAnimal(animal) {
+    Animal.alive = Animal.alive.filter((a) => a !== animal);
```
BudnikOleksii, line 13: What is `a`? Avoid one-letter names

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor changes are required

PR #3201, opened 2024-11-13
Solution

MakksymSly:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  }
+  static alive = [];
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+    Animal.alive.push(this);
```
BudnikOleksii: You should do it in `Animal` class so you don't have to do it in all classes that inherit it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    Animal.alive.push(this);
+  }
```
BudnikOleksii: The same, and you don't need constructor if you don't set any new properties

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,41 @@
...
+  bite(animal) {
+    if (animal instanceof Herbivore && !animal.hidden) {
+      animal.health -= 50;
+
+      if (animal.health <= 0) {
+        if (Animal.alive.indexOf(animal) !== -1) {
+          Animal.alive.splice(Animal.alive.indexOf(animal), 1);
```
BudnikOleksii: You do `Animal.alive.indexOf(animal)` twice, also it's better to use `filter` instead

BudnikOleksii (changes requested):
Good work 👍 

BudnikOleksii (approved):
Looks good 👍

PR #3206, opened 2024-11-13
Create classes

AlyonaV22:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  constructor(name, health = 100) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
+  static alive = [];
+
+  static animalHealth(animal) {
```
```suggestion
  static checkHealth(animal) {
```
Function/method name should start from verb

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    if (animal.health <= 0) {
+      Animal.alive = Animal.alive.filter((a) => a !== animal);
```
BudnikOleksii: What is `a`? Avoid one-letter names

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+    }
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
```
BudnikOleksii: You set a default health in the parent class, so you don't need it here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
BudnikOleksii: You don't need constructor if you don't set other properties or do other things except calling super

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,46 @@
...
+  bite(herb) {
+    if (herb instanceof Herbivore && !herb.hidden) {
+      herb.health -= 50;
+
+      Animal.animalHealth(herb);
+    }
+
+    if (herb instanceof Carnivore) {
+      return herb.health;
+    }
```
BudnikOleksii: It isn't related to bite method

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments, check them and improve your solution

BudnikOleksii (approved):
Nicely done 🔥

PR #3209, opened 2024-11-14
Solution

natalia646:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(name, health = 100) {
```
BudnikOleksii, line 5: 50, 100 are magic values, create constants

BudnikOleksii (approved):
Good work 👍

PR #3210, opened 2024-11-14
add task solution

taniabarkovskya:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name, health = 100) {
```
BudnikOleksii: 50, 100 are magic values, create constants

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+    this.name = name;
+    this.health = health;
+
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
```
BudnikOleksii: You set default health in the Animal class so you don't need it to do it here

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health = 100) {
+    super(name, health);
+  }
```
BudnikOleksii: If you don't add new properties or do other operations except calling `super` you can get rid of `constructor` at all

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+
+  bite(herb) {
+    if (herb instanceof Herbivore && herb.hidden === false) {
```
```suggestion
    if (herb instanceof Herbivore && !herb.hidden) {
```
You don't need to compare boolean values

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+      herb.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```
BudnikOleksii: Do you need to do it in any case?

BudnikOleksii (changes requested):
Good work 👍 

StasSokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #3216, opened 2024-11-15
Solution

SemenVodolazskij:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
BudnikOleksii: 100, 50 are magic values

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    Animal.alive.push(this);
+  }
 }
 
+// ______________________________//
```
BudnikOleksii: Remove all comments

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    this.hidden = true;
+  }
 }
 
+// ______________________________//
+
 class Carnivore extends Animal {
-  // write your code here
+  bite(bittenAnimal) {
+    if (bittenAnimal.hidden !== true && bittenAnimal instanceof Herbivore) {
```
```suggestion
    if (!bittenAnimal.hidden && bittenAnimal instanceof Herbivore) {
```
You don't need to compare with true or false if it's a boolean value

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+      bittenAnimal.health -= 50;
+    }
+    Animal.alive = Animal.alive.filter((animal) => animal.health > 0);
```
BudnikOleksii, line 29: You don't need to do it on every `bite` trigger, only in case the target animal has no health

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

src/herbivoresAndCarnivores.js
SemenVodolazskij: ok, thank)

src/herbivoresAndCarnivores.js
SemenVodolazskij: What do you mean?

BudnikOleksii: > What do you mean?

I mean filter `Animal.alive`

BudnikOleksii (approved):

PR #3235, opened 2024-11-17
Solution

LiliiaVol:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  health = 100;
```
Moroz-Dmytro: Magic value - 100
Would be good to create variable for it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(selectedHerbivore) {
+    if (selectedHerbivore.hidden === false) {
```
Moroz-Dmytro:
```suggestion
    if (!selectedHerbivore.hidden) {
```

And you should also check if this is a `Herbivore`

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+      selectedHerbivore.health = selectedHerbivore.health - 50;
```
Moroz-Dmytro: Magic value - 50

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    }
+
+    if (selectedHerbivore.health === 0) {
```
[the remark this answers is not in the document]
Moroz-Dmytro: Good point, fix it

Moroz-Dmytro (changes requested):
Good work 👍 
But fix comments below

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name) {
+    super(name);
+    this.hidden = false;
+  }
```
VolodymyrKirichenko: 1) u will have the same initial value in each instance.
2) In this case, we may not call the constructor
```suggestion
  hidden = false
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,43 @@
...
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(selectedHerbivore) {
+    if (
+      !selectedHerbivore.hidden &&
+      Object.hasOwn(selectedHerbivore, 'hidden')
```
VolodymyrKirichenko: u can use `selectedHerbivore instanceof Herbivore` for i, cuz in the future, carnivore can learn to hide too))

VolodymyrKirichenko (approved):
Well done 🔥 
But review my comments 🥹

PR #3250, opened 2024-11-22
sort herb and carn

yzhyhaliuk:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,35 @@
...
+    if (prey instanceof Herbivore && prey.hidden === false) {
+      prey.health -= 50;
+    }
+
+    if (prey.health <= 0) {
+      const index = Animal.alive.indexOf(prey);
+
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
+      }
```
etojeDenys, line 30: you can just simplify it using the filter method

etojeDenys (approved):

PR #3279, opened 2024-12-07
Solution

oleg19790807:


VitaliyBondarenko1982 (changes requested):
Good job.
Try do not use eslint-disable.
It is real to solve without ignore eslint rules.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
...
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  constructor(name, health, hidden) {
+    super(name, health, hidden);
+  }
+
+  bite(item) {
+    if (item.hidden === false && item instanceof Herbivore) {
```
volodymyr-soltys97, line 35:
```suggestion
    if (!item.hidden && item instanceof Herbivore) {
```

volodymyr-soltys97 (approved):
Good job!

PR #3306, opened 2024-12-20
solution

tkachuk2291:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+  }
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target.hidden === false) {
```
VitaliyBondarenko1982, line 24:
```suggestion
    if (!target.hidden) {
```

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,38 @@
...
+      target.health -= 50;
+    }
+
+    if (target.health <= 0) {
+      const index = Animal.alive.indexOf(target);
+
+      if (index !== -1) {
+        Animal.alive.splice(index, 1);
+      }
```
VitaliyBondarenko1982, line 33:
```suggestion
        Animal.alive = Animal.alive.filter(animal => animal.health > 0);
```

looks like it is possible to make in one array iteration.

VitaliyBondarenko1982 (approved):
Looks good.

PR #3336, opened 2025-01-08
add task solution

Uusuff:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.name = name;
+    this.health = health;
+    Animal.alive.push(this);
+  }
 }
 
 class Herbivore extends Animal {
-  // write your code here
+  constructor(name, health = 100, hidden = false) {
+    super(name, health, hidden);
```
maxim2310, line 18: you dont need to pass it in supper, because constructor Animal doesn't have it
```suggestion
    super(name, health);
```

maxim2310 (approved):
Good work

PR #3504, opened 2025-03-03
Solution

MariiaKapustkina:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,42 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+  constructor(health = 100, name) {
+    this.health = 100;
```
alexander-ignatow, line 6: You pass a `health` parameter to the constructor, but you do not use it.
Please fix this

alexander-ignatow (approved):
Good job 👍 

But please, be careful and mind the issues, like one below!

PR #3652, opened 2025-04-26
Update herbivoresAndCarnivores.js

mslol21:


raulriato (approved):
Good job!

A tip, though: Try to create a new `branch` `develop` to write your code. It's not a big problem to use the `master` branch to study, but it's nice to practice creating and using different branches, because it will be mandatory on your day-to-day work.

PR #3715, opened 2025-05-12
add solution

ystyslo:


ystyslo:
@mateacademy-ai-mentor fixed all misstakes.

PR #3914, opened 2025-08-15
Solution

Augustwise:


Augustwise:
I corrected the solution, but the original requirements (`readme.md` file) did not specify that `Array.prototype.filter` should be used. I consider this to be an error on your part, as I only learned about it from you after the commit.

PR #3922, opened 2025-08-18
Solution

Olivio-Neves:


pedro-ruas (changes requested):
Well, your code does not follow all requests from the checklist, and it was clearly made by AI.
Please, be sure to understand the requirements and implement your own logic, or at least check the code before sending it.

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  constructor(name, health = 100) {
+    this.health = health;
+    this.name = name;
+
+    Animal.alive.push(this);
+  }
+
+  dano(valor) {
+    this.health = this.health - valor;
+    Animal.alive = Animal.alive.filter((a) => a !== this);
```
pedro-ruas: This is wrong. 
- You cannot directly compare 2 objects, so `a !== this` will not work
- The animal should not die for every damage, only when the health falls bellow 0

pedro-ruas (changes requested):
There is wrong logic implemented, check my coment

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,40 @@
...
+    this.name = name;
+
+    Animal.alive.push(this);
+  }
+
+  dano(valor) {
+    this.health = this.health - valor;
+
+    if (this.health <= 0) {
+      Animal.alive = Animal.alive.filter((a) => a !== this);
```
raulriato: Again, as the other mentor already said, you can't compare an object with another object, so `a !== this` does not work. You should filter by the `health` instead.

raulriato (changes requested):
You still have the same problem.

pedro-ruas (approved):

PR #4018, opened 2025-09-29
Solution

Artur2912:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
Moroz-Dmytro, line 8: 100 is a magic value, create variable for it

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+    Animal.alive.push(this);
+  }
+
+  checkAlive() {
+    Animal.alive = Animal.alive.filter(a => a.health > 0);
```
Moroz-Dmytro, line 13: `a` is a bad name

src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,37 @@
...
+  hide() {
+    this.hidden = true;
+  }
 }
 
 class Carnivore extends Animal {
-  // write your code here
+  bite(target) {
+    if (target instanceof Herbivore && !target.hidden) {
+      target.health -= 50;
```
Moroz-Dmytro, line 31: Magic value 50, create variable

Moroz-Dmytro (approved):
Good work 👍 
But check comments below

PR #4020, opened 2025-09-29
Add task solution

petrykgit:
Made classes Animal, Herbivore and Carnivore according to the conditions of the task.


Moroz-Dmytro (approved):
Approve it, but don't use "magic values"

PR #4048, opened 2025-10-19
add: solution

6ORSH:


src/herbivoresAndCarnivores.js
```diff
@@ -1,15 +1,34 @@
 'use strict';
 
 class Animal {
-  // write your code here
+  static alive = [];
+
+  constructor(name) {
+    this.name = name;
+    this.health = 100;
```
FaiHamid: It is worth avoiding magic numbers and adding global variables for the values 100, 50,0.
`const MAX_HEALTH = 100;
const BITE_DAMAGE = 50;
const DEAD_VALUE = 0;`

FaiHamid (changes requested):
Good job!
Let's just change one small detail:

vitalii-bondarenko-ua (approved):
Looks good.

PR #4433, opened 2026-06-29
Add herbivoresAndCarnivores.js

JuliaTr:
Solved problem on JavaScript OOP, using classes, inheritance, static public properties, constants, objects collaboration (aggregation, reference).


2pasha (approved):
lgtm 🔥

